PR #463, opened 2013-11-08
Working with older versions of cURL

mtdowling:


ejunker:
Trying to use AsyncPlugin but it still references `CURLOPT_TIMEOUT_MS`

```php
curl_setopt($event['handle'], CURLOPT_TIMEOUT_MS, 1);
```

I am using PHP 5.3.3 and Guzzle 3.9.3. Any chance this could be fixed? This is for a legacy application that cannot be upgraded to newer versions of PHP or Guzzle otherwise I would upgrade.

Issue #189, opened 2012-12-06
Reponse JSON Object vs Associative Arrays


splicebison:
It would be nice if the json() function could have the option to convert to an object as well as an array. Simply passing json(true|false) should follow PHP's method and default to an object rather than an array.

PR #580, opened 2014-02-28
Optional parameter to get JSON Objects back as Objects.

frankiethekneeman:
This is a feature that's been asked for, and upvoted at least a couple times.  While I see you closed [Issue #189](https://github.com/guzzle/guzzle/issues/189), it's a simple and easy feature that allows users greater control over what is, after all, _their own data_.  I think this is a minor upgrade that would mean a lot.

### DIFF
--- a/src/Guzzle/Http/Message/Response.php
+++ b/src/Guzzle/Http/Message/Response.php
@@ -851,12 +851,14 @@ class Response extends AbstractMessage implements \Serializable
     /**
      * Parse the JSON response body and return an array
      *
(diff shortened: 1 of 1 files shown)


kimoi:
+1

Using objects is the default behaviour of json_decode(). There should at least be the option to have the convenience method json() return them as objects.

jeskew:
@kimoi In Guzzle 5.3, you can call `$response->json(['object' => true])` to get the response decoded as objects. The `json` method was not kept in Guzzle 6.

kimoi:
@jeskew Sorry, you're right. The library I'm using references an archaic version of Guzzle.

PR #643, opened 2014-04-18
Use client session to persist the cookie jar

MarkBaker:
Working from multiple load-balanced app servers where I'm making regular requests to a web-service, but can't persist the web service session cookies to file because there is no shared file across all the app servers, but where session (maintained in redis) is accessible from all app servers, I created the following to store cookies in session


geoidesic:
Could someone provide a link to said documentation?

PR #890, opened 2014-10-31
Guzzle HTTP Namespace

JREAM:
This doesn't work with GuzzleHttp it's Guzzle\Http


Erutan409:
I was going to create an issue for this exact same reason.  The only way I even became aware of this "package" was because of Elasticsearch's dependency on guzzle.  Their composer config is setup to get v3, however.  Should they update their dependencies to use a newer version?

PR #918, opened 2014-11-21
Adding integration tests for Digest authentication

outtersg:
This intends to prepare Digest implementation, to prevent regressing from the current (cURL) implementation.

Note that this requires an npm install http-auth; I'm quite uneasy on how to require it: what is the usage regarding this? Should it be included in the composer, or will it be sufficient to mention this new requirement, along Node.js, on the blog?
(the RingPHP pull request https://github.com/guzzle/RingPHP/pull/14, to include Digest auth in server.js, is not dependent on this nom install: the http-auth won't be required unless the specific Digest-requiring URI is called)

### DIFF
--- a/docs/events.rst
+++ b/docs/events.rst
@@ -11,12 +11,12 @@ Event Emitters
 ==============
 
 Clients, requests, and any other class that implements the
-``GuzzleHttp\Common\HasEmitterInterface`` interface have a
-``GuzzleHttp\Common\EventEmitter`` object. You can add event *listeners* and
+``GuzzleHttp\Event\HasEmitterInterface`` interface have a
+``GuzzleHttp\Event\Emitter`` object. You can add event *listeners* and
 event *subscribers* to an event *emitter*.
 
 emitter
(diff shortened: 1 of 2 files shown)


outtersg:
OK, back to work (after some holidays, diapers occupation, and a lot of other things…).
I like your idea. So I made a new pull request on guzzle/RingPHP (to add replying with a 501), and added a commit to the (not-yet-merged) guzzle/guzzle pull request (to catch the 501 and decide to skip the test).
Hmm… I'm quite new with github, so I did reset in my outtersg/guzzle master branch (the one requested for pull) to get rid of the travis.yml "npm install" addition, before adding the (more applicative-way) HTTP-501 handling. Tell me if this breaks the pull request.

mtdowling:
This is great. Can you rebase against master?

outtersg:
Still learning git (and github)… removing non-mergeable patches (before rebasing) made GitHub close the pull request (in a non-reopenable state, as it was considered merged; a 0-commit merge, but still a merge). So followup is at https://github.com/guzzle/guzzle/pull/962

PR #944, opened 2014-12-14
Added new PostField class to allow header with simple fields (multipart request fix)

Nek-:
Hello,

I had a problem using Guzzle: I needed to make a "classic" multipart request. That means that I didn't want to make a "form post". And for now guzzle can only make multipart request like a form.

With my changes you can now add fields that doesn't have name. It will generate part without "Content-Disposition" header. Also you can add an header for a field, like that you can specify the type of your part or any other header you want.

Actually my changes are not perfect IMO but here are my motivation to change like that:
* Avoid modifying the logic of the current code
* Access of methods to avoid adding potential deprecations needs in the future

I think the test is complete but i'm not sure about that, please check it.

I hope you will like this PR :-) .

> For more info, see here: http://www.w3.org/Protocols/rfc1341/7_2_Multipart.html

### DIFF
--- a/src/Post/MultipartBody.php
+++ b/src/Post/MultipartBody.php
@@ -51,18 +51,30 @@ class MultipartBody implements StreamInterface
      */
     private function getFieldString($name, $value)
     {
+        $boundary = sprintf("--%s\r\n", $this->boundary);
+
+        if ($value instanceof PostFieldInterface) {
+            $headers = $value->getHeaders();
+            if ($name = $value->getName() && !isset($headers['Content-Disposition'])) {
+                $value->addHeader('Content-Disposition', sprintf('form-data; name="%s"', $value->getName()));
+            }
+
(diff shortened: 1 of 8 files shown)


Nek-:
rebased

mtdowling:
Thanks for the PR. I've been thinking about this one for a while, and I don't think Guzzle needs this abstraction. A multipart/form-data abstraction is provided because it handles an extremely common use case. I see that plain multipart support is useful in some cases, so perhaps a multipart abstraction that creates Guzzle streams would be a good project that people could use when they need multipart bodies. They could simple create the stream and pass it to the `body` request option.

Nek-:
Hello,

You closed this PR, you let me 2 choices:

* Fork guzzle adding my feature
* Drop guzzle from all my open source project assuming it's not complete and you don't want to complete it

Because as you probably notice, I can't add a plugin upon classes that use their own private method to work.

I would like to avoid that. Do you have something to suggest to me ?

PR #953, opened 2015-01-07
Add multipart stream/response iterator

FabioBatSilva:
Hi Guys,

This patch add support for multipart [responses](http://www.w3.org/Protocols/rfc1341/7_2_Multipart.html)

### DIFF
--- a/src/Message/MultipartResponseIterator.php
+++ b/src/Message/MultipartResponseIterator.php
@@ -1,0 +1,115 @@
+<?php
+
+namespace GuzzleHttp\Message;
+
+use Iterator;
+use InvalidArgumentException;
+use GuzzleHttp\Message\MessageParser;
+use GuzzleHttp\Message\MessageFactory;
+use GuzzleHttp\Message\ResponseInterface;
+use GuzzleHttp\Message\MessageFactoryInterface;
+
+/**
+ * Multipart response iterator
+ *
+ * @author Fabio B. Silva <fabio.bat.silva@gmail.com>
+ */
+class MultipartResponseIterator implements Iterator
+{
+    /**
+     * @var \GuzzleHttp\Message\ResponseInterface
+     */
+    private $response;
+
+    /**
+     * @var \GuzzleHttp\Message\MultipartStreamIterator
+     */
+    private $iterator;
+
+    /**
+     * @var \GuzzleHttp\Message\MessageParser
+     */
+    private $parser;
+
+    /**
+     * @var \GuzzleHttp\Message\Response
+     */
+    private $current;
+
+     /**
+     * @var \GuzzleHttp\Message\MessageFactoryInterface
+     */
+    private $factory;
+
+    /**
+     * @param \GuzzleHttp\Message\ResponseInterface       $response
+     * @param \GuzzleHttp\Message\MessageParser           $parser
+     * @param \GuzzleHttp\Message\MessageFactoryInterface $messageFactory
+     */
+    public function __construct(ResponseInterface $response, MessageParser $parser = null, MessageFactoryInterface $messageFactory = null)
+    {
+        $matches = null;
+        $body    = $response->getBody();
+        $header  = $response->getHeader('Content-Type');
+
+        if ( ! preg_match('/boundary=(.*)$/', $header, $matches) || ! isset($matches[1])) {
+            throw new InvalidArgumentException("Unable to parse multipart content [Content-Type : $header]");
+        }
+
+        $this->response = $response;
+        $this->parser   = $parser ?: new MessageParser();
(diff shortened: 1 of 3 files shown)


mtdowling:
This looks really cool! I don't think that this belongs in Guzzle's core though. What would you think about making a separate repo that people can use if they need multipart iteration?

Some feedback though:
- Usage examples would help a lot.
- Functions like current() and valid() should be idempotent so that they can be called multiple times in a row without causing a side-effect.
- Not sure if it's coded this way now, but it would be nice if each part did not require loading the entire part into memory.

FabioBatSilva:
Hi @mtdowling,

I'm OK making it a separate repo, 
but you guys already have a ```MultipartBody``` to create responses why not support reading it to ?

* I updated the iterator behaviour
* It reads each part as a stream
* Assuming you have a multipart response like :
* Usage example

```bash
curl -v http://127.0.0.1:8098/buckets/test/keys/doc -H "Accept: multipart/mixed"
* About to connect() to 127.0.0.1 port 8098 (#0)
*   Trying 127.0.0.1... connected
* Connected to 127.0.0.1 (127.0.0.1) port 8098 (#0)
> GET /buckets/test/keys/doc HTTP/1.1
> User-Agent: curl/7.19.4 (universal-apple-darwin10.0) libcurl/7.19.4 OpenSSL/0.9.8l zlib/1.2.3
> Host: 127.0.0.1:8098
> Accept: multipart/mixed
>
< HTTP/1.1 300 Multiple Choices
< X-Riak-Vclock: a85hYGDgyGDKBVIszMk55zKYEhnzWBlKIniO8kGF2TyvHYIKf0cIszUnMTBzHYVKbIhEUl+VK4spDFTPxhHzFyqhEoVQz7wkSAGLMGuz6FSocFIUijE3pt5HlsgCAA==
< Vary: Accept, Accept-Encoding
< Server: MochiWeb/1.1 WebMachine/1.9.0 (participate in the frantic)
< Date: Fri, 30 Sep 2011 15:24:35 GMT
< Content-Type: multipart/mixed; boundary=YinLMzyUR9feB17okMytgKsylvh
< Content-Length: 766
<

--YinLMzyUR9feB17okMytgKsylvh
Content-Type: application/x-www-form-urlencoded
Link: </buckets/test>; rel="up"
Etag: 16vic4eU9ny46o4KPiDz1f
Last-Modified: Wed, 10 Mar 2010 18:01:06 GMT

{"bar":"baz"}
--YinLMzyUR9feB17okMytgKsylvh
Content-Type: application/json
Link: </buckets/test>; rel="up"
Etag: 4v5xOg4bVwUYZdMkqf0d6I
Last-Modified: Wed, 10 Mar 2010 18:00:04 GMT

{"bar":"baz"}
--YinLMzyUR9feB17okMytgKsylvh
Content-Type: application/json
Link: </buckets/test>; rel="up"
Etag: 6nr5tDTmhxnwuAFJDd2s6G
Last-Modified: Wed, 10 Mar 2010 17:58:08 GMT

{"bar":"baz"}
--YinLMzyUR9feB17okMytgKsylvh
Content-Type: application/json
Link: </buckets/test>; rel="up"
Etag: 6zRSZFUJlHXZ15o9CG0BYl
Last-Modified: Wed, 10 Mar 2010 17:55:03 GMT

{"foo":"bar"}
--YinLMzyUR9feB17okMytgKsylvh--
* Connection #0 to host 127.0.0.1 left intact
* Closing connection #0
```

Guzzle :
```php
$client  = new \GuzzleHttp\Client();
$request = $client->createRequest('GET', 'http://127.0.0.1:8098/buckets/test/keys/doc');

$request->setHeader('Accept', 'multipart/mixed');

$response = $client->send($request);
$iterator = new \GuzzleHttp\Message\MultipartResponseIterator($response);
$list[]   = [];

/** @var $item \GuzzleHttp\Message\Response */
foreach ($iterator as $item) {
    $list[] = $item->json();
}

var_dump($list);
/*
array(4) {
  [0] =>
  array(1) {
    'bar' =>
    string(3) "baz"
  }
  [1] =>
  array(1) {
    'bar' =>
    string(3) "baz"
  }
  [2] =>
  array(1) {
    'bar' =>
    string(3) "baz"
  }
  [3] =>
  array(1) {
    'foo' =>
    string(3) "bar"
  }
}
*/
```

mtdowling:
It looks cool. I'd be happy to accept a pull request that links the Guzzle docs to your project.

The reason I don't think this belongs in the Guzzle repo is that Guzzle only supports building multipart/form-data request bodies so that it can support sending POST requests. Sending POST requests is an extremely common requirement of HTTP clients. Parsing mutlpart responses is not as common of a usecase, so I think it doesn't need to be in the main repo.

Issue #872, opened 2014-10-20
Fix missing/inconsistent docblocks


cleentfaar:
tl:dr;
Are you open to merging a PR containing fixed doc-blocks for all classes in the library?

Hey @mtdowling / contributors

First of all, thanks a lot for making this library, I've been using it in quite a few of my home- and workprojects now and enjoying it's easy-of-use a lot.

Working with it so often in my IDE (happens to be phpstorm), I have been stumbling on to something that got in my way a little too often; the lack or inconsistent use of phpdoc-blocks in the Guzzle classes. Especially in the public methods of the `Request`, `Response` and `Client` classes, which most users will probably need to access in their code in some way or another.

This makes using IDEs for writing Guzzle-related code less intuitive, as most IDEs rely on proper phpdocs for some of their auto-completion and error-detection strategies. Aside from that, I think these days everyone should doc-block (at least) their public methods and properties, but that may be open to a discussion which I do not mean to get into here :smile:.

On that note I was wondering if you were open to potentially merging a PR containing a big refactoring/addition of phpdocs in all classes of the guzzle library. 

Having all the classes properly documented would really improve my IDE's auto-completion functionality, as well as solve some false-positive error-traces that an IDE may find, such as wrong/missing `@return` statements making them lose track of what a value represents.

As there may be quite some work in figuring out the exact docblocks that are required for each part of the code, I hope any of you will let me know if this is something that is not open to contributions. If it's all good, I'll get a PR ready soon.

PR #960, opened 2015-01-16
Added some basic PHPDoc to Client class

cebe:
Without this, IDE autocompletion does not work correctly. At least in PHPStorm.

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -175,6 +175,13 @@ class Client implements ClientInterface
         return (string) $this->baseUrl;
     }
 
+    /**
+     * Creates a Request object
+     * @param string $method the HTTP verb e.g. GET, POST, PUT, ...
+     * @param string $url the request URL
+     * @param array $options for configuring the Request object.
+     * @return Message\Request|RequestInterface
+     */
     public function createRequest($method, $url = null, array $options = [])
     {
         $options = $this->mergeDefaults($options);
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
:-1: This is a duplicate and has already been rejected. You MUST NOT duplicate docblocks from interfaces. If phpstorm isn't working, then phpstorm is broken...

cebe:
Found the issue for this: https://youtrack.jetbrains.com/issue/WI-23395
What about adding 

```php
/**
 * @inheritDoc
 */
```

until the issue is fixed?

GrahamCampbell:
I don't think so. https://github.com/guzzle/guzzle/issues/872#issuecomment-59844746

mtdowling:
Thanks for handling this for me, @GrahamCampbell :)

I appreciate the PR @cebe, but I don't like the duplication, and using @inheritdoc is a misuse of the tag. PHPStorm people say they're working on... The only problem there is it is slated for v9 which means everyone will have to pay that bugfix.

cebe:
Thanks for the pointer to that issue. I fully agree that PHPStorm should fix the bug and support this.
Nevertheless  adding `@inheritdoc` does not hurt and makes the library more compatible and users more happy with it. I more often prefer to go the pracitcal way to make it work for everyone.
May I ask which IDE you are using?

PR #962, opened 2015-01-21
Adding integration tests for Digest authentication

outtersg:
(continuation of https://github.com/guzzle/guzzle/pull/918, that was marked unmodifiable by GitHub)

This intends to prepare Digest implementation, to prevent regressing from the current (cURL) implementation.

### DIFF
--- a/tests/IntegrationTest.php
+++ b/tests/IntegrationTest.php
@@ -120,4 +120,29 @@ class IntegrationTest extends \PHPUnit_Framework_TestCase
             $this->assertEquals(502, $e->getResponse()->getStatusCode());
         }
     }
+    
+    public function testDigestAuthWithBadPasswordDetectsRejection()
+    {
+        Server::enqueue([new Response(200)]);
+        $c = new Client();
+        try {
+            $response = $c->get(Server::$url . 'secure/by-digest/anything');
+        } catch(\GuzzleHttp\Exception\ClientException $e) {
+            if ($e->hasResponse()) {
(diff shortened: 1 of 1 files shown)


sagikazarmark:
As far as I can see this was opened against the 5.x series. Can you please resubmit it against the latest master?

sagikazarmark:
@outtersg Is this still a thing? I can see some digest auth tests in the 6.x release branch.

outtersg:
Uh, I would say that it can be abandonned. The tests were done to prevent regressions in case a pure-PHP Digest authentication would be implemented in Guzzle *; but neither I did take time to work on it, nor exist any plans in Guzzle to implement it AFAIK.

So unless I'm wrong on my last assertion, rejecting this pull request would make sense.

* One big advantage on cURL would have been to allow Digest persistence. curl_init() has no way to be told that "latest request was Digest with this given nonce, retry with the same one in case the server did not change". With the right combination of server options (I noticed it on Baikal CalDAV, which has a sabre/dav server layer), implementing this persistence on client's side would cut the retries by half:
A 3-requests sequence without persistence would be (C: client request; S: server response):
  C Basic/S 401, C Digest/S 200; C Basic/S 401, C Digest/S 200; C Basic/S 401, C Digest/S 200
With persistence client would retry with the same nonce, giving us:
  C Basic/S 401, C Digest/S 200; C Digest/S 200; C Digest/S 200

sagikazarmark:
Thanks for your work.

Issue #964, opened 2015-01-22
"Maximum function nesting level of '100' reached" on low number on parallel retries


paslandau:
This issue might be related to https://github.com/guzzle/guzzle/issues/905

I noticed that Guzzle will terminated with a "Maximum function nesting level of '100' reached" fatal error when some retries (via `AbstractRetryableEvent::retry()`) are made in parallel. Code to reproduce:

```php
$requestNum = 5;
$retryPerRequest = 5;

$c = new Client();

/* Bug is unfortunately not reproducable when using the Mock Subscriber */
//$responses = [];
//for($i = 0; $i < $requestNum*$retryPerRequest; $i++){
//    $responses[] = new Response(200);
//}
//$mock = new Mock($responses);
//$c->getEmitter()->attach($mock);

$rq = [];
for($i = 0; $i<$requestNum; $i++){
    $request = $c->createRequest("GET","http://example.org/");
    $request->getConfig()->set("id",$i);
    $rq[] = $request;
}

$i = 0;
$f = function(AbstractRetryableEvent $event)  use ($retryPerRequest){
    $retries = $event->getRequest()->getConfig()->get("retries");
    if($retries === null){
        $retries = 0;
    }
    if($retries < $retryPerRequest){
        $retries++;
        $event->getRequest()->getConfig()->set("retries",$retries);
        $id = $event->getRequest()->getConfig()->get("id");
        echo "Request $id, Retry $retries\n";
        $event->retry();
    }
};
$o = [
    "pool_size" => 25,
    "error" => $f,
    "complete" => $f,
    "end" => function(EndEvent $event) use (&$i){
        $id = $event->getRequest()->getConfig()->get("id");
        echo "Finished $id\n";
    }
];
$pool = new Pool($c,$rq,$o);
$pool->wait();
```

Result
```
Request 0, Retry 1
Request 1, Retry 1
Request 2, Retry 1
Request 3, Retry 1
Request 4, Retry 1
Request 0, Retry 2
Fatal error: Maximum function nesting level of '100' reached, aborting! in D:\xampp\htdocs\paslandau\vendor\guzzlehttp\guzzle\src\RequestFsm.php on line 146

Call Stack:
    0.0000     135408   1. {main}() D:\xampp\htdocs\paslandau\tests\guzzle-mass-retry-test.php:0
    0.0190    1177184   2. GuzzleHttp\Pool->wait() D:\xampp\htdocs\paslandau\tests\guzzle-mass-retry-test.php:59
[...]

PR #965, opened 2015-01-23
Updated the RequestFSM to use gotos.

mtdowling:
- This commit updates the request FSM to use goto statements to reduce
  function call overhead and removes 5 stack frames from each request.
- Removed `$finalState` from RequestFSM::__invoke
- Finishing FSM transitions is not handled in the FSM rather than the
  RingBridge.
- Slightly related to #964

@jeremeamia


jeremeamia:
Very interesting! `goto` for the win. :smile:

PR #970, opened 2015-01-28
Update index.rst

jpujari:
There is no class called GuzzleHttp

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -20,7 +20,7 @@ trivial to integrate with web services.
 
 .. code-block:: php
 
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Really?

GrahamCampbell:
Have you actually looked at the code in this repository? https://github.com/guzzle/guzzle/blob/master/src/Client.php

jpujari:
Sorry my bad, there are alternating examples on the internet on how to use guzzle, may be I was referring to an old article

http://www.sitepoint.com/guzzle-php-http-client/

PR #972, opened 2015-01-31
Explicitly show how to cast getBody() result to string.

k4ml:
Lot of tutorials still show example of:-

    $body = $response->getBody(true);

to get a string as a result. This confuse me a lot until I found that it [no longer apply](https://www.drupal.org/node/2401159) in recent version.

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -78,13 +78,16 @@ asynchronously using the promise interface of a future response.
 Response Body
 -------------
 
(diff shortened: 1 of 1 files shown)


mtdowling:
I don't think this is needed. A comment about a previous version just adds to the confusion I think, and documentation shows that it can be cast to a string which is thoroughly documented on php.net. I'd merge a PR that only showed casting the body to a string, but this PR has a lot of extraneous additions.

Issue #421, opened 2013-09-21
JSON decoding fails if response starts with BOM


cviebrock:
I ran into this issue with nategood/httpful#94 as well: the server I'm connecting to is including the BOM on the UTF-8 responses, which causes `json_decode()` to fail.

~~Any chance you can adapt the pull request I made for that project into Guzzle?~~

EDIT: Pull request below.

PR #977, opened 2015-02-11
json strip UTF-8 BOM

majkel89:
#421


mtdowling:
I don't think Guzzle needs BOM detection or removal built-in. Instead, I think this could be handled by your usage of the library such that you remove the BOM before using any of Guzzle or PHP's built-in JSON functionality.

PR #993, opened 2015-03-04
Last guzzle response

sngrl:
Adding method that return last guzzle response.


yamaha252:
```
$history = new History();
$client = new Client()->getEmitter()->attach($history);
...
$lastRequest = $history->getLastRequest();
```

mtdowling:
@yamaha252 has provided the correct way to achieve this.

PR #994, opened 2015-03-05
Add Travis-CI check for PHP 7.0

SammyK:
Added Travis-CI check for PHP 7.0 since that's [gonna be thing soon](https://wiki.php.net/rfc/php7timeline#proposal). I've got `allow_failures` on so it won't cause the whole build to fail. Spoiler-alert: [the 7.0 build fails](https://travis-ci.org/facebook/facebook-php-sdk-v4/jobs/53147975#L187). :) I inadvertently discovered this about Guzzle when [checking against PHP 7.0 in the Facebook PHP SDK](https://github.com/facebook/facebook-php-sdk-v4/pull/375).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ php:
   - 5.4
   - 5.5
   - 5.6
+  - 7.0
   - hhvm
 
 before_script:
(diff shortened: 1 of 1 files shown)


mtdowling:
Thanks!

PR #1004, opened 2015-03-13
Update quickstart.rst

flyingbuddha:
GuzzleHttp namespace wrong, should be Guzzle\Http

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -9,7 +9,7 @@ page.
 Make a Request
 ==============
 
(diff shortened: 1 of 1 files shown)


mtdowling:
The namespace was changed in v4.0 of Guzzle.

flyingbuddha:
@mtdowling Ah, my bad, I've just come across this. It's not obvious from the github namespace to use guzzlehttp/guzzle (even though admittedly it does say in the documentation the correct one)

PR #1012, opened 2015-03-20
Added some helper methods to the response and added tests for those methods

jcakes69:


mtdowling:
Thanks, but I don't think these methods are necessary for Guzzle's response objects.

jcakes69:
Sounds good, they were helper methods that were used in the Zend/HTTP/Response that we used quite a bit so I was hoping to get them into Guzzle since we were looking to switch over to that soon.
http://framework.zend.com/manual/current/en/modules/zend.http.response.html

PR #1016, opened 2015-03-29
Fixed a typo

GrahamCampbell:
Also, I'm not sure if you're even allowed to alias branches that are already valid versions. Maybe name this branch "6.0" and ditch the alias to avoid this issue?


GrahamCampbell:
Ping @mtdowling.

PR #1020, opened 2015-04-03
Fix save_to for Mock

ivank:
Great project, just a small fix for an issue I ran into while using your library - one cannot use "save_to" and write proper tests for it at the moment. 

This allows using “save_to” in mocked requests. Now you can write tests for code that uses “save_to” easily.
Test included.

### DIFF
--- a/src/Subscriber/Mock.php
+++ b/src/Subscriber/Mock.php
@@ -60,6 +60,12 @@ class Mock implements SubscriberInterface, \Countable
             }
         }
 
+        $save_to = $event->getRequest()->getConfig()->get('save_to');
+
+        if ($save_to) {
+            file_put_contents($save_to, $item->getBody());
+        }
+
         $event->intercept($item);
     }
 
--- a/src/Subscriber/Mock.php
+++ b/src/Subscriber/Mock.php
@@ -7,6 +7,7 @@ use GuzzleHttp\Event\BeforeEvent;
 use GuzzleHttp\Exception\RequestException;
 use GuzzleHttp\Message\MessageFactory;
 use GuzzleHttp\Message\ResponseInterface;
+use GuzzleHttp\Stream\StreamInterface;
 
 /**
  * Queues mock responses or exceptions and delivers mock responses or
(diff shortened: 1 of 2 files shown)


src/Subscriber/Mock.php
```diff
@@ -60,6 +60,12 @@ public function onBefore(BeforeEvent $event)
             }
         }
 
+        $save_to = $event->getRequest()->getConfig()->get('save_to');
+
+        if ($save_to) {
+            file_put_contents($save_to, $item->getBody());
```
mtdowling: save_to can be of various types: http://docs.guzzlephp.org/en/latest/clients.html#save-to. I think this method would need to account for that.

ivank: Yeah, sorry about that, totally forgot. here's the updated PR with all the 3 use cases

src/Subscriber/Mock.php
```diff
@@ -60,6 +61,20 @@ public function onBefore(BeforeEvent $event)
             }
         }
 
+        $save_to = $event->getRequest()->getConfig()->get('save_to');
```
GrahamCampbell: Shouldn't the variable be called `$saveTo`?

ivank: I think scrutinizer could be configured to spot (and create patches for) these kinds of style violations automatically. Anyway fixed.

src/Subscriber/Mock.php
```diff
@@ -60,6 +61,20 @@ public function onBefore(BeforeEvent $event)
...
+
+        if (null !== $saveTo) {
+        	$body = $item->getBody();
+
+        	if (is_resource($saveTo)) {
+        		fwrite($saveTo, $body);
+        	} elseif (is_string($saveTo)) {
+	            file_put_contents($saveTo, $body);
+        	} elseif ($saveTo instanceof StreamInterface) {
+        		$saveTo->write($body);
```
mtdowling: Can you fix the indentation here to match the rest of the library (4 spaces)?

mtdowling:
Looks good! Just need a minor tweak to fix the spacing, but other than that, LGTM.

Issue #1024, opened 2015-04-10
Path with reserved characters does not return percent encoding (PSR-7 violation)


Maks3w:
I found a possible double issue with the management of encoded/unencoded paths.

Given this path `/email/foo+extension@example.com` (an email address as path segment)

the following snippet return wrong results.

```php
$request = new Request('GET', 'http://example.com/email/foo+extension@example.com');

echo $request->getPath(); /* /email/foo+extension@example.com */
```

### First error: `getPath` *MUST* return `percent-encoded` values.

>  * The value returned MUST be percent-encoded, but MUST NOT double-encode
>  * any characters. To determine what characters to encode, please refer to
>  * RFC 3986, Sections 2 and 3.3.
> 
> https://github.com/php-fig/fig-standards/blob/17945666b0b5c78691dca65405aae21e3e403539/proposed/http-message.md

So `$request->getPath();` *MUST* return `/email/foo%2Bextension%40example.com` (characters `+` and `@` percent-ecoded`

### Second issue: `+`, `@` and the rest of reserved words if present in the path are interpreted as already encoded

What's the problem? IMO is the same problem fixed on the release 5.0.3 with the query parameters encoding

mtdowling:
Guzzle v5 will not change this behavior.

PR #1025, opened 2015-04-10
[BC\Break][PSR-7] Apply percent-encoding to paths with not percent-encoding

Maks3w:
Following the issue #1024 (1K) `getPath()` are not PSR-7 compliant.

PSR-7 only allows `percent-encoding` paths as only possible encoding in `getPath()` method

The PR is breaks the BC because the output changes. However the new outputs should still working transparently on previous implementatins.

### DIFF
--- a/src/Url.php
+++ b/src/Url.php
@@ -16,7 +16,7 @@ class Url
     private $path = '';
     private $fragment;
     private static $defaultPorts = ['http' => 80, 'https' => 443, 'ftp' => 21];
-    private static $pathPattern = '/[^a-zA-Z0-9\-\._~!\$&\'\(\)\*\+,;=%:@\/]+|%(?![A-Fa-f0-9]{2})/';
+    private static $pathPattern = '/[^a-zA-Z0-9\-\._~!\/%]+|%(?![A-Fa-f0-9]{2})/';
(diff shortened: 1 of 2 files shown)


Maks3w:
Notice that paths with whitespaces encoded with `+` character won't be unencoded to whitespace, instead have to be unencoded twice.

mtdowling:
Just glancing over this, I see that ";" is being percent encoded in the path now. That's an allowed character in a path so this is incorrect I think. IIRC, all of the test cases were yanked from RFC 3896, so they shouldn't need to change (section 5.1 of RFC 3986). If PSR-7's wording conflicts with RFC 3986, then it needs to change.

Maks3w:
Sorry, your memory is wrong ;-)

> reserved    = gen-delims / sub-delims
> gen-delims  = ":" / "/" / "?" / "#" / "[" / "]" / "@"
> sub-delims  = "!" / "$" / "&" / "'" / "(" / ")" / "*" / "+" / "," / ";" / "="

mtdowling:
This wont change in Guzzle v5, and I don't think this change is valid according to RFC 3986. We took the tests that have been modified in this PR directly from RFC 3986.

PR #1038, opened 2015-04-27
Make CookieJar->cookies usable by derivatives

JeroenDeDauw:
Advantages:

* Better performance, since the not needed array_values call
and ArrayIterator construction are skipped
* Static analysis tools now understand the type of $cookie
in the loops

Disadvantage:

* Encapsulation decrease


mtdowling:
Sorry, I'd rather keep this private.

JeroenDeDauw:
Thanks for looking at this. I'd be interested in how you decided on this trade-of. If you have the time of course.

PR #1041, opened 2015-04-27
Workaround for large integers in JSON decoding

yuksbg:
Quick workaround of problem with large integers in JSON, instead of giving
json_decode(): integer overflow detected.

Demo int: 18425396933527319036

Problem:
image
"Solution":
image

### DIFF
--- a/src/Utils.php
+++ b/src/Utils.php
@@ -138,7 +138,8 @@ final class Utils
             JSON_ERROR_SYNTAX => 'JSON_ERROR_SYNTAX - Syntax error, malformed JSON',
             JSON_ERROR_UTF8 => 'JSON_ERROR_UTF8 - Malformed UTF-8 characters, possibly incorrectly encoded'
         ];
-
+		
+		$json = preg_replace ('/:\s?(\d{14,})/', ': "${1}"', $json);		
         $data = \json_decode($json, $assoc, $depth, $options);
 
         if (JSON_ERROR_NONE !== json_last_error()) {


kamilsk:
-1

How about JSON_BIGINT_AS_STRING ?

first: integer overflow is different on x32 and x64
on x64 integer overflow begins with 9223372036854775808 (19 characters, not 14)
```php
$a = '{"a":9223372036854775808}';
$b = json_decode($a);
var_dump($b);
~~~
class stdClass#2 (1) {
  public $a =>
  double(9.2233720368548E+18)
}

$a = '{"a":18425396933527319036}';
$b = json_decode($a);
~~~
var_dump($b);
class stdClass#1 (1) {
  public $a =>
  double(1.8425396933527E+19)
}
```

second: depending on the version of the json extension behavior is different
https://github.com/remicollet/pecl-json-c/issues/13

yuksbg:
it's just quick workaround, and JSON_BIGINT_AS_STRING is availible at php 5.4+

kamilsk:
https://github.com/guzzle/guzzle/blob/master/composer.json#L16

kamilsk:
@yuksbg, it is a good trick, but not as a system solution.

kamilsk:
Just my opinion.

mtdowling:
I think this is too specific to be in Guzzle and JSON_BIGINT_AS_STRING is in PHP 5.4, which Guzzle currently targets. However, you could do this outside of Guzzle and just not use the `json()` method of the response.

kamilsk:
@yuksbg just look at source code https://github.com/guzzle/guzzle/blob/master/src/Message/Response.php#L145
`isset($config['big_int_strings']) ? JSON_BIGINT_AS_STRING : 0`

PR #1047, opened 2015-04-30
Allow emitter be overwrited

AlloVince:
Hi Guys

I am working on a OAuth project which based on guzzle https://github.com/AlloVince/EvaOAuth

In my project, I want to extend GuzzleHttp\Client and replace emitter into mine own.

Currently the emitter of  GuzzleHttp\Client is private and not allow to overwrite, please consider my pull request, I think a replaceable is better.

Best regards


mtdowling:
You can do this in the constructor of the client: https://github.com/guzzle/guzzle/blob/master/src/Client.php#L61. No need to make this protected.

AlloVince:
You are right, thanks for the solution.

PR #1050, opened 2015-05-06
Updated copyright to 2015

maqnouch:


GrahamCampbell:
Should still be 2014. It needs to be the earliest year.

GrahamCampbell:
Alternatively 2014-2015 could be used.

maqnouch:
indeed, thank you for correcting me :)

PR #1056, opened 2015-05-13
Fix typo "GuzzleHttp" to "Guzzle\Http"

PatchRanger:


PatchRanger:
Sorry, I haven't understood at first that `guzzle/guzzle` Github project uses `GuzzleHttp` Packagist package and therefor the same namespace.

PR #1059, opened 2015-05-19
added trim BOM (for some response from bad .Net)

andboson:


kamilsk:
like https://github.com/guzzle/guzzle/pull/1041

-1

andboson:
@kamilsk what about bom in beginning of file?

and preg...always should find to how use string functions instead regexp

kamilsk:
Presence of the BOM at the beginning of the file is a special case (specific for "bad .Net").
You can remove it like this.
```php
$responseBody = json_decode((string) $response->getBody(), ...options);
```
Special cases will be resolved privately. IMO

andboson:
I want use existing function jsonDecode with error handling, etc.

kamilsk:
You can do this:
```php
try {
    $responseBody = \GuzzleHttp\Utils::jsonDecode(str_replace("\xEF\xBB\xBF", '', (string) $response->getBody()), ...options);
} catch (...handle required exception) {
}
```

kamilsk:
or you want to use `$response->json()` ?

andboson:
yep

kamilsk:
@andboson, how about event listeners? just modify body (remove bom) after success response and use `->json` you want.

PR #1067, opened 2015-05-23
add .gitattributes

kamilsk:


.gitattributes
```diff
@@ -0,0 +1,9 @@
+build/*          export-ignore
+docs/*           export-ignore
+tests/*          export-ignore
+.editorconfig    export-ignore
+.gitattributes   export-ignore
+.gitignore       export-ignore
+.travis          export-ignore
```
Aliance: There is no `.travis` file, only `.travis.yml` – is this a bug or just `.travis` consumed as file prefix, so `.travis.yml` suits this pattern?

kamilsk: this is a mistake, you are right @Aliance

kamilsk: Yes, you can fix it. Thank you!

Aliance: #1649

Issue #1075, opened 2015-05-26
Invalid response Content-Encoding and Content-Length headers when the body is decoded


ajgarlag:
If the `decode_content` option is set, and the upstream response body is decoded, the `Content-Length` header of the returned response must be fixed.

mtdowling:
Why does it need to change?

You can use the `$response->getBody()->getSize()` to get the actual number of bytes you have available. The `Content-Length` header is the actual Content-Length header received over the wire.

ajgarlag:
I'm writing a simple PSR-7 proxy script that listens for the user request, forwards the received request to the upstream server with guzzle, and returns the guzzle response to the client.

In this scenario, the guzzle response must be valid to be successfully rendered by the user-agent, so when the body content is decoded, I think that at least `Content-Length` and `Content-Encoding` must be fixed.

mtdowling:
Can you provide some data as to what other popular HTTP clients do with regards to updating or removing the Content-Length header when the body is decoded?

ajgarlag:
I've found this two middleware examples. Their mission is not to decode the upstream response, but to compress the downstream body, but the idea is the same. If a middleware modifies the body encoding or length, it is responsible for fixing the response headers to avoid returning an invalid response.

* [Django](https://github.com/django/django/blob/master/django/middleware/gzip.py#L30): Gzip Middleware
* [Compression](https://github.com/expressjs/compression/blob/master/index.js#L193): Node.js compression middleware.

mtdowling:
Your PR has been merged. Thanks!

PR #1076, opened 2015-05-26
Fix Content-Encoding and Content-Length headers when the response body is decoded

ajgarlag:
When the response body is decoded, the `Content-Length` of the returned response must be equal to the decoded body length, or removed if not known and the `Content-Encoding` header must be absent.

See #1075

### DIFF
--- a/src/Handler/EasyHandle.php
+++ b/src/Handler/EasyHandle.php
@@ -51,10 +51,24 @@ final class EasyHandle
         // HTTP-version SP status-code SP reason-phrase
         $startLine = explode(' ', array_shift($this->headers), 3);
 
+        $headers = \GuzzleHttp\headers_from_lines($this->headers);
+        $normalizedKeys = array_combine(array_map('strtolower', array_keys($headers)), array_keys($headers));
+        if (!empty($this->options['decode_content']) && isset($normalizedKeys['content-encoding'])) {
+            unset($headers[$normalizedKeys['content-encoding']]);
+            if (isset($normalizedKeys['content-length'])) {
+                $bodyLength = (int) $this->sink->getSize();
+                if ($bodyLength) {
+                    $headers[$normalizedKeys['content-length']] = $bodyLength;
+                } else {
+                    unset($headers[$normalizedKeys['content-length']]);
+                }
+            }
+        }
+
         // Attach a response to the easy handle with the parsed headers.
         $this->response = new Response(
             $startLine[1],
-            \GuzzleHttp\headers_from_lines($this->headers),
+            $headers,
             $this->sink,
             substr($startLine[0], 5),
             isset($startLine[2]) ? (int) $startLine[2] : null
--- a/src/Handler/StreamHandler.php
+++ b/src/Handler/StreamHandler.php
@@ -66,7 +66,8 @@ class StreamHandler
         $status = $parts[1];
         $reason = isset($parts[2]) ? $parts[2] : null;
         $headers = \GuzzleHttp\headers_from_lines($hdrs);
-        $stream = Psr7\stream_for($this->checkDecode($options, $headers, $stream));
+        list ($stream, $headers) = $this->checkDecode($options, $headers, $stream);
+        $stream = Psr7\stream_for($stream);
         $sink = $this->createSink($stream, $options);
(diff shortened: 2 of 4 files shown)


ajgarlag:
@mtdowling Could you review this PR?

src/Handler/StreamHandler.php
```diff
@@ -106,18 +107,30 @@ private function checkDecode(array $options, array $headers, $stream)
     {
         // Automatically decode responses when instructed.
         if (!empty($options['decode_content'])) {
-            foreach ($headers as $key => $value) {
-                if (strtolower($key) == 'content-encoding') {
-                    if ($value[0] == 'gzip' || $value[0] == 'deflate') {
-                        return new Psr7\InflateStream(
-                            Psr7\stream_for($stream)
-                        );
+            $normalizedKeys = array_combine(array_map('strtolower', array_keys($headers)), array_keys($headers));
```
mtdowling: If this actually needs to change, then may I recommend array_change_key_case()? It would do less work to convert the keys to lowercase, and the only value that needs to be lowercased in `content-encoding` (so you don't need to lowercase all the values).

mtdowling:
Is there any reason why this change was not applied to the curl handler as well?

ajgarlag:
I've applied the change to the curl handler too.

src/Handler/StreamHandler.php
mtdowling: Can you modify this to use `array_change_key_case` here and in the curl handler code? http://php.net/manual/en/function.array-change-key-case.php

src/Handler/StreamHandler.php
```diff
@@ -106,18 +107,30 @@ private function checkDecode(array $options, array $headers, $stream)
...
+
+            if (isset($normalizedKeys['content-encoding'])) {
+                $encoding = $headers[$normalizedKeys['content-encoding']];
```
mtdowling: After the `array_change_key_case` change, modify this to be something like `$encoding = strtolower($lcheaders['content-encoding'])`

src/Handler/StreamHandler.php
```diff
@@ -106,18 +107,30 @@ private function checkDecode(array $options, array $headers, $stream)
...
+                if ($encoding[0] == 'gzip' || $encoding[0] == 'deflate') {
+                    $stream = new Psr7\InflateStream(
+                        Psr7\stream_for($stream)
+                    );
+                    // Remove content-encoding header
+                    unset($headers[$normalizedKeys['content-encoding']]);
+                    // Fix content-length header
+                    if (isset($normalizedKeys['content-length'])) {
+                        $length = (int) $stream->getSize();
+                        if ($length == 0) {
```
mtdowling: should be a null check as a `0` length size is valid.

src/Handler/StreamHandler.php
```diff
@@ -106,18 +107,30 @@ private function checkDecode(array $options, array $headers, $stream)
...
+                        } else {
+                            $headers[$normalizedKeys['content-length']] = array($length);
+                        }
                     }
                 }
             }
         }
 
-        return $stream;
+        return array($stream, $headers);
```
mtdowling: Can you use the short array syntax please? `[]`

src/Handler/StreamHandler.php
ajgarlag: I'm not sure if I understand you. Your recommendation is to change this line to:
```php
$normalizedKeys = array_combine(array_keys(array_change_key_case($headers))), array_keys($headers));
```
or to change it to:
```php
$lcheaders = array_change_key_case($headers));
```
and then, return the `$lcheaders` array overwriting the original headers case?

src/Handler/EasyHandle.php
```diff
@@ -51,10 +51,24 @@ public function createResponse()
         // HTTP-version SP status-code SP reason-phrase
         $startLine = explode(' ', array_shift($this->headers), 3);
 
+        $headers = \GuzzleHttp\headers_from_lines($this->headers);
+        $normalizedKeys = array_combine(array_map('strtolower', array_keys($headers)), array_keys($headers));
```
ajgarlag: Here, we can leave the original headers, construct the response with the original code, and fix the headers of the built response, taking advantage of the `withHeader` method which is case insensitive.

What do you think?

PR #1088, opened 2015-05-28
Update testing docs about History middleware

eXtreme:
I think this should be the correct order of instructions in the code. ;) Also, there is `HandlerStack::push`, not `::append`.


mtdowling:
thanks!

PR #1089, opened 2015-05-28
Updated UPGRADING guide about Mock Subscriber

eXtreme:
UPGRADING guide says there is Mock Middleware as a replacement for Mock Subscriber which is incorrect because we have `MockHandler`.


mtdowling:
Thanks!

PR #1096, opened 2015-05-29
Allow custom log levels for logging middleware.

jaredm4:
This would be great as I generally prefer to log these to debug instead of info. Didn't want to duplicate all that logic for my own middleware, however.

### DIFF
--- a/src/Middleware.php
+++ b/src/Middleware.php
@@ -9,6 +9,7 @@ use GuzzleHttp\Promise\RejectedPromise;
 use GuzzleHttp\Psr7;
 use Psr\Http\Message\ResponseInterface;
 use Psr\Log\LoggerInterface;
+use Psr\Log\LogLevel;
 
 /**
  * Functions used to create and wrap handlers with handler middleware.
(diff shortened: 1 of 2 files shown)


src/Middleware.php
```diff
@@ -176,17 +177,18 @@ public static function retry(callable $decider, callable $delay = null)
...
-        return function (callable $handler) use ($logger, $formatter) {
-            return function ($request, array $options) use ($handler, $logger, $formatter) {
+        return function (callable $handler) use ($logger, $formatter, $logLevel) {
+            return function ($request, array $options) use ($handler, $logger, $formatter, $logLevel) {
                 return $handler($request, $options)->then(
-                    function ($response) use ($logger, $request, $formatter) {
+                    function ($response) use ($logger, $request, $formatter, $logLevel) {
                         $message = $formatter->format($request, $response);
-                        $logger->info($message);
+                        $logger->$logLevel($message);
```
kamilsk: don't like this part. how about `$logger->log($loggerLevel, ...)` ?

jaredm4: Great suggestion. Sure.

mtdowling:
Thanks!

PR #1098, opened 2015-05-30
Replace form_files with multipart in docs

konforce:
Replace reference to ``form_files`` with ``multipart``.
Add ``filename`` key to the second file in example.


mtdowling:
Thanks!

PR #1101, opened 2015-05-31
Add requestFactory method

### LINKED ISSUE #1100: Missing request factory function
I would really like a way to create an instance of `RequestInterface` using the form:

```php
$request = $guzzle->request($method, $uri, $options);
```

Right now this call will create a request __and__ send it. I want all the functionality of the current `request` without the sending part, so that I can later do:

```php
$response = $guzzle->send($request);
```

This is exacerbated by the fact that `GuzzleHttp\Psr7\modify_request` does not have the same functionality as `GuzzleHttp\Client::applyOptions`, which does several more transformations (eg: converting `form_params` to `body`) than just calling `modify_request`.

shadowhand:
Fixes #1100

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -105,9 +105,8 @@ class Client implements ClientInterface
         return $this->sendAsync($request, $options)->wait();
     }
 
-    public function requestAsync($method, $uri = null, array $options = [])
+    public function requestFactory($method, $uri = null, array $options = [])
     {
-        $options = $this->prepareDefaults($options);
         // Remove request modifying parameter because it can be done up-front.
         $headers = isset($options['headers']) ? $options['headers'] : [];
         $body = isset($options['body']) ? $options['body'] : null;
@@ -120,6 +119,14 @@ class Client implements ClientInterface
         $request = new Psr7\Request($method, $uri, $headers, $body, $version);
         // Remove the option so that they are not doubly-applied.
         unset($options['headers'], $options['body'], $options['version']);
+
+        return $this->applyOptions($request, $options);
+    }
+
+    public function requestAsync($method, $uri = null, array $options = [])
+    {
+        $request = $this->requestFactory($method, $uri, $options);
+        $options = $this->prepareDefaults($options);
 
         return $this->transfer($request, $options);
     }
@@ -245,7 +252,7 @@ class Client implements ClientInterface
      *
      * @return Promise\PromiseInterface
      */
-    private function transfer(RequestInterface $request, array $options)
+    private function transfer(RequestInterface $request, array $options = [])
     {
         // save_to -> sink
         if (isset($options['save_to'])) {
--- a/src/Client.php
+++ b/src/Client.php
@@ -99,35 +99,42 @@ class Client implements ClientInterface
         );
     }
 
+    public function requestFactory($method, $uri = null, array $options = [])
+    {
+        $request = $this->prepareRequest($method, $uri, $options);
+
+        return $this->applyOptions($request, $options);
+    }
+
     public function send(RequestInterface $request, array $options = [])
     {
         $options[RequestOptions::SYNCHRONOUS] = true;
         return $this->sendAsync($request, $options)->wait();
     }
 
-    public function requestFactory($method, $uri = null, array $options = [])
+    private function prepareRequest($method, $uri = null, array &$options = [])
     {
-        // Remove request modifying parameter because it can be done up-front.
+        $options = $this->prepareDefaults($options);
         $headers = isset($options['headers']) ? $options['headers'] : [];
+
         $body = isset($options['body']) ? $options['body'] : null;
-        $version = isset($options['version']) ? $options['version'] : '1.1';
-        // Merge the URI into the base URI.
-        $uri = $this->buildUri($uri, $options);
         if (is_array($body)) {
             $this->invalidBody();
         }
-        $request = new Psr7\Request($method, $uri, $headers, $body, $version);
-        // Remove the option so that they are not doubly-applied.
-        unset($options['headers'], $options['body'], $options['version']);
 
-        return $this->applyOptions($request, $options);
+        // Merge the URI into the base URI.
+        $uri = $this->buildUri($uri, $options);
+
+        $version = isset($options['version']) ? $options['version'] : '1.1';
+        return new Psr7\Request($method, $uri, $headers, $body, $version);
     }
 
     public function requestAsync($method, $uri = null, array $options = [])
     {
-        $request = $this->requestFactory($method, $uri, $options);
-        $options = $this->prepareDefaults($options);
+        $request = $this->prepareRequest($method, $uri, $options);
 
+        // Remove the option so that they are not doubly-applied.
+        unset($options['headers'], $options['body'], $options['version']);
         return $this->transfer($request, $options);
     }
 
@@ -252,7 +259,7 @@ class Client implements ClientInterface
      *
      * @return Promise\PromiseInterface
      */
-    private function transfer(RequestInterface $request, array $options = [])
+    private function transfer(RequestInterface $request, array $options)
     {
         // save_to -> sink
         if (isset($options['save_to'])) {
--- a/tests/ClientTest.php
+++ b/tests/ClientTest.php
@@ -493,4 +493,53 @@ class ClientTest extends \PHPUnit_Framework_TestCase
         $client->send($request, ['query' => ['foo' => 'bar', 'john' => 'doe']]);
         $this->assertEquals('foo=bar&john=doe', $mock->getLastRequest()->getUri()->getQuery());
     }
+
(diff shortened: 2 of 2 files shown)


shadowhand:
@GrahamCampbell so this doesn't work exactly right, but it shows my intent. Should I keep working on it?

GrahamCampbell:
> Should I keep working on it?

Don't ask me, ask @mtdowling.

rtheunissen:
@shadowhand How does it not work quite right?

shadowhand:
It causes errors because of the way the options are applied. Some of them are only meant to be applied late (just before send) and some can be applied early. Because the current assumption is that options are _always_ applied just before send, it fails.

shadowhand:
@mtdowling thanks to @rtheunissen this is functional now.

rtheunissen:
:+1:

mtdowling:
I like the premise of this change. How would you use this function in your library? Would you require that the client passed in is a concrete `Client` class? I wonder if it would be better to make this behavior a static method on the Client like `createRequest()`.

rtheunissen:
> How would you use this function in your library?

I'm discussing some details with @shadowhand at the moment but will get back to this as soon as we have some concrete. It's safe to say that we would like to use `requestFactory` **internally only**, ie. not expose the Guzzle-specific options (query, form_params etc). We would like to use the client for internal requests too, but in the scope of this pull request, as a *Request preparer / configurator / generator*, producing a PSR-7 request which can be sent by any PSR-7 compliant client. 

A large part of the Guzzle client is responsible for converting request options into query parameters, uri, body etc. We would like to avoid having to do all that ourselves, and seeing as Guzzle is already doing that in a well-supported way, all we're hoping to achieve here is a hook into the pipeline between when options are applied and the request being dispatched.   

> Would you require that the client passed in is a concrete Client class?

Honestly I think that is our best option, but outside of the scope of this request. Let's assume that we somehow have a Guzzle 6 client on hand; how we got it doesn't matter.

> I wonder if it would be better to make this behavior a static method on the Client like createRequest()

Hmm probably could. Are there any options set on the Client (not the Response) that can affect a Response during its preparation before being dispatched?

shadowhand:
> How would you use this function in your library?

You can see exactly how it would be used in thephpleague/oauth2-client#315.

> Would you require that the client passed in is a concrete Client class?

We type hint against `ClientInterface` as a constructor (and setter) option. So we need to allow the end user to pass in their own implementation of Guzzle's Client, if they choose to do so.

> I wonder if it would be better to make this behavior a static method on the Client like createRequest()

No, not for our purposes. From pretty much every angle, having a static method would make it more difficult to use.

mtdowling:
> No, not for our purposes. From pretty much every angle, having a static method would make it more difficult to use.

Why is that? I don't think his method belongs on ClientInterface (and adding it would be a breaking change for anyone who already implemented this interface), so that limits the clients that can be passed into your class as only instances or subclasses of Client. By making the request creation function static, you can use that method and not care what concrete client is passed to your class. I think that coupling to the static method would be better than coupling to a specific client class.

shadowhand:
> By making the request creation function static, you can use that method and not care what concrete client is passed to your class.

This is exactly why it is undesirable. It prevents an implemented `ClientInterface` from creating a decorated request. If we _always_ call `Guzzle\Client::requestFactory` we will _always_ get `Guzzle\Psr7\Request` even if the overloaded `Guzzle\Client` returned, say:

```php
class AcmeRequest extends Guzzle\Psr7\Request { ... }
```

>  I don't think his method belongs on ClientInterface

As we've just seen, it absolutely should exist in `ClientInterface`. It's the only way that a client could decorate the default Request object.

> By making the request creation function static, you can use that method and not care what concrete client is passed to your class.

But we __do__ care, and we should. We want to respect Liskov Substitution Principle, otherwise we are violating dependency inversion and preventing extension.

rtheunissen:
> It's the only way that a client could decorate/replace/extend the default Request object.

Not true. If you extend Client rather than implement ClientInterface, you can still override the factory. That does mean that we can't assume that a ClientInterface provides a factory, which is not what we want. 

I agree with @shadowhand that this should be on the interface, and not the client.

If we had an abstract client we could have added default behaviour without breaking BC and type hint abstract class instead, but because we have an interface it's the only way.

mtdowling:
This change seems suboptimal to me because `requestFactory` would only be used when creating requests. It would create a PSR-7 request and apply options to the created request. However, when using `send()` or `sendAsync()`, you provide the request being sent but you can also provide request options to apply to the request. However, the `requestFactory` function would not be invoked. This reduces the usefulness of adding `requestFactory` in terms of decorating a client.

I wonder if a better approach would be to add a `prepareRequest(RequestInterface $request, array $options)` method that modifies a request with request options. This would be used for both `send()` and `request()` and could be another extension point outside of the middleware system. `prepareRequest` would basically be `applyOptions`, but `$options` would not be passed in by reference. Removing keys from the options array isn't *really* necessary, so I'm ok with losing that. Any other mutation (i.e., adding curl options) could be moved out of that method into another method.

With this approach, you'd need to create a PSR-7 request in your library. I don't really see that as an issue if you're abstracting Guzzle away as an implementation detail. For example, you could allow users to inject a PSR-7 request factory into your client.

There's still the issue with adding a new method to an interface is a sort of breaking change... In fact Symfony for example does not allow it on "@api" classes: http://symfony.com/doc/2.7/contributing/code/bc.html. How do you propose to work around this?

shadowhand:
> There's still the issue with adding a new method to an interface is a sort of breaking change...

It's not. According to the Symfony docs:

> Minor releases (such as 2.5, 2.6 etc.) may introduce new features, but must do so without breaking the existing API of that release branch 

Adding a new method does not break anything, it only adds new functionality. The next release of Guzzle that could include these changes would be 6.1 and everything would be acceptable, unless I completely misunderstand something about semver.

> However, the requestFactory function would not be invoked. This reduces the usefulness of adding requestFactory in terms of decorating a client.

This statement is only partially true. While the `requestFactory` itself is not called by existing code, the _new_ `prepareRequest` method is, which is effectively the same thing, only with better abstraction. The bulk of the that method could be inside of `requestFactory`, but I feel the current abstraction is more flexible for future changes.

> With this approach, you'd need to create a PSR-7 request in your library. 

If we were okay with doing that, we wouldn't need this PR. The current implementation is exactly what we want because:

1. Our dependency is only on `ClientInterface` and PSR-7 interfaces.
2. There is no hard dependency on a specific implementation of `RequestInterface`.

If we were to create the request ourselves, we would have *two* hard dependencies, one of which could not be decorated because it wouldn't be created by `Client`. The current implementation allows returning any type of `RequestInterface`, whereas having `new Guzzle\Psr7\Request` would increase coupling.

> For example, you could allow users to inject a PSR-7 request factory into your client.

We would prefer that Guzzle be usable as a Request factory so that we don't need to create more dependencies. :sunglasses:

mtdowling:
> It's not. According to the Symfony docs:

Did you see the interfaces section? http://symfony.com/doc/2.7/contributing/code/bc.html#changing-interfaces. Look at the API column for the "Add method" row. It says they don't allow it.

And yeah, I think it's a breaking change (albeit somewhat minor) because someone who already created a client will now need to add a new method or their code will break. Semver says that you can't break the public API unless it's a major version bump. It sounds like you don't think it's a big deal, but I just want to be cautious.

> This statement is only partially true. While the requestFactory itself is not called by existing code, the new prepareRequest method is, which is effectively the same thing, only with better abstraction. The bulk of the that method could be inside of requestFactory, but I feel the current abstraction is more flexible for future changes.

Can you elaborate? I don't understand what point you're making.

> If we were to create the request ourselves, we would have two hard dependencies, one of which could not be decorated because it wouldn't be created by Client. The current implementation allows returning any type of RequestInterface, whereas having new Guzzle\Psr7\Request would increase coupling.

So allow your end user to inject a factory and if they do not inject a factory, create a request using Guzzles PSR-7 classes. You know what your dependency graph is because you're using Guzzle.

rtheunissen:
What if created an AbstractClient which implements ClientInterface, and make requestFactory the only abstract function, and have Client extend AbstractClient with a default version of requestFactory, and we then type hint / depend on AbstractClient instead. That way all current implementations of ClientInterface are still compatible, and we get the contract we are after.

shadowhand:
> Did you see the interfaces section?

I missed that. Which leaves things in a very unfortunate place, since Guzzle 5 had `createRequest` and 4 had multiple methods to create requests and send them later. Hopefully this will be incorporated into Guzzle 7 API.

> So allow your end user to inject a factory and if they do not inject a factory, create a request using Guzzles PSR-7 classes.

Unfortunately this is the approach we'll have to take, due to the limitations of the existing interface. I'm disappointed with this solution, but I see no other path forward at this point.

GrahamCampbell:
Adding to an interface is always breaking. Try it out and observe the fatal error.

Issue #1102, opened 2015-06-01
[6.0.1] Handler stack documentation


bmancone:
In the latest documention of `Handlers and Middleware`, the way to retrieve the handler stack is :

`$stack = $client->getHandlerStack();`

However, the method `getHandlerStack()` does not exist.

eXtreme:
I think there should be a `$client->getConfig('handler')` call but.. will always `handler` be a "HandlerStack" instance? By default yes, but you can pass any callable s a client handler so you can't rely on that.

And of course in this doc these calls are on "client" object which is created later so I think it should be fixed just to `$stack = HandlerStack::create()`;

bmancone:
Yes, the way to retrieve the stack is to call `getConfig('handler')`. 

Replacing `$stack = $client->getHandlerStack();` by `$stack = new \GuzzleHttp\HandlerStack();` in the doc should improve comprehension, thanks.

PR #1103, opened 2015-06-01
fixed code examples on "Handlers and Middleware"

eXtreme:
fixes guzzle/guzzle#1102

### DIFF
--- a/docs/handlers-and-middleware.rst
+++ b/docs/handlers-and-middleware.rst
@@ -147,7 +147,7 @@ a function that takes the request argument and returns the request to send.
 
     use Psr7\Http\Message\RequestInterface;
 
-    $stack = $client->getHandlerStack();
+    $stack = new \GuzzleHttp\HandlerStack();
 
     $stack->push(Middleware::mapRequest(function (RequestInterface $request) {
         return $request->withHeader('X-Foo', 'bar');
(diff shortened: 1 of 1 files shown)


docs/handlers-and-middleware.rst
```diff
@@ -147,7 +147,7 @@ a function that takes the request argument and returns the request to send.
 
     use Psr7\Http\Message\RequestInterface;
 
-    $stack = $client->getHandlerStack();
+    $stack = new \GuzzleHttp\HandlerStack();
```
GrahamCampbell: Should import `GuzzleHttp\HandlerStack` first.

eXtreme: Well, I'm just consistent with some other code examples in this doc.. Just open https://github.com/guzzle/guzzle/blob/master/docs/handlers-and-middleware.rst and find `\GuzzleHttp\HandlerStack`. Even in the same codeblock there is `new \GuzzleHttp\Client` so.. what's the point. We should either make all code examples use imports or stick to FQCN.

Ff we take a closer look in the same codeblock there is `Middleware::mapRequest` which is neither imported nor FQCN.

eXtreme:
I changed all FQCN in this doc file to use statements.

mtdowling:
Thanks!

PR #1119, opened 2015-06-10
failing test for merging Uris using $client->request

ged15:
Hi!
Just ran into this today. Is this expected behaviour?
What I've found so far is that the `/` prefix gets prepended all the time in https://github.com/guzzle/psr7/blob/master/src/Uri.php#L574 then, since the relative `Uri` starts with a slash, it replaces the `basu_uri`.


mtdowling:
I believe it will work if you use a string as your base URI. There's a bug in the PSR7 package right now: https://github.com/guzzle/psr7/issues/18

mtdowling:
This should be fixed by https://github.com/guzzle/psr7/commit/644db73bfa1571af785172887836d6ec01d787ed

PR #1124, opened 2015-06-14
Date format to match Apache Common Log

rtheunissen:
Small update to match the referenced log format.

### DIFF
--- a/src/MessageFormatter.php
+++ b/src/MessageFormatter.php
@@ -108,7 +108,7 @@ class MessageFormatter
                         $result = $response ? $response->getBody() : 'NULL';
                         break;
                     case 'ts':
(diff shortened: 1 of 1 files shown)


mtdowling:
Heh. I just beat you to it :)

mtdowling:
See https://github.com/guzzle/guzzle/commit/861d2da555ebe00a036235700291793d6d554b79

rtheunissen:
Wasn't much of a race, but thanks for making the change. :) You could change the doc reference link to match Apache 2.4 (currently it points to 1.3) even though nothing seems to have changed. Nice to keep up to date regardless.

mtdowling:
Ah good point. I'll update the link. Thanks!

PR #1143, opened 2015-06-24
Support HHVM. See if CURLOPT_PROTOCOLS is defined before using it and other HHVM errors.

hussainweb:
This constant is not available on hhvm at this time. It causes test fails in other projects. I am aware that Guzzle allows failures for HHVM at this time but this check fixes it.

I am not sure if this could be a security issue. If the constant is not available, the feature is not available in curl and it wouldn't make any difference.

Lastly, FYI: HHVM has added this constant in an upcoming release as seen here: https://github.com/facebook/hhvm/blob/20287e468c5727578af55690b48099f1c09d55fd/NEWS#L87. It is not in any release yet AFAICT.


hussainweb:
The error fixed in https://github.com/guzzle/guzzle/commit/690f7af914953830bcc8da997b2e4089c97deac9 is best explained with this eval: http://3v4l.org/1LTC3. I am not sure if this counts as a bug in HHVM, or just a discrepancy. I fixed the bug in the commit though.

hussainweb:
@mtdowling Any suggestions or help here would be greatly appreciated. Thanks!

src/Handler/CurlFactory.php
```diff
@@ -173,9 +175,12 @@ private function getDefaultConf(EasyHandle $easy)
             CURLOPT_RETURNTRANSFER => false,
             CURLOPT_HEADER         => false,
             CURLOPT_CONNECTTIMEOUT => 150,
-            CURLOPT_PROTOCOLS      => CURLPROTO_HTTP | CURLPROTO_HTTPS,
         ];
 
+        if (defined('CURLOPT_PROTOCOLS')) {
+            $conf += [CURLOPT_PROTOCOLS => CURLPROTO_HTTP | CURLPROTO_HTTPS];
```
stof: This could be more readable:

```php
$conf[CURLOPT_PROTOCOLS] = CURLPROTO_HTTP | CURLPROTO_HTTPS;
```

hussainweb: That's embarrassing. I was so lost in array declarations. Fix coming up.

GrahamCampbell:
You can probably remove `composer self-update` from the travis file.

GrahamCampbell:
The file permissions on CurlFactory.php shouldn't have changed.

hussainweb:
@GrahamCampbell, I would be happy to remove composer self-update. I was just wondering if it is relevant here and does it make sense to be a separate PR?

I will fix the file permissions. That was a mistake.

src/Handler/CurlFactory.php
```diff
@@ -109,9 +109,11 @@ private static function finishError(
     ) {
         // Get error information and release the handle to the factory.
         $ctx = [
-            'errno'  => $easy->errno,
+            'errno' => $easy->errno,
             'error' => curl_error($easy->handle)
-                ?: curl_strerror($easy->errno)
+              ?: (function_exists('curl_strerror')
```
mtdowling: I think this can be updated to just use `curl_error` and not use this conditional.

src/Handler/StreamHandler.php
```diff
@@ -45,6 +45,7 @@ public function __invoke(RequestInterface $request, array $options)
             // This list can probably get more comprehensive.
             if (strpos($message, 'getaddrinfo') // DNS lookup failed
                 || strpos($message, 'Connection refused')
+                || strpos($message, "couldn't connect to host")
```
mtdowling: Can you add a comment that this is HHVM specific?

tests/Handler/MockHandlerTest.php
```diff
@@ -65,7 +65,7 @@ public function testCanGetLastRequestAndOptions()
     public function testCanEnqueueCallables()
     {
         $r = new Response();
-        $fn = function ($r, $o) use ($r) { return $r; };
+        $fn = function ($req, $o) use ($r) { return $r; };
```
mtdowling: Is this necessary for HHVM?

hussainweb: Yes, see http://3v4l.org/1LTC3

src/Handler/CurlFactory.php
hussainweb: You mean remove this line entirely? It will just be:

    'error' => curl_error($easy->handle),

tests/Handler/MockHandlerTest.php
hussainweb: Also, it fixes the corresponding failure.

src/Handler/CurlFactory.php
mtdowling: Yes exactly

stof:
IMO, you should allow failures on HHVM again until the Pool tests are fixed (unless you plan to fix them here before the merge, but I would like to see a version where the CURLOPT_PROTOCOLS issue is fixed)

mtdowling:
This PR seems more or less ready to be merged, but I think HHVM should be allowed to fail until all of the tests pass. That could be a separate set of PRs though. I'd like this PR to be in the next Guzzle release.

hussainweb:
Sorry for the delay on this. I had a busy weekend. I have pushed changes to allow failures on HHVM. I am looking forward to see this get in.

mtdowling:
Merged. Thanks! We can do more HHVM work in other PRs.

PR #1144, opened 2015-06-24
Fixing wrong documentation

csarrazi:
Fixed wrong documentation telling that both form_params and multipart could be used at the same time.


csarrazi:
Great reactivity, @mtdowling! :)

Issue #676, opened 2014-05-20
Cannot redeclare GuzzleHttp\Stream\create()


crynobone:
```
Fatal error: Cannot redeclare GuzzleHttp\Stream\create() (previously declared in /Users/crynobone/.composer/vendor/guzzlehttp/streams/src/functions.php:14) in /Users/crynobone/Projects/katsana/katsana/vendor/guzzlehttp/streams/src/functions.php on line 14
```

I'm facing an issue where two library depends on guzzle, one install using `composer global` and one as per project now throw above fatal error.

PR #1146, opened 2015-06-25
Conditionally require functions.php

itafroma:
Composer unconditionally "autoloads" files using `require` (see composer/composer#3003): this causes problems with globally-installed dependencies or where the composer autoload happens to be called more than once (see guzzle/guzzle#676).

One solution is to guard against function redeclaration using `function_exists()`: this was ruled out as not being done in the wild, but this PR uses this solution based on guzzle/promises and React doing the same thing (see [L4-7 in functions.php](https://github.com/guzzle/promises/blob/master/src/functions.php#L4-L7) and reactphp/promise#25, respectively).

The solution guzzle/promises currently uses does not work: PHP will attempt to load the functions before the conditional is evaluated. I've opted for the React solution, which is to point to a shim `functions_include.php` file that checks for the existence of the first function within `functions.php` before including `functions.php`.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,7 @@
         "psr/log": "~1.0"
     },
     "autoload": {
-        "files": ["src/functions.php"],
+        "files": ["src/functions_include.php"],
         "psr-4": {
             "GuzzleHttp\\": "src/"
         }
--- a/src/functions_include.php
+++ b/src/functions_include.php
@@ -1,0 +1,6 @@
+<?php
+
+// Don't redefine the functions if included multiple times.
+if (!function_exists('GuzzleHttp\uri_template')) {
+    require __DIR__ . '/functions.php';
+}


itafroma:
Mostly-identical PRs for guzzle/promises and guzzle/psr7 were filed:

* guzzle/promises#6
* guzzle/psr7#22

curry684:
Please note that this still completely breaks production environments using 6.0.2 as it still depends on the outdated versions of PSR7 and Promises with the same issue. I have patched it for now by reversing the conditions and adding it in PSR7 lib but that's no stable solution.

The problem as a whole is that PHP parses the top scope before evaluating anything. So 

    if(function_exists('aap'))
      return;
    function aap() { }

crashes, while

    if(!function_exists('aap')) {
      function aap() { }
    }

Does not.

PR #1147, opened 2015-06-25
Updating StreamHandler to add a size to "stream" response bodies

mtdowling:
@jeremeamia @jeskew 

This resolves https://github.com/aws/aws-sdk-php/issues/590


mtdowling:
Actually. I take this back. This feels a little weird to add here.

PR #1163, opened 2015-07-08
Force form_params to use & as fields separator

kkopachev:
According to w3.org separator [should always be](http://www.w3.org/TR/html401/interact/forms.html#h-17.13.4) `&`


mtdowling:
Thanks!

Issue #1166, opened 2015-07-09
Get the effective URL


dawehner:
We (the drupal project, see https://www.drupal.org/node/2493911) try to update to guzzle 6,
especially in order to profit from the long term support.

The old code had \GuzzleHttp\Message\ResponseInterface::getEffectiveUrl which was pretty easy to use.
For guzzle 6 / PSR-7 this method seems to no longer exist.
As I saw there is the redirect middleware, but still, there is nothing to get out of it.

Note: Drupal currently stores the client on the container, in order to always use the same configuration.
This makes it harder to use some of the middlewares, like history, which might be useful to implement something as getEffectiveUrl.

My only thought was to implement an interface which extends ResponseInterface and somehow we wrap in a middleware the response to contain that URL.

Feedback or some general ideas would be great!

Crell:
This does seem to be a regression, since callers need to know if they should update their own records for a 301 (which is nominally what you're supposed to do).  I see 2 options:

1) Have them redirect middleware stick an X-EffectiveUrl header on it, or something, that people can check for.

2) As dawehner suggests, add:

```php
interface RedirectedResponseInterface extends ResponseInterface {
  public function getEffectiveUrl();
  public function getRedirectType();
}
```

And then the middleware, if there was a redirect, returns that instead of just a ResponseInterface.  Callers can then do:

```php
if ($response instanceof RedirectedResponseInterface && $response->getRedirectType() == 301) {
    // Update my database with $response->getEffectiveUrl();
}
```

I somewhat prefer option 2.

dawehner:
Redirect type is kinda pointless, isn't it? You already have the status code but yeah +1

mtdowling:
I don't think extending the response would be a great idea since the contract is that Guzzle returns a PSR-7 response; any middleware that modifies the return value to use a specific implementation of a PSR-7 response could be undone by another middleware that returns a different response object. i.e., there's no mandate that responses are chained using `with` methods, which means decorating would not solve the problem in a generic way.

I think the idea of adding a redirect header when a redirect is encountered is cool.

[...]

PR #1183, opened 2015-07-17
Add a redirect history

dawehner:
See https://github.com/guzzle/guzzle/issues/1166#issuecomment-122248990

### DIFF
--- a/src/RedirectMiddleware.php
+++ b/src/RedirectMiddleware.php
@@ -21,7 +21,8 @@ class RedirectMiddleware
         'max'       => 5,
         'protocols' => ['http', 'https'],
         'strict'    => false,
-        'referer'   => false
+        'referer'   => false,
+        'track_redirects' => false,
     ];
 
     /** @var callable  */
@@ -98,7 +99,18 @@ class RedirectMiddleware
             );
         }
 
-        return $this($nextRequest, $options);
+        /** @var PromiseInterface|ResponseInterface $response */
+        $options['_previous_uris'][] = $request->getUri();
+        $response = $this($nextRequest, $options);
+
+        // Add headers to be able to track history of redirects.
+        if (!empty($options['allow_redirects']['track_redirects'])) {
+            $response = $response->wait(TRUE);
+            if (!$response->hasHeader('X-Guzzle-Redirect')) {
+                $response = $response->withHeader('X-Guzzle-Redirect', $options['_previous_uris']);
+            }
+        }
+        return $response;
     }
 
     private function guardMax(RequestInterface $request, array &$options)
--- a/tests/RedirectMiddlewareTest.php
+++ b/tests/RedirectMiddlewareTest.php
@@ -6,6 +6,7 @@ use GuzzleHttp\HandlerStack;
 use GuzzleHttp\Middleware;
 use GuzzleHttp\Psr7\Request;
 use GuzzleHttp\Psr7\Response;
+use Psr\Http\Message\ResponseInterface;
 
 /**
  * @covers GuzzleHttp\RedirectMiddleware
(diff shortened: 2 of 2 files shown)


dawehner:
Mh, this seems to be unrelated to that patch

tests/RedirectMiddlewareTest.php
```diff
@@ -128,6 +129,41 @@ public function testAddsRefererHeader()
...
+            'allow_redirects' => [
+                  'track_redirects' => true,
+            ],
+        ]);
+        /** @var ResponseInterface $response */
+        $response = $promise->wait(TRUE);
+
+        $this->assertEquals(
+          'http://example.com?a=b',
+          $response->getHeaderLine('X-Guzzle-Redirect-0')
```
Crell: Do we want the headers distinctly named or should they share a name?  Distinctly named is easier if you know what you're looking for.  Shared name means they come back as a collection, which is better if you don't know exactly how many you're working with.

I think I'd marginally prefer a shared header, but defer to @mtdowling

mtdowling: I agree. I'd prefer a shared header.

mtdowling:
What do you think, @dawehner? Can we use the same header and just append a value to the header? Using the same header field to track ordered values should be acceptable based on RFC 7230:

```
  A recipient MAY combine multiple header fields with the same field
   name into one "field-name: field-value" pair, without changing the
   semantics of the message, by appending each subsequent field value to
   the combined field value in order, separated by a comma.  The order
   in which header fields with the same field name are received is
   therefore significant to the interpretation of the combined field
   value; a proxy MUST NOT change the order of these field values when
   forwarding a message.
```

See http://tools.ietf.org/html/rfc7230#section-3.2.2

dawehner:
Does that look more like you expect it to look like?

src/RedirectMiddleware.php
```diff
@@ -98,7 +99,18 @@ public function checkRedirect(
...
         }
 
-        return $this($nextRequest, $options);
+        /** @var PromiseInterface|ResponseInterface $response */
+        $options['_previous_uris'][] = $request->getUri();
+        $response = $this($nextRequest, $options);
+
+        // Add headers to be able to track history of redirects.
+        if (!empty($options['allow_redirects']['track_redirects'])) {
+            $response = $response->wait(TRUE);
```
mtdowling: We don't want to wait here. Instead, you should return a new promise that modifies the response before it is returned.

src/RedirectMiddleware.php
```diff
@@ -98,7 +99,18 @@ public function checkRedirect(
...
+            if (!$response->hasHeader('X-Guzzle-Redirect')) {
```
mtdowling: Instead of this, just use the `withAddedHeader` method of a response and remove the if statement.

mtdowling:
Any news here?

mtdowling:
Any update?

dawehner:
Not yet sorry. I need to fine some time, but in case anyone wants to, feel
free to take it over.

src/RedirectMiddleware.php
```diff
@@ -21,7 +21,8 @@ class RedirectMiddleware
         'max'       => 5,
         'protocols' => ['http', 'https'],
         'strict'    => false,
-        'referer'   => false
+        'referer'   => false,
+        'track_redirects' => false,
```
GrahamCampbell: odd how the rest of the array is aligned, but this isn't

tests/RedirectMiddlewareTest.php
```diff
@@ -128,6 +129,30 @@ public function testAddsRefererHeader()
         );
     }
 
+    public function testAddsGuzzleRedirectHeader()
+    {
+        $mock = new MockHandler([
+          new Response(302, ['Location' => 'http://example.com']),
```
GrahamCampbell: you need 4 spaces for indentation, not 2

src/RedirectMiddleware.php
dawehner: I tried really hard to use a promise, but I simply cannot figure it out properly.
The problem is that the most outer promise will need to have the information of all redirects. On the other hand the most outer promise is created within the scope of none already made redirect.

The most inner promise though has the information of all redirects available, but I have absolutely no clue at that point, how to pass down that information. For quite some time I considered the $response from checkRedirect as the previous response, but this is actually not the case. Its, as you know, a response created for manipulation only, something which is designed in a way that the API is independent of the concrete PSR7 implementation.

src/RedirectMiddleware.php
dawehner: Well, this doesn't work as you expect it. With this if() we ensure that just the most outer response, so the one all the previous URIs available, sets the header, I think, but swapping just that lets the test fail

stefanruijsenaars:
Added a PR that includes a fix for the conflict and the coding standard fixes at https://github.com/guzzle/guzzle/issues/1244

mtdowling:
I cleaned this up and made a few tweaks.

dawehner:
You are truly a great opensource maintainer! The project queue is cleaned up regularly, you respond quickly, you are friendly etc. Thank you for being awesome

mtdowling:
Thanks, @dawehner :)

PR #1187, opened 2015-07-21
Changed the reason phrase casting from int to string 

pinkal-vansia:
int casting of reason phrase was converting legitimate string response header message to 0. e.g. If api send res `header("HTTP/1.1 401 Some message");` , "Some message" was resulted in 0 because of int casting and hence could not be fetched with `getReasonPhrase()` method. Instead default matching message from `$phrases` message map from `Response` class was returned. I looked in docs for reason behind int casting but could not find it.


mtdowling:
Thanks!

pinkal-vansia:
Pleasure!

PR #1189, opened 2015-07-23
Add content-length header to outgoing requests if not supplied

jeskew:
Curl will assign a content-length header of '0' when it's missing from a request, but the StreamHandler will not. This appeared to be the cause of [an issue on the AWS SDK](https://github.com/aws/aws-sdk-php/issues/701).


jeskew:
Curl will assign a content-length header of '0' when it's missing from a request, but the StreamHandler will not. This appeared to be the cause of [an issue on the AWS SDK](https://github.com/aws/aws-sdk-php/issues/701).

PR #1197, opened 2015-07-28
Allow guzzle to read environment NO_PROXY variable

fbiesse:
Avoid to use proxy if NO_PROXY is configured

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -172,6 +172,11 @@ class Client implements ClientInterface
             $defaults['proxy']['https'] = $proxy;
         }
 
+        if ($noProxy = getenv('NO_PROXY')) {
+        	$cleanedNoProxy = preg_replace('/\s/', '', $noProxy);
+        	$defaults['proxy']['no'] = explode(',', $cleanedNoProxy);
+        }
+        
         $this->config = $config + $defaults;
 
         if (!empty($config['cookies']) && $config['cookies'] === true) {
--- a/src/Client.php
+++ b/src/Client.php
@@ -173,7 +173,7 @@ class Client implements ClientInterface
         }
 
         if ($noProxy = getenv('NO_PROXY')) {
-            $cleanedNoProxy = preg_replace('/\s/', '', $noProxy);
+            $cleanedNoProxy = str_replace(' ', '', $noProxy);
             $defaults['proxy']['no'] = explode(',', $cleanedNoProxy);
         }
         
--- a/src/Client.php
+++ b/src/Client.php
@@ -173,8 +173,8 @@ class Client implements ClientInterface
         }
 
         if ($noProxy = getenv('NO_PROXY')) {
-        	$cleanedNoProxy = preg_replace('/\s/', '', $noProxy);
-        	$defaults['proxy']['no'] = explode(',', $cleanedNoProxy);
+            $cleanedNoProxy = preg_replace('/\s/', '', $noProxy);
+            $defaults['proxy']['no'] = explode(',', $cleanedNoProxy);
         }
         
         $this->config = $config + $defaults;
--- a/src/Handler/CurlFactory.php
+++ b/src/Handler/CurlFactory.php
@@ -362,9 +362,12 @@ class CurlFactory implements CurlFactoryInterface
             if (!is_array($options['proxy'])) {
                 $conf[CURLOPT_PROXY] = $options['proxy'];
             } elseif ($scheme = $easy->request->getUri()->getScheme()) {
-                if (isset($options['proxy'][$scheme])) {
-                    $conf[CURLOPT_PROXY] = $options['proxy'][$scheme];
-                }
+            	$host = $easy->request->getUri()->getHost();
+            	if(!isset($options['proxy']['no']) || !in_array($host, $options['proxy']['no'])){
+                    if (isset($options['proxy'][$scheme])) {
+                        $conf[CURLOPT_PROXY] = $options['proxy'][$scheme];
+                    }        		
+            	}
             }
         }
 
--- a/src/Handler/CurlFactory.php
+++ b/src/Handler/CurlFactory.php
@@ -363,7 +363,7 @@ class CurlFactory implements CurlFactoryInterface
                 $conf[CURLOPT_PROXY] = $options['proxy'];
             } elseif ($scheme = $easy->request->getUri()->getScheme()) {
             	$host = $easy->request->getUri()->getHost();
-            	if(!isset($options['proxy']['no']) || !in_array($host, $options['proxy']['no'])){
+            	if(!isset($options['proxy']['no']) || !$this->isHostInProxyArea($host, $options['proxy']['no'])) {
                     if (isset($options['proxy'][$scheme])) {
                         $conf[CURLOPT_PROXY] = $options['proxy'][$scheme];
                     }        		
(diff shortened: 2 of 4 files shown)


src/Client.php
```diff
@@ -172,6 +172,11 @@ private function configureDefaults(array $config)
             $defaults['proxy']['https'] = $proxy;
         }
 
+        if ($noProxy = getenv('NO_PROXY')) {
+        	$cleanedNoProxy = preg_replace('/\s/', '', $noProxy);
```
mtdowling: can you use 4 spaces?

src/Handler/CurlFactory.php
```diff
@@ -362,9 +362,12 @@ private function applyHandlerOptions(EasyHandle $easy, array &$conf)
             if (!is_array($options['proxy'])) {
                 $conf[CURLOPT_PROXY] = $options['proxy'];
             } elseif ($scheme = $easy->request->getUri()->getScheme()) {
-                if (isset($options['proxy'][$scheme])) {
-                    $conf[CURLOPT_PROXY] = $options['proxy'][$scheme];
-                }
+            	$host = $easy->request->getUri()->getHost();
+            	if(!isset($options['proxy']['no']) || !in_array($host, $options['proxy']['no'])){
```
mtdowling: Can you put a space between the braces and the surrounding characters?

mtdowling:
We should match what other clients are doing with NO_PROXY. For example: https://github.com/kennethreitz/requests/pull/945/files

[cURL says this](http://curl.haxx.se/docs/manpage.html#NOPROXY):

> list of host names that shouldn't go through any proxy. If set to a asterisk '*' only, it matches all hosts.

So adding support for `*` might be nice.

[wget says](http://www.gnu.org/software/wget/manual/html_node/Proxies.html):

> no_proxy: This variable should contain a comma-separated list of domain extensions proxy should not be used for. For instance, if the value of no_proxy is ‘.mit.edu’, proxy will not be used to retrieve documents from MIT.

This makes me think we should make it match if the host ends with any value in the CSV list of NO_PROXY servers to match wget and Requests.

fbiesse:
I add support to subdomains and wilcards

src/Client.php
```diff
@@ -172,6 +172,11 @@ private function configureDefaults(array $config)
             $defaults['proxy']['https'] = $proxy;
         }
 
+        if ($noProxy = getenv('NO_PROXY')) {
+            $cleanedNoProxy = preg_replace('/\s/', '', $noProxy);
```
mtdowling: I think this should just be `str_replace(' ', '', $noProxy);`

src/Handler/CurlFactory.php
```diff
@@ -362,9 +362,12 @@ private function applyHandlerOptions(EasyHandle $easy, array &$conf)
             if (!is_array($options['proxy'])) {
                 $conf[CURLOPT_PROXY] = $options['proxy'];
             } elseif ($scheme = $easy->request->getUri()->getScheme()) {
-                if (isset($options['proxy'][$scheme])) {
-                    $conf[CURLOPT_PROXY] = $options['proxy'][$scheme];
-                }
+            	$host = $easy->request->getUri()->getHost();
+            	if(!isset($options['proxy']['no']) || !$this->isHostInProxyArea($host, $options['proxy']['no'])) {
```
mtdowling: Can yo limit lines to 80 characters and follow PSR-2? http://www.php-fig.org/psr/psr-2/

src/Handler/CurlFactory.php
```diff
@@ -419,6 +422,28 @@ private function applyHandlerOptions(EasyHandle $easy, array &$conf)
...
+        $matches = false;
+        foreach ($areas as $area) {
+            $areaToMatch = $area;
+            // if include all subdomains 
+            if(preg_match('#^\.#', $area)) {
+                $areaToMatch = "*$area";
+            }
+            // Use wilcards
+            $delimiter = '#';
+            $starEscaped = preg_quote('*', $delimiter);
```
mtdowling: Why do this here? You know that this will be `\*` every time.

src/Handler/CurlFactory.php
```diff
@@ -419,6 +422,28 @@ private function applyHandlerOptions(EasyHandle $easy, array &$conf)
...
         }
     }
+    
+    private function isHostInProxyArea($host, array $areas)
+    {
+        $matches = false;
+        foreach ($areas as $area) {
+            $areaToMatch = $area;
+            // if include all subdomains 
+            if(preg_match('#^\.#', $area)) {
```
mtdowling: Couldn't this just be `if (substr($area, 0, 1) == '.')`?

fbiesse: I'll change but my solution allow to manage the impossible case with empty area

kamilsk: > I'll change but my solution allow to manage the impossible case with empty area

In what cases?

```
php > error_reporting(E_ALL);
php > var_dump((substr(null, 0, 1) === '.') === (preg_match('/^\./', null) === 1));
bool(true)
php > var_dump((substr('', 0, 1) === '.') === (preg_match('/^\./', '') === 1));
bool(true)
```

src/Handler/CurlFactory.php
```diff
@@ -419,6 +422,28 @@ private function applyHandlerOptions(EasyHandle $easy, array &$conf)
...
+                $areaToMatch = "*$area";
+            }
+            // Use wilcards
+            $delimiter = '#';
```
kamilsk: Any advantages of using this delimeter instead of the standard?

src/Client.php
fbiesse: Me too, but it was just to avoid tab

kamilsk: It may be better to use trim function?
```
php > var_dump(trim(" \t "));
string(0) ""
```
If no why only spaces will be removed?

src/Handler/CurlFactory.php
```diff
@@ -419,6 +423,27 @@ private function applyHandlerOptions(EasyHandle $easy, array &$conf)
...
+    {
+        $matches = false;
+        foreach ($areas as $area) {
+            $areaToMatch = $area;
+            // if include all subdomains 
+            if (!empty($area) && substr($area, 0, 1) === '.'){
+                $areaToMatch = "*$area";
+            }
+            // Use wilcards
+            $delimiter = '#';
```
kamilsk: what is the need?

src/Client.php
fbiesse: Yes, sure, but I don't know if it is the best solution. On these lines, I remove all spaces / tabs of a list of coma separated string. So my first solution allow to remove all spaces on the string. Your just remove spaces on the begining and at the end, not all.

src/Handler/CurlFactory.php
fbiesse: The need is to escape regex characters that we can have in the hostname and take in consideration the wilcard.

src/Client.php
kamilsk: Sorry did not see
```
$defaults['proxy']['no'] = explode(',', $cleanedNoProxy);
```
Then preg_replace isn't so bad.

src/Handler/CurlFactory.php
kamilsk: but you already using preg_quote for $areaToMatch

fbiesse: Yes, sorry, I really do not understand what is the problem.

kamilsk: No questions, thank you!

src/Client.php
mtdowling: > Me too, but it was just to avoid tab

I get removing spaces, but I don't think it's necessary to remove tabs here.

src/Handler/CurlFactory.php
```diff
@@ -362,9 +362,13 @@ private function applyHandlerOptions(EasyHandle $easy, array &$conf)
             if (!is_array($options['proxy'])) {
                 $conf[CURLOPT_PROXY] = $options['proxy'];
             } elseif ($scheme = $easy->request->getUri()->getScheme()) {
-                if (isset($options['proxy'][$scheme])) {
-                    $conf[CURLOPT_PROXY] = $options['proxy'][$scheme];
-                }
+            	$host = $easy->request->getUri()->getHost();
+            	if(!isset($options['proxy']['no'])
```
mtdowling: Please ensure spaces between all `if` and opening parens.

mtdowling:
I've cleaned this up a bit and merged it. I removed all of the regex stuff and made sure that proxy rules are applied for both curl and the PHP stream wrapper. In order to achieve this, I moved the no_proxy matching code into `functions.php`.

andig:
Love this bit. Any chance of backporting to 5.3 for those stuck with php 5.4?

PR #1202, opened 2015-07-31
on_stats request option

### LINKED ISSUE #1152: Cannot get connection time in middleware
In previous versions it was easy to get the (curl) connection time via `$event->getTransferInfo()`.
In guzzle6 I can't find a good way to do this. It seems that this information is not even stored anywhere.

I am aware that including this additional information would break the elegance and purity of the code so I can live without it. But maybe I am missing sth and it would be possible to pass this information so it's accessible in a handler?

I am using guzzle with a [Symfony2] (https://github.com/pinkeen/ApiDebugBundle) bundle which simplifies API debugging  in a few projects. Having the connection time really helps with identifying network problems and slow API endpoints/calls.

mtdowling:
This is  WIP commit that adds the `on_stats` request option.

Closes #1152

### DIFF
--- a/tests/PrepareBodyMiddlewareTest.php
+++ b/tests/PrepareBodyMiddlewareTest.php
@@ -25,7 +25,7 @@ class PrepareBodyMiddlewareTest extends \PHPUnit_Framework_TestCase
         $stack->push($m);
         $comp = $stack->resolve();
         $p = $comp(new Request('PUT', 'http://www.google.com', [], '123'), []);
-        $this->assertInstanceOf('GuzzleHttp\Promise\FulfilledPromise', $p);
+        $this->assertInstanceOf('GuzzleHttp\Promise\PromiseInterface', $p);
         $response = $p->wait();
         $this->assertEquals(200, $response->getStatusCode());
     }
@@ -47,7 +47,7 @@ class PrepareBodyMiddlewareTest extends \PHPUnit_Framework_TestCase
         $stack->push($m);
         $comp = $stack->resolve();
         $p = $comp(new Request('PUT', 'http://www.google.com', [], $body), []);
-        $this->assertInstanceOf('GuzzleHttp\Promise\FulfilledPromise', $p);
+        $this->assertInstanceOf('GuzzleHttp\Promise\PromiseInterface', $p);
         $response = $p->wait();
         $this->assertEquals(200, $response->getStatusCode());
     }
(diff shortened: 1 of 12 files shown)


GrahamCampbell:
Looks pretty nice. :)

mtdowling:
Do you think the TransferStats class actually needs accessor methods, or should it just use public properties?

GrahamCampbell:
Private properties with accessors is the best design wise, but public properties might save time. I use public properties more and more as I find it makes things easy to test and fast to write.

stefanruijsenaars:
Just out of interest, what still needs to happen here?

mtdowling:
@stefanruijsenaars mostly just tests and docs

mtdowling:
This feature is now complete. How does it look from a functionality point of view,@GrahamCampbell / @Crell?

tests/TransferStatsTest.php
```diff
@@ -0,0 +1,29 @@
+<?php
+namespace GuzzleHttp\Tests;
+
+use GuzzleHttp\TransferStats;
+use GuzzleHttp\Psr7;
+
+class TransferStatsTest extends \PHPUnit_Framework_TestCase
```
dawehner: Just curious: you don't use @coversDefaultClass and @covers right?

tests/TransferStatsTest.php
```diff
@@ -0,0 +1,29 @@
...
+{
+    public function testHasData()
```
dawehner: This test method name sounds wrong, what about testTransferStats instead?

tests/PrepareBodyMiddlewareTest.php
```diff
@@ -126,7 +126,7 @@ function (RequestInterface $request) {
             new Request('PUT', 'http://www.google.com', ['Expect' => 'Foo'], $bd),
             ['expect' => true]
         );
-        $this->assertInstanceOf('GuzzleHttp\Promise\FulfilledPromise', $p);
+        $this->assertInstanceOf('GuzzleHttp\Promise\PromiseInterface', $p);
```
dawehner: I'm curious, did you considered to use PromiseInterface::class instead?

dawehner:
The general code looks pretty sane for me

tests/TransferStatsTest.php
liveoutloud2day: In the other tests, classes are named SomethingTest, the functions are called testSomething, so I think this is OK. (Not really sure, but that is what the file tests/ClientTest.php has in its code).

tests/PrepareBodyMiddlewareTest.php
mtdowling: I forgot about that feature, but it would be nice to update the tests to use it at some point.

lussoluca:
It's possibile to extract that information from a middleware?

dawehner:
Can't your middleware manipulate the $options passed into the further down middleware/handler and by that register a callable?

lussoluca:
Oh yes, I can do that! Thanks

tests/PrepareBodyMiddlewareTest.php
dawehner: Here is one: https://github.com/guzzle/guzzle/pull/1253

PR #1211, opened 2015-08-10
Cleanup import

GrahamCampbell:
Not sure why you had it aliased like that?


mtdowling:
I think there used to be way more mentions of this exception. This seems fine I guess.

Issue #1208, opened 2015-08-06
Ability to use PHP_QUERY_RFC1738 for query params


kfuchs:
Currently the way the `query` params are url encoded is too opinionated (referring to [Client LN:351](https://github.com/guzzle/guzzle/blob/master/src/Client.php#L351)). 

There is no straight forward way to pass in query params utilizing `PHP_QUERY_RFC1738`.

One must `html_build_query()` them selves before hand, and then pass it in as a string instead of an array. This isn't extremely complicated, but it makes for guzzle a little clunkier to use, and when someone writes that line of code it will leave a future developer wondering why it is they build the query string instead of just passing it into guzzle as an array.

Perhaps instead there should be an option, like such:
`http_build_query($value, null, '&', $option['enc_type']);`

kfuchs:
Made a pull request #1212 which fixes this.

It will still default to `PHP_QUERY_RFC3986` as the encode type. But now it can be changed via:

```php
$guzzle->get('...', [
			'enc_type' => PHP_QUERY_RFC1738
		]);
```

mtdowling:
See https://github.com/guzzle/guzzle/issues/1196

Issue #1196, opened 2015-07-28
duplicateAggregator() functionality missing with update to Psr7


LKS90:
Hey there! I just had a problem when porting some code to the newer Guzzle version. With the removal of `GuzzleHttp\Query`, the `Query::duplicateAggregator()` functionality is now gone and it seems impossible to easily replace some queries now.

My query is an array, important is that 'q' is a non-associative array.

```
$request_query = [
  'key' => 'INSERT-YOUR-KEY',
  'source' => 'en',
  'target' => 'de',
  'q' => [
    'Hello World',
    'My name is Jeff',
  ],
];
```

 The code I am porting:

```
$query = $request->getQuery();
$query->merge($request_query);
$query->set('key', $translator->getSetting('api_key'));
$query->setAggregator($query::duplicateAggregator());
```

The result should be

```
?key=INSERT-YOUR-KEY&source=en&target=de&q=Hello%20world&q=My%20name%20is%20Jeff
```

My problem when porting the code: There seems to be no possibility to repeat a key for an non-associative array, the way it used to work with `setAggregator($query::duplicateAggregator())`.

I resorted to building the query string myself and attaching it when sending, but if there is some easy solution I overlooked, I'd be happy if you would hint me towards it.

If this is a regression, is it possible to reintroduce?

mtdowling:
Removing the Query class was intentional and not a regression.

Guzzle now accepts an array or a string as the `query` request option. The big shift from Guzzle 5 to 6 is that PSR-7 represents query strings on request objects simply as strings. Including a Query class in Guzzle 6 would really just be a query string builder that creates a string of data (a request would not own the query object).

Berdir:
Yes, such a builder is exactly what would be helpful if you have query strings that can't be represented/built by PHP from an array, for example the duplicate aggregator implementation.

See https://www.drupal.org/node/2539228#comment-10160842 for a before/after

mtdowling:
Sure. We can add query string serialization into Guzzle 6. I don't know if we need a full on builder or just a couple convenience functions that can be used to convert an array into a query string.

for example:

- duplicate aggregation
- PHP style aggregation
- PHP style aggregation that does not use an incrementing index.

Berdir:
Sure, I don't really care about if it's a builder class or a function.. whatever works for you.

peec:
I'm in need for this. Guzzle should support this since it's library is very often used for 3rdparty API's.



```
'query' => [
[...]

PR #1212, opened 2015-08-10
Added option to encode url via RFC 1738

kfuchs:
You can now add the option to encode using PHP_RFC_1738 by passing ['enc_type' => PHP_RFC_1738] as an option to the guzzle client.

fixes issue: #1208

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -160,7 +160,8 @@ class Client implements ClientInterface
             'http_errors'     => true,
             'decode_content'  => true,
             'verify'          => true,
-            'cookies'         => false
+            'cookies'         => false,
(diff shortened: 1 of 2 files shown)


mtdowling:
I like the simplicity of the current approach. If you want to encode using RFC 1738, then just pass in a pre-encoded query string to the `query` request option. I think a better approach to this would be to introduce query builders into Guzzle that can then be passed into the `query` option. They would have custom encoding and aggregation built-in and know how to cast to a string. Or perhaps it's just a few functions that take an array and create a string.. Not sure yet. Either way, I think it's better to push this responsibility somewhere else instead of adding a new option.

See: https://github.com/guzzle/guzzle/issues/1196

PR #1220, opened 2015-08-17
If the name string is empty, ignore the set-cookie-string entirely

### LINKED ISSUE #1205: Incorrect interpretation of RFC 6265 in SetCookie::validate
[SetCookie::validate](https://github.com/guzzle/guzzle/blob/6692db3cda013e8357330243c33fe1838d0ea6f5/src/Cookie/SetCookie.php#L366) checks if the cookie is valid according to RFC 6265

It includes the following checks:
  1. Names must not be empty, but can be 0
  1. Value must not be empty, but can be 0

[RFC 6265](http://tools.ietf.org/html/rfc6265#page-18)
```
   3.  The (possibly empty) name string consists of the characters up
       to, but not including, the first %x3D ("=") character, and the
       (possibly empty) value string consists of the characters after
       the first %x3D ("=") character.

   4.  Remove any leading or trailing WSP characters from the name
       string and the value string.

   5.  If the name string is empty, ignore the set-cookie-string
       entirely.

```

It only says to ignore the cookie, not to do something drastic like throwing exception, when name string is empty and says nothing about empty value.

halabuda:
really nice library guys!
fixes #1205

### DIFF
--- a/src/Cookie/CookieJar.php
+++ b/src/Cookie/CookieJar.php
@@ -148,6 +148,11 @@ class CookieJar implements CookieJarInterface
 
     public function setCookie(SetCookie $cookie)
     {
+        //If the name string is empty, ignore the set-cookie-string entirely
+        if(empty($cookie->getName())){
+            return false;        
+        }
+
         // Only allow cookies with set and valid domain, name, value
         $result = $cookie->validate();
         if ($result !== true) {


mtdowling:
Sweet! Can you add a test case for this?

src/Cookie/CookieJar.php
```diff
@@ -148,6 +148,11 @@ function (SetCookie $cookie) {
 
     public function setCookie(SetCookie $cookie)
     {
+        //If the name string is empty, ignore the set-cookie-string entirely
+        if(empty($cookie->getName())){
```
GrahamCampbell: cs

src/Cookie/CookieJar.php
```diff
@@ -148,6 +148,11 @@ function (SetCookie $cookie) {
 
     public function setCookie(SetCookie $cookie)
     {
+        //If the name string is empty, ignore the set-cookie-string entirely
```
GrahamCampbell: cs

halabuda:
@mtdowling sure thing. working on it now.

georaldc:
I think this should follow the "Names must not be empty, but can be 0" rule too, or move the empty name check condition in the validate function outside

stefanruijsenaars:
Added a PR in https://github.com/guzzle/guzzle/pull/1245

mtdowling:
Addressed by #1245

PR #1221, opened 2015-08-18
Fix changing curl options

BostjanOb:
Currently you were able to add new curl options. But you were unable to change existing one (as for HTTP VERSION).

This pull request fix this.


mtdowling:
Thanks!

PR #1230, opened 2015-08-24
Fix phpdoc for configureDefaults() that does not have a return value

Tobion:


src/Client.php
```diff
@@ -149,8 +149,6 @@ private function buildUri($uri, array $config)
      * Configures the default options for a client.
      *
      * @param array $config
-     *
-     * @return array
```
GrahamCampbell: shouldn't be deleted - should be replaced with `@return void`

GrahamCampbell: hmmm, maybe not, guzzle doesn't seem do always do that

mtdowling:
Thanks!

@GrahamCampbell I am not a big fan of `@return void` :)

PR #1243, opened 2015-09-06
Fix options merge, ensure latter options override

dshafik:
As shown in [this 3v4l example](https://3v4l.org/Z3Deo), using `+=` does not have the desired result of overriding the matching keys in the left operand with those as the right operand. `array_merge()` will just append duplicate numeric keys as a new entry in the array; `array_replace()` as used here gives the desired behavior.

### DIFF
--- a/src/Handler/CurlFactory.php
+++ b/src/Handler/CurlFactory.php
@@ -46,7 +46,7 @@ class CurlFactory implements CurlFactoryInterface
 
         // Add handler options from the request configuration options
         if (isset($options['curl'])) {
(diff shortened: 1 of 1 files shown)


mtdowling:
This commit previously addressed this issue: https://github.com/guzzle/guzzle/commit/12fe5b4a9eddb0a53efc2371aaf20cf0bdd0f427. The current behavior correctly favors an explicitly provided hash of options that are then combined with the options created for the request. Any option that was passed in by the user overrides any option that may have been built up when creating the cURL handle.

PR #1244, opened 2015-09-06
Add a redirect history

stefanruijsenaars:
Includes coding standard fixes in https://github.com/guzzle/guzzle/pull/1183

### DIFF
--- a/src/RedirectMiddleware.php
+++ b/src/RedirectMiddleware.php
@@ -18,10 +18,11 @@ use Psr\Http\Message\UriInterface;
 class RedirectMiddleware
 {
     public static $defaultSettings = [
-        'max'       => 5,
-        'protocols' => ['http', 'https'],
-        'strict'    => false,
-        'referer'   => false
+        'max'             => 5,
+        'protocols'       => ['http', 'https'],
+        'strict'          => false,
+        'referer'         => false,
+        'track_redirects' => false,
     ];
 
     /** @var callable  */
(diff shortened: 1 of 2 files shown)


tests/RedirectMiddlewareTest.php
```diff
@@ -130,6 +130,30 @@ public function testAddsRefererHeader()
...
+        $request = new Request('GET', 'http://example.com?a=b');
+        $promise = $handler($request, [
+            'allow_redirects' => [
+                  'track_redirects' => true,
+            ],
+        ]);
+        /** @var ResponseInterface $response */
+        $response = $promise->wait(TRUE);
+
+        $this->assertEquals(['http://example.com?a=b', 'http://example.com', 'http://example.com/foo'], $response->getHeader('X-Guzzle-Redirect'));
```
GrahamCampbell: this line is quite long

mtdowling:
You should really branch off the existing work that this originated from so that the original author's commits are still in the git history.

src/RedirectMiddleware.php
```diff
@@ -98,7 +99,18 @@ public function checkRedirect(
...
         }
 
-        return $this($nextRequest, $options);
+        /** @var PromiseInterface|ResponseInterface $response */
+        $options['_previous_uris'][] = $request->getUri();
+        $response = $this($nextRequest, $options);
+
+        // Add headers to be able to track history of redirects.
+        if (!empty($options['allow_redirects']['track_redirects'])) {
+            $response = $response->wait(TRUE);
```
mtdowling: We should not block here. It should be a promise.

mtdowling:
#1183 was merged with modifications so this is no longer necessary.

PR #1246, opened 2015-09-06
Fix options merge, ensure latter options override

dshafik:


mtdowling:
Because some curl options must be set in a specific order in relation to other options, I think `array_replace()` makes more sense than doing a union with `+`.

PR #1250, opened 2015-09-07
on_headers if isset

ferchichianis:
when $easy->options['on_headers'] is set it will not be checked by is_callable


mtdowling:
This doesn't look right. We want `$onHeaders` to be allowed to be null or a callable. I suppose your issue is that you can set `on_headers` to `null` directly?

PR #1253, opened 2015-09-09
Use ::class in tests where it was possible

dawehner:
As guzzle depends on PHP 5.5  we can use ::class in some tests and make the code a bit easier
to browser in IDEs etc.


mtdowling:
Thanks!

PR #1261, opened 2015-09-19
Add back the query classes from 5.3

jeskew:
Guzzle 6 (and most PSR-7 clients) only accept a string as a query, which is not the best interface if you'd like to add and remove values across multiple functions. This PR brings `GuzzleHttp\Query` from the 5.3 branch back into Guzzle 6. It also brings pack `GuzzleHttp\QueryParser`, as `parse_str` is unable to handle a number of characters.

/cc @mtdowling


mtdowling:
I think this functionality should be brought back in some way, but this needs to be scaled back quite a bit. The first thing that comes to mind is the collection class and has data trait should not be brought back. If we GAE a mutable query class, it should be the bare minimum (I.e, no need for filter, map, etc).

jeskew:
True. My initial impulse was just to port the query classes and their ancestors whole cloth from Guzzle 5, but the collection operations aren't really in the spirit of Guzzle 6/PSR-7. The intended use case for `GuzzleHttp\Query` is that you pass into its constructor an array of a query *as you wish it to be sent*. The only responsibility of the `Query` class is to apply the appropriate encoder and aggregator.

Instances are not immutable, but they can only be mutated through `ArrayAccess` methods. This interface is implemented so that if you do want to evaluate path-style expressions against a query string, you can use JMESPath.

mtdowling:
Does this need to be updated to utilize `__toString()`?

How does this PR address https://github.com/aws/aws-sdk-php/issues/631?

jeskew:
There's a `__toString` on Query that returns a formatted querystring. 

https://github.com/aws/aws-sdk-php/pull/631 is currently blocked by the transformations `parse_str` performs on certain characters when they appear in a query string. This can result in presigned URLs being signed with a different query string than is actually sent. 

A query string parser could be written into the AWS SDK, but it makes more sense as part of Guzzle.

mtdowling:
I think my copy paste didn't work. Does this need to be updated to support `__toString()`: https://github.com/guzzle/guzzle/blob/master/src/Client.php#L347

jeskew:
I think this is a bit large given how much of the logic would be duplicated from `Psr7\parse_query` and `Psr7\build_query`. The only advantage the `Query` class has over those functions is the custom aggregator functionality, but I'm sure that can be handled in a more lightweight way.

PR #1262, opened 2015-09-19
rewind the stream after logging

mustafavelioglu:
Purpose of the change is rewinding the stream after using the response in log middleware

### DIFF
--- a/src/Middleware.php
+++ b/src/Middleware.php
@@ -175,6 +175,8 @@ final class Middleware
      * Middleware that logs requests, responses, and errors using a message
      * formatter.
      *
+     * If the stream of the request or response has consumed, rewinds the stream.
+     *
      * @param LoggerInterface  $logger Logs messages.
      * @param MessageFormatter $formatter Formatter used to create message strings.
      * @param string           $logLevel Level at which to log requests.
@@ -189,6 +191,8 @@ final class Middleware
                     function ($response) use ($logger, $request, $formatter, $logLevel) {
                         $message = $formatter->format($request, $response);
                         $logger->log($logLevel, $message);
+                        $request->getBody()->eof() && $request->getBody()->rewind();
+                        $response->getBody()->eof() && $response->getBody()->rewind();
                         return $response;
                     },
                     function ($reason) use ($logger, $request, $formatter) {
--- a/tests/MiddlewareTest.php
+++ b/tests/MiddlewareTest.php
@@ -174,6 +174,36 @@ class MiddlewareTest extends \PHPUnit_Framework_TestCase
         $this->assertContains('"PUT / HTTP/1.1" 200', $logger->output);
     }
 
+    public function testRewindStreamAfterLogsRequestsAndResponses()
+    {
+        $h = new MockHandler([new Response(200, [], 'bar')]);
+        $stack = new HandlerStack($h);
+        $logger = new Logger();
+        $formatter = new MessageFormatter(MessageFormatter::DEBUG);
+        $stack->push(Middleware::log($logger, $formatter));
+        $comp = $stack->resolve();
+        $p = $comp(new Request('GET', 'http://www.google.com', [], 'foo=bar'), []);
+        $response = $p->wait();
+        $this->assertContains('bar', $logger->output);
+        $this->assertContains('foo', $h->getLastRequest()->getBody()->getContents());
+        $this->assertContains('bar', $response->getBody()->getContents());
+    }
+
+    public function testLogDoesNotRewindStreamWhenRequestsAndResponsesNotUsed()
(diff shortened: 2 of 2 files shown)


mtdowling:
Wouldn't you only want to rewind if the body was consumed? Can you also add tests?

mustafavelioglu:
@mtdowling , thanks for feedback. 

log middleware rewinds the stream if body has consumed.

mtdowling:
I think this approach works for 

Actually, I think it might be better to first call `tell()` on each stream before calling log, then call `tell()` on each stream again, and only calling `rewind()` if the position of the stream changed since originally calling `tell()`. This approach will ensure that `rewind` is only called in the logger is actually responsible for consuming the stream.

mustafavelioglu:
Hi @mtdowling ,
Actually, I thought about your advice but i couldn't get a clear point on that. Because, formatter doesn't seek on the stream of request nor response, It consumes all things on it then we will have drained streams.
If I change the code regarding on your approach it will looks like this and calling the `tell()` four times seems bad to me.
Did i miss something ?
...
```php
function ($response) use ($logger, $request, $formatter, $logLevel) {
    $positionOfRequest = $request->getBody()->tell();
    $positionOfResponse = $response->getBody()->tell();
    $message = $formatter->format($request, $response);
    $logger->log($logLevel, $message);
    $afterLogPositionOfRequest = $request->getBody()->tell();
    $afterLogPositionOfResponse = $response->getBody()->tell();
    if($positionOfRequest <> $afterLogPositionOfRequest) {
        $request->getBody()->rewind();
    }
    if($positionOfResponse <> $afterLogPositionOfResponse) {
        $response->getBody()->rewind();
    }

    return $response;
}
```
...

mtdowling:
You're right, we're adding a lot of conditional code here... I'm starting to lean away from this change being accepted. If we did add it though, it would be something like this:

```php
function ($response) use ($logger, $request, $formatter, $logLevel) {
    $requestPos = $request->getBody()->tell();
    $responsePos = $response->getBody()->tell();
    $message = $formatter->format($request, $response);
    $logger->log($logLevel, $message);
    if ($request->getBody()->isSeekable() && $requestPos != $request->getBody()->tell()) {
        $request->getBody()->rewind();
    }
    if ($response->getBody()->isSeekable() && $responsePos != $response->getBody()->tell()) {
        $response->getBody()->rewind();
    }
    return $response;
}
```

@jeremeamia @jeskew: What do y'all think about this change? Is it necessary? Do we need to rewind the streams after usage or is this something the end users should manage by rewinding themselves or by injecting another middleware that rewinds? Should this change be made to the MessageFormatter instead?

jeskew:
This change or something like it does seem necessary. Users shouldn't have to know how much was logged in order to determine how to use a response object.

I would personally prefer to see it in the MessageFormatter class, as not all logging types cause the request and/or response bodies to be consumed.

EAnushan:
I don't think putting it in MessageFormatter is a bad idea. At the end of the day though, I do think the Log middleware should be responsible for rewinding the stream. It would be very unintuitive if the user had to manually add a rewind middleware after the log middleware every time.

gmponos:
Let me also add that I believe that this MR is a little related with an issue I have created on guzzle/psr-7 
https://github.com/guzzle/psr7/issues/101

sagikazarmark:
The problem with streams is that they are not immutable. Which means if you rewind a stream it modifies the state of the request. This is fine, until you can get into the same state. But this is not always the case. For example: some streams are not rewindable, which means logging would make the request unreadable.

The other issue is that if the stream body is large, it could cause memory problems too.

The PHP HTTP solution for this problem: https://github.com/php-http/message/blob/master/src/Formatter/FullHttpMessageFormatter.php#L78

https://github.com/php-http/message/blob/master/src/Stream/BufferedStream.php

alborq:
Same problem discover today, I'm sad to see that, Hard to understand why apps is off with adding logger middleware. This PR may be usefull..

sagikazarmark:
I am not really a fan of the current state of this PR. If we would like to solve it somehow, I would go with the solutions proposed above.

klaude:
I've run into this issue recently at work. If nobody objects I'm comfortable with implementing @sagikazarmark 's recommendation, `rewind()`ing the stream if it's rewindable. I'll have to do it in a different pull request. Do y'all mind that?

Nyholm:
:+1:

juniorb2ss:
I had the same problem using version `^6.3`. It was necessary to invoke the method `$response-> getBody()->rewind();` before call `$response->getBody()->getContents();`

Nyholm:
Im closing this because work has continued in #1842

PR #1271, opened 2015-10-05
Makes the HandlerStack more reusable

### LINKED ISSUE #1263: Make HandlerStack more flexible
Right now the [`HandlerStack`](https://github.com/aws/aws-sdk-php/blob/master/src/HandlerList.php) class has an `__invoke` method signature of

```php
public function __invoke(RequestInterface $request, array $options) : Promise
```

This works well as a middleware registration system for the Guzzle client and the `Request -> Promise<Response>` lifecycle; however, I think the `HandlerStack` could be useful in other contexts, too.

I've been starting to think about the [Guzzle Command](https://github.com/guzzle/command) project and how to update for Guzzle 6. Some of the big changes include using middleware instead of events and using the new promises instead of futures. Just like there is a `HandlerStack` associated with a Guzzle HTTP client, I think it makes sense to carry that pattern over to the Command package and have a HandlerStack associated with a Guzzle Service Client. The `HandlerStack` logic seems to work just fine for this case; the only issue is that the method signature on the `__invoke` method couples it to `Request` objects. In the case of the Command package, you are worried about the `Command -> Promise<Result>`.

However, there is a simple, backward-compatible change we could do to make the `HandlerStack` useful outside of the Request/Response context.

If we move the logic related to `$this->cached` from the `__invoke` method to `resolve`, then someone can use the entire `HandlerStack`'s features without having to use `__invoke`.

What do you think?

jeremeamia:
Fixes #1263. Makes the HandlerStack more reusable (without BC) by moving the "cached" logic from __invoke() to resolve().

### DIFF
--- a/src/HandlerStack.php
+++ b/src/HandlerStack.php
@@ -62,11 +62,8 @@ class HandlerStack
      */
     public function __invoke(RequestInterface $request, array $options)
     {
-        if (!$this->cached) {
-            $this->cached = $this->resolve();
-        }
+        $handler = $this->resolve();
 
-        $handler = $this->cached;
         return $handler($request, $options);
     }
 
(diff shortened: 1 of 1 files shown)


mtdowling:
Cool thanks

PR #1281, opened 2015-10-16
Added some Docblocks

vanbrabantf:
### Added
Added some docblock suggar to the methodes

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -87,11 +87,6 @@ class Client implements ClientInterface
             : $this->request($method, $uri, $opts);
     }
 
-    /**
-     * @param RequestInterface $request
-     * @param array $options
-     * @return Promise\PromiseInterface
-     */
     public function sendAsync(RequestInterface $request, array $options = [])
     {
         // Merge the base URI into the request URI if needed.
@@ -103,23 +98,12 @@ class Client implements ClientInterface
         );
     }
 
-    /**
-     * @param RequestInterface $request
-     * @param array $options
-     * @return Promise\PromiseInterface
-     */
     public function send(RequestInterface $request, array $options = [])
     {
         $options[RequestOptions::SYNCHRONOUS] = true;
         return $this->sendAsync($request, $options)->wait();
     }
 
-    /**
-     * @param string $method
-     * @param UriInterface $uri
-     * @param array $options
-     * @return Promise\PromiseInterface
-     */
     public function requestAsync($method, $uri = null, array $options = [])
     {
         $options = $this->prepareDefaults($options);
@@ -139,22 +123,12 @@ class Client implements ClientInterface
         return $this->transfer($request, $options);
     }
 
-    /**
-     * @param string $method
-     * @param UriInterface $uri
-     * @param array $options
-     * @return Promise\PromiseInterface
-     */
     public function request($method, $uri = null, array $options = [])
     {
         $options[RequestOptions::SYNCHRONOUS] = true;
         return $this->requestAsync($method, $uri, $options)->wait();
     }
 
-    /**
-     * @param $option
-     * @return array|null
-     */
     public function getConfig($option = null)
     {
         return $option === null
--- a/src/Client.php
+++ b/src/Client.php
@@ -136,11 +136,6 @@ class Client implements ClientInterface
             : (isset($this->config[$option]) ? $this->config[$option] : null);
     }
 
-    /**
-     * @param $uri
-     * @param array $config
-     * @return UriInterface
-     */
     private function buildUri($uri, array $config)
     {
         if (!isset($config['base_uri'])) {
--- a/src/Client.php
+++ b/src/Client.php
@@ -165,7 +165,7 @@ class Client implements ClientInterface
     /**
      * @param $uri
      * @param array $config
-     * @return Psr7\Uri|UriInterface
+     * @return UriInterface
      */
     private function buildUri($uri, array $config)
     {
--- a/src/Client.php
+++ b/src/Client.php
@@ -87,6 +87,11 @@ class Client implements ClientInterface
             : $this->request($method, $uri, $opts);
     }
 
+    /**
+     * @param RequestInterface $request
+     * @param array $options
+     * @return Promise\PromiseInterface
+     */
     public function sendAsync(RequestInterface $request, array $options = [])
     {
         // Merge the base URI into the request URI if needed.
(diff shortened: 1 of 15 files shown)


src/Client.php
```diff
@@ -136,6 +162,11 @@ public function getConfig($option = null)
             : (isset($this->config[$option]) ? $this->config[$option] : null);
     }
 
+    /**
+     * @param $uri
+     * @param array $config
+     * @return Psr7\Uri|UriInterface
```
hannesvdvreken: I'd stick to `UriInterface`

src/Handler/MockHandler.php
```diff
@@ -149,6 +154,12 @@ public function count()
         return count($this->queue);
     }
 
+    /**
+     * @param RequestInterface $request
+     * @param array $options
+     * @param ResponseInterface|null $response
```
hannesvdvreken: I'd leave off the `|null`

src/Handler/StreamHandler.php
```diff
@@ -127,6 +141,11 @@ private function createResponse(
         return new FulfilledPromise($response);
     }
 
+    /**
+     * @param StreamInterface $stream
+     * @param array $options
+     * @return Psr7\Stream|StreamInterface
```
hannesvdvreken: Same here: `StreamInterface` is all the requester needs to know.

src/RedirectMiddleware.php
```diff
@@ -115,6 +115,11 @@ public function checkRedirect(
         return $promise;
     }
 
+    /**
+     * @param PromiseInterface $promise
+     * @param $uri
+     * @return mixed
```
hannesvdvreken: $promise->then will always return `PromiseInterface` see https://github.com/guzzle/promises/blob/master/src/PromiseInterface.php#L26-L28

src/Handler/MockHandler.php
```diff
@@ -59,6 +59,11 @@ public function __construct(
         }
     }
 
+    /**
+     * @param RequestInterface $request
+     * @param array $options
+     * @return mixed
```
hannesvdvreken: afaik middlewares return objects which implement `ResponseInterface`

src/RetryMiddleware.php
```diff
@@ -102,6 +112,11 @@ private function onRejected(RequestInterface $req, array $options)
         };
     }
 
+    /**
+     * @param RequestInterface $request
+     * @param array $options
+     * @return mixed
```
hannesvdvreken: same here: `@return \Psr\Http\Message\RequestInterface` (or import it).

src/UriTemplate.php
```diff
@@ -35,6 +35,11 @@ class UriTemplate
         '%40', '%21', '%24', '%26', '%27', '%28', '%29', '%2A', '%2B', '%2C',
         '%3B', '%3D');
 
+    /**
+     * @param $template
+     * @param array $variables
+     * @return mixed
+     */
```
hannesvdvreken:
```php
/**
 * @param string $template
 * @param array $variables
 *
 * @return string
 */
```

hannesvdvreken:
Try not to leave the return type undefined. `@return mixed` is not very useful to anyone. I explained one use case in this blog post: http://blog.madewithlove.be/post/return-types/

Thanks for the docblocks! Super useful for my IDE to help me use these methods without having to dive into the source.

mtdowling:
The Guzzle project does not place docblocks on methods that are documented by their interface. Newer versions of PHPStorm and possibly other PHP IDE's now properly show interface documentation on implementations that have no docblock.

Any of the docblocks you've added that are not covered by an interface or parent's class docblock are good additions though. I'd be happy to merge a PR that adds those kinds of missing docblocks.

hannesvdvreken:
So, not even a docblock like this:

```php
/**
 * {@inherit}
 */
```

Ok :-) @vanbrabantf this is project specific advice from the maintainer. Too bad I can't find that guideline in a contributing.md file. There's still a good chance to contribute here! Go for it.

mtdowling:
That's a good point. I need to add a contributing.md file with this called out.

vanbrabantf:
Should I include the change to the "contributing.md" in my next commit? No problem, I'll just use: 

"The Guzzle project does not place docblocks on methods that are documented by their interface. Newer versions of PHPStorm and possibly other PHP IDE's now properly show interface documentation on implementations that have no docblock."

hannesvdvreken:
@vanbrabantf Maybe in a separate PR. There is no contributing.md file yet. It will need to contain more than that sentence.

vanbrabantf:
I updated the PR with only the methodes that are not covered with the interface. These are mostly private functions so I'm not sure how useful those are.

mtdowling:
Thanks for the modifications. I think the docblocks that you added to private methods are unnecessary -- they don't add anything over the method signature. Can you remove those docblocks?

mtdowling:
If we're going to add docs, we shouldn't just add them for the sake of having a docblock. Let's actually document the parameters. If we aren't documenting the parameters, then adding a docblock doesn't provide any value because everything can be inferred already by an IDE.

jfountain:
In regards to contributions guidelines, this is a good place to start...

http://todogroup.org/opencodeofconduct/

src/RetryMiddleware.php
```diff
@@ -101,7 +101,7 @@ private function onRejected(RequestInterface $req, array $options)
             return $this->doRetry($req, $options);
         };
     }
-
+
```
mtdowling: extra spaces

src/Handler/EasyHandle.php
```diff
@@ -77,6 +77,9 @@ public function createResponse()
         );
     }
 
+    /**
+     * @param string $name
```
mtdowling: I think this docblock is unnecessary

src/Handler/CurlMultiHandler.php
```diff
@@ -41,6 +41,10 @@ public function __construct(array $options = [])
             ? $options['select_timeout'] : 1;
     }
 
+    /**
+     * @param $name
+     * @return resource
+     */
     public function __get($name)
```
mtdowling: I think this docblock is unnecessary

src/Cookie/SetCookie.php
```diff
@@ -82,6 +82,9 @@ public function __construct(array $data = [])
         }
     }
 
+    /**
+     * @return string
+     */
     public function __toString()
```
mtdowling: I think this docblock is unnecessary

src/Handler/CurlMultiHandler.php
GrahamCampbell: This docblock is also incorrectly formatted. The type is always required.

mtdowling:
I'll go ahead and close this. I think there are several additions here that aren't necessary. There are a couple good docblocks, but I think those could be submitted in a fresh PR.

PR #1282, opened 2015-10-16
Switch Middleware to use exception factory

shadowhand:
The `RequestException::create` factory was not being used anywhere.
Using it inside of `Middleware::httpErrors` simplifies the code and
allows for simpler changes in the future.

Also updates the message formatting of the exception factory to
provide a more complete status, including the body of the response.

### DIFF
--- a/src/Exception/RequestException.php
+++ b/src/Exception/RequestException.php
@@ -79,22 +79,61 @@ class RequestException extends TransferException
 
         $level = floor($response->getStatusCode() / 100);
         if ($level == '4') {
-            $label = 'Client error response';
+            $label = 'Client Error';
             $className = __NAMESPACE__ . '\\ClientException';
         } elseif ($level == '5') {
-            $label = 'Server error response';
+            $label = 'Server Error';
             $className = __NAMESPACE__ . '\\ServerException';
         } else {
-            $label = 'Unsuccessful response';
+            $label = 'Unsuccessful Request';
             $className = __CLASS__;
         }
 
-        $message = $label . ' [url] ' . $request->getUri()
-            . ' [http method] ' . $request->getMethod()
-            . ' [status code] ' . $response->getStatusCode()
-            . ' [reason phrase] ' . $response->getReasonPhrase();
+        // Server Error: `GET /` resulted in a `404 Not Found` response:
+        // <html> ... (truncated)
+        $message = sprintf(
+            '%s: `%s` resulted in a `%s` response',
+            $label,
+            $request->getMethod() . ' ' . $request->getUri(),
+            $response->getStatusCode() . ' ' . $response->getReasonPhrase()
+        );
+
+        $summary = static::getResponseBodySummary($response);
+
+        if (is_string($summary)) {
+            $message .= ":\n{$summary}\n";
+        }
 
         return new $className($message, $request, $response, $previous, $ctx);
+    }
+
+    /**
+     * Get a short summary of the response
+     *
+     * Will return `null` if the response is not printable.
+     *
+     * @param ResponseInterface $response
+     *
+     * @return string|null
+     */
+    public static function getResponseBodySummary(ResponseInterface $response)
+    {
+        $body = $response->getBody();
+
+        if (!$body->isSeekable()) {
+            return null;
+        }
+
+        $summary = $body->read(120);
+        $body->rewind();
+
+        // Matches any printable character, including unicode characters:
+        // letters, marks, numbers, punctuation, spacing, and separators.
+        if (preg_match('/[^\pL\pM\pN\pP\pS\pZ\n\r\t]/', $summary)) {
+            return null;
+        }
+
+        return $summary;
     }
 
     /**
--- a/tests/Exception/RequestExceptionTest.php
+++ b/tests/Exception/RequestExceptionTest.php
@@ -31,8 +31,12 @@ class RequestExceptionTest extends \PHPUnit_Framework_TestCase
     public function testCreatesClientErrorResponseException()
     {
         $e = RequestException::create(new Request('GET', '/'), new Response(400));
-        $this->assertEquals(
-            'Client error response [url] / [http method] GET [status code] 400 [reason phrase] Bad Request',
+        $this->assertContains(
+            'GET /',
+            $e->getMessage()
+        );
+        $this->assertContains(
+            '400 Bad Request',
             $e->getMessage()
         );
         $this->assertInstanceOf('GuzzleHttp\Exception\ClientException', $e);
@@ -41,8 +45,12 @@ class RequestExceptionTest extends \PHPUnit_Framework_TestCase
     public function testCreatesServerErrorResponseException()
     {
         $e = RequestException::create(new Request('GET', '/'), new Response(500));
-        $this->assertEquals(
-            'Server error response [url] / [http method] GET [status code] 500 [reason phrase] Internal Server Error',
+        $this->assertContains(
+            'GET /',
+            $e->getMessage()
+        );
+        $this->assertContains(
+            '500 Internal Server Error',
             $e->getMessage()
         );
         $this->assertInstanceOf('GuzzleHttp\Exception\ServerException', $e);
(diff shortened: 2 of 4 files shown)


mtdowling:
I like the factory change a lot.

I'm open to modifying the exception message, but I'm not sure that the response body should be part of the message. There have been several issues over time related to exception messages being too large because the response body is part of the message. It looks like the `$label` variable is no longer part of the message-- can we make sure to include that? I wonder if the message should be more human readable too-- something like `Server error response: sending a GET request to /xyz resulted in a 500 Internal Server Error response`

The tests will need to be updated for this change as well.

src/Exception/RequestException.php
```diff
@@ -79,20 +79,25 @@ public static function create(
...
-            . ' [status code] ' . $response->getStatusCode()
-            . ' [reason phrase] ' . $response->getReasonPhrase();
+        // Server Error: `GET /` resulted in a `404 Not Found` response:
+        // <html> ... (truncated)
+        $message = sprintf(
+            "%s: `%s` resulted in a `%s` response:\n%s",
+            $label,
+            $request->getMethod() . ' ' . $request->getUri(),
+            $response->getStatusCode() . ' ' . $response->getReasonPhrase(),
+            substr((string) $response->getBody(), 0, 80)
```
shadowhand: @mtdowling how does this look? When working with a JSON response, it can be really useful to have part of the response body... I limited the result to the first 80 chars.

mtdowling: Sure, having a limited snippet of the response body could work, but there are some caveats we should consider:

1. We should use the `read` method of the response body instead of casting it to a string to ensure that we aren't loading more data into memory than needed.
2. We should only include the response body in the message if the body is seekable. If it is, and we read from the response, we should rewind the stream. This allows users to work with the stream in error handling.
3. This is maybe not necessary, but it could possibly be an issue for weirder APIs. Should we make sure that the response is printable? Maybe grab the first 80 chars and only include the response body if `ctype_print()` returns true.

jeremeamia: I agree with those three. Regarding **3** specifically, I recently worked with an API where the payload was in `igbinary` format, so that type of situation is not unheard of.

shadowhand: The main problem with `ctype_print` is that it returns false for any control character, including `\n` and `\t`. Maybe we should only include the summary for known content types?

src/Exception/RequestException.php
```diff
@@ -79,25 +79,86 @@ public static function create(
...
+     */
+    protected static function isResponsePrintable(ResponseInterface $response)
+    {
+        $types = [
+            // Plain text, javascript, css, etc
+            'text/.*',
+            // JSON of various types
+            'application/.*json',
+            // XML of various types, including XHTML
+            'application/.*xml',
```
shadowhand: @mtdowling @jeremeamia I think this should be a sufficient whitelist to start with for printable.

tests/Exception/RequestExceptionTest.php
```diff
@@ -51,8 +59,58 @@ public function testCreatesServerErrorResponseException()
...
+    public function dataPrintableResponses()
+    {
+        return [
+            ['text/plain', 'You broke the test!'],
+            ['text/html', '<h1>It Broken!</h1>'],
+            ['text/xml', '<xml><text>Your friendly test</text></xml>'],
+            ['text/javascript', 'document.body.write("here comes a test");'],
+            ['text/css', 'body:before { content: "test style" }'],
+            ['application/json', '{"type": "test"}'],
+            ['application/xhtml+xml', '<p>yet another<br/>test!</p>'],
```
shadowhand: This covers all of the default types. Should anything else be added?

tests/Exception/RequestExceptionTest.php
```diff
@@ -51,8 +59,58 @@ public function testCreatesServerErrorResponseException()
...
+        );
+        $this->assertInstanceOf('GuzzleHttp\Exception\RequestException', $e);
+    }
+
+    public function testCreatesExceptionWithoutPrintableBody()
+    {
+        $response = new Response(
+            500,
+            ['Content-Type' => 'image/gif'],
+            $content = base64_decode('R0lGODlhAQABAIAAAAAAAP///yH5BAEAAAAALAAAAAABAAEAAAIBRAA7') // 1x1 gif
```
shadowhand: https://css-tricks.com/snippets/html/base64-encode-of-1x1px-transparent-gif/

src/Exception/RequestException.php
mtdowling: I think this is too limiting because unfortunately, some web services use funky content-types or don't even provide them sometimes (e.g., AWS). I also think maintaining a list of content-types here could lead to bloat over time (e.g., someone saying please include CSV, text, html, etc...)

I think we can do a regex on the first N characters to ensure that they are valid. Something like this could work (untested): 

```
$chunk = $response->read(80);
return preg_match('/^[\r\n\t[:print:]]*$/', $response) ? $chunk : null;
```

src/Exception/RequestException.php
```diff
@@ -79,25 +79,64 @@ public static function create(
...
+        if (!$body->isSeekable()) {
+            return null;
+        }
+
+        $summary = $body->read(160);
+        $body->rewind();
+
+        // Matches any printable character, including unicode characters:
+        // letters, marks, numbers, punctuation, spacing, and separators.
+        if (preg_match('/[^\pL\pM\pN\pP\pS\pZ\n\r\t]/', $summary)) {
```
shadowhand: @mtdowling how about this?

tests/Exception/RequestExceptionTest.php
```diff
@@ -51,8 +59,57 @@ public function testCreatesServerErrorResponseException()
...
+            $e->getMessage()
+        );
+        $this->assertInstanceOf('GuzzleHttp\Exception\RequestException', $e);
+    }
+
+    public function dataPrintableResponses()
+    {
+        return [
+            ['You broke the test!'],
+            ['<h1>zlomený zkouška</h1>'],
```
shadowhand: I also added unicode characters into the test for verification.

tests/Exception/RequestExceptionTest.php
```diff
@@ -51,8 +59,57 @@ public function testCreatesServerErrorResponseException()
...
+            ['{"tester": "Philépe Gonzalez"}'],
+            ["<xml>\n\t<text>Your friendly test</text>\n</xml>"],
```
shadowhand: And newlines, tabs.

src/Exception/RequestException.php
mtdowling: Nice! Given that PHP has a default [character limit](http://php.net/manual/en/errorfunc.configuration.php#ini.log-errors-max-len) for printing/logging exception messages of 1024 characters, I wonder if 160 characters of the body is too long. What do you think?

Unless you have a use case for it, I also don't think this method needs to be protected.

shadowhand: I don't want to reduce it less than 120 characters... the current formatting would be ~300 characters at most.

src/Exception/RequestException.php
```diff
@@ -79,25 +79,64 @@ public static function create(
...
+     */
+    public static function getResponseBodySummary(ResponseInterface $response)
+    {
+        $body = $response->getBody();
+
+        if (!$body->isSeekable()) {
+            return null;
+        }
+
+        $summary = $body->read(120);
```
shadowhand: Dropped to 120.

src/Exception/RequestException.php
```diff
@@ -79,25 +79,64 @@ public static function create(
...
     /**
+     * Get a short summary of the response
+     *
+     * Will return `null` if the response is not printable.
+     *
+     * @param ResponseInterface $response
+     *
+     * @return string|null
+     */
+    public static function getResponseBodySummary(ResponseInterface $response)
```
shadowhand: Made public.

shadowhand:
@mtdowling anything else I need to do to get this accepted? It is causing us some issues debugging.

mtdowling:
I made a couple small tweaks and merged. Thanks!

PR #1285, opened 2015-10-20
Documented throwing an exception in MockHandler

tpavlek:
While exceptions are briefly mentioned in the preceding documentation, it is not immediately clear that exceptions can be thrown in the MockHandler, since my eyes are immediately drawn to the example code rather than the textual pre-amble.

With this, it is immediately clear to anyone that the stack can consist of both Responses and Exceptions.


mtdowling:
Thanks!

Issue #1256, opened 2015-09-11
verify = FALSE with StreamHandler still results in certificate verification. Probably, ['ssl']['verify_peer_name'] should be set to FALSE as well


drscre:
I am using StreamHandler to make an https request with `verify = FALSE` option.
But I still get "Peer certificate CN=... did not match expected CN=..." error.

I managed to fix this by adding `['stream_context']['ssl']['verify_peer_name'] = FALSE`

Probably, we should add 
`$options['ssl']['verify_peer_name'] = false;`
 to [StreamHandler::add_verify()](https://github.com/guzzle/guzzle/blob/master/src/Handler/StreamHandler.php#L361)

mtdowling:
Sounds good. Could be added around here: https://github.com/guzzle/guzzle/blob/master/src/Handler/StreamHandler.php#L361

PR #1286, opened 2015-10-21
https://github.com/guzzle/guzzle/issues/1256

daleattree:
Code added to resolve issue 1256

### DIFF
--- a/src/Handler/StreamHandler.php
+++ b/src/Handler/StreamHandler.php
@@ -359,6 +359,7 @@ class StreamHandler
             }
         } elseif ($value === false) {
             $options['ssl']['verify_peer'] = false;
+            $options['ssl']['verify_peer_name'] = false;
             return;
         } else {
             throw new \InvalidArgumentException('Invalid verify request option');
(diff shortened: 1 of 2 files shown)


mtdowling:
Looks like a test failure.

daleattree:
Added an incorrect assertion into the test. Apologies for that. Tests are working.

mtdowling:
Can you add a test or update an existing test to ensure that this value is set to false when verify is set to false?

daleattree:
StreamHandlerTest.php updated to ensure both true and false for SSL verify_peer_name

mtdowling:
Should a test be added to https://github.com/guzzle/guzzle/pull/1286/files#diff-25965a11c4d928e67905c6d782fa52e3R230 that asserts that `verify_peer_name` was set to `false`?

Blackout:
I've added some code that handles the SSL Stream Context Options in my pull request

The Tests have been updated too

https://github.com/guzzle/guzzle/pull/1298

PR #1287, opened 2015-10-22
Handle disabled functions with curl_exec() and curl_multi_exec()

dawehner:
On shared hosting providers disable curl_exec() and curl_multi_exec() so the choose_handler() function
has to take that into account.

### DIFF
--- a/src/functions.php
+++ b/src/functions.php
@@ -105,7 +105,12 @@ function choose_handler()
 {
     $handler = null;
     if (extension_loaded('curl')) {
-        $handler = Proxy::wrapSync(new CurlMultiHandler(), new CurlHandler());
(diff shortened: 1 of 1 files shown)


mtdowling:
thanks!

dawehner:
Thanks you!

src/functions.php
```diff
@@ -104,8 +104,12 @@ function debug_resource($value = null)
 function choose_handler()
 {
     $handler = null;
-    if (extension_loaded('curl')) {
+    if (function_exists('curl_multi_exec') && function_exists('curl_exec')) {
```
dawehner: This should have gotten some documentation IMHO, but well, if this is fine for you @mtdowling

mtdowling:
A PR that adds relevant documentation would be fine with me :)

PR #1292, opened 2015-10-30
Implement "sink" for MockHandler

ivank:
This is a reimplementation of my previous pull request https://github.com/guzzle/guzzle/pull/1020 which regressed with the 6.0 rewrite

Fingers crossed for 7.0, maybe I can make one of those too :)

### DIFF
--- a/src/Handler/MockHandler.php
+++ b/src/Handler/MockHandler.php
@@ -87,6 +87,19 @@ class MockHandler implements \Countable
                 if ($this->onFulfilled) {
                     call_user_func($this->onFulfilled, $value);
                 }
+                if (isset($options['sink'])) {
+                    $contents = (string) $value->getBody();
+                    $sink = $options['sink'];
+
+                    if (is_resource($sink)) {
+                        fwrite($sink, $contents);
+                    } elseif (is_string($sink)) {
+                        file_put_contents($sink, $contents);
+                    } elseif ($sink instanceof \Psr\Http\Message\StreamInterface) {
+                        $sink->write($contents);
+                    }
+                }
+
                 return $value;
             },
             function ($reason) use ($request, $options) {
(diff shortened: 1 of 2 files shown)


src/Handler/MockHandler.php
```diff
@@ -87,6 +87,19 @@ function ($value) use ($request, $options) {
                 if ($this->onFulfilled) {
                     call_user_func($this->onFulfilled, $value);
                 }
+                if (isset($options['sink'])) {
+                    $contents = (string) $value->getBody();
+                    $sink = $options['sink'];
+
+                    if (is_resource($sink)) {
```
mtdowling: I think this could be cleaned up and made more memory efficient for files and stream instances by updating this to something like:

```php
$sink = $options['sink'];

if (is_string($sink)) {
    file_put_contents($sink, $value->getBody());
} else {
    $contents = Psr7\stream_for($sink);
    Psr7\copy_to_stream($body->getContents(), $sink);
}
```

mtdowling:
Thanks for the PR, and sorry for the regression. I made one small comment on the PR.

src/Handler/MockHandler.php
ivank: Yeah I've started with that implementation, but found out that creating a new stream (Psr7\stream_for) had some issues. I use a tmpfile which deletes the file once the stream is closed, which happens on Stream::__destroy. Any suggestion how this can be fixed?

mtdowling: PHP tmp stream only create a temp file when the stream exceeds a certain size (2 megs by default I believe). Are you saying that you're seeing a temp file created no matter what or that you had to manually create a temp file to make what I suggested work?

ivank: The way I'm using it is "create a tmpfile" pass it as a "sink" parameter (its a smallish zip file that I want downloaded and read, but to make sure it gets deleted afterwards). 
I use get_stream_meta() to get the location of the actual file in the system and then use the php's Zip reading lib to do the job. 
This all works fine, but I couldn't test it out with a mocked request. And implementing it with `stream_for` closes the stream before I could pass it to zip library, giving me a bunch of errors. The way shown in this pull request works fine. 
You can test it out yourself, as the test here is basically doing the same.

PR #1295, opened 2015-11-02
Refactoring

Imangazaliev:


mtdowling:
I'm not sure why this change is necessary. Can you elaborate? I personally don't find this change to help with readability, and I don't think it addresses a bug.

Imangazaliev:
I think it will work faster

mtdowling:
If this change is faster (which it definitely wouldn't be for every case), it would be insignificant. I'm going to pass on this change, but thanks for trying to help.

PR #1298, opened 2015-11-06
SSL Stream Context Options Expanded

Blackout:
* Added support for SSL Stream Context Options 'allow_self_signed' & 'verify_peer_name' via options array
* Added StreamHandler::add_allow_self_signed() & StreamHandler::add_verify_peer_name()
* Cleaned up StreamHandler::add_verify() as some code was unreachable

Example usage:

```
// Allow self signed certificates & dont verify the peer name
$client = new GuzzleHttp\Client(['allow_self_signed' => true, 'verify_peer_name' => false]);
```

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -161,7 +161,7 @@ class Client implements ClientInterface
             'decode_content'	=> true,
             'verify'		=> true,
             'cookies'		=> false,
-            'allow_self_signed'	=> false,
+            'allow_self_signed' => false,
             'verify_peer_name'	=> false,
         ];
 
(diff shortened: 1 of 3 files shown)


Blackout:
Sorry about the amount of file updates, first time using GIT

Blackout:
* Updated Stream Handler Test's

mtdowling:
It's already possible to add custom stream handler options: http://docs.guzzlephp.org/en/latest/faq.html#how-can-i-add-custom-stream-context-options. I don't think the parameters added in this pull request are easily mapped to curl (verify peer is actually already an option covered by the `verify` request option).

I think the best thing to do is just use custom stream context options.

Blackout:
You are correct, thank you for the reply

PR #1309, opened 2015-11-13
Quoted big int in json

akaraznevic:
According to http://stackoverflow.com/a/27909889

This error seems to come from a buggy version of pecl-json-c, which Ubuntu packages as an alias for php5-json because of licensing issues.


mtdowling:
I'm not sure this is something that needs to be added to Guzzle considering that you can easily decode the body of the response yourself without using the `json()` method of a response. In fact, the `json()` and `xml()` methods on responses were removed in Guzzle 6 specifically for cases like this.

I think I prefer to keep this kind of version specific / platform specific logic out for now.

PR #1312, opened 2015-11-17
Reworked options keys to be used from RequestOptions capsule.

xbugster:
Changed the Client and other classes who used $options with hard coded strings as keys to use RequestOptions constants as keys, as they are equal and explained in documentation, but the constants itself almost never used in code, for example !empty($option['headers']) became !empty(RequestOptions::HEADERS), etc.
Thank You.

### DIFF
--- a/src/RequestOptions.php
+++ b/src/RequestOptions.php
@@ -151,6 +151,14 @@ final class RequestOptions
     const MULTIPART = 'multipart';
 
     /**
+     * on_redirect: (callable) PHP callable that is invoked when a redirect
+     * is encountered. The callable is invoked with the request, the redirect
+     * response that was received, and the effective URI. Any return value
+     * from the on_redirect function is ignored.
+     */
+    const ON_REDIRECT = 'on_redirect';
+
+    /**
      * on_headers: (callable) A callable that is invoked when the HTTP headers
      * of the response have been received but the body has not yet begun to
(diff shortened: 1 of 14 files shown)


src/RequestOptions.php
```diff
@@ -151,6 +156,14 @@
     const MULTIPART = 'multipart';
 
     /**
+     * on_redirect: (callable) PHP callable that is invoked when a redirect
+     * is encountered. The callable is invoked with the request, the redirect
+     * response that was received, and the effective URI. Any return value
+     * from the on_redirect function is ignored.
+     */
+    const ON_REDIRECT = 'on_redirect';
```
mtdowling: I'm -1 on adding `on_redirect` and `retries` to RequestOptions. These options are basically children of another option, and have no effect on sending a request unless you enable their parent options (e.g., `allows_redirects`) or their associated middleware (`RetryMiddleware`).

mtdowling:
Thanks for the PR. Can you make sure lines are no longer than 80 characters? To achieve this without making the code ugly, you could try aliasing `RequestOptions` as `Options` when importing it. Also, can you squash this into a single commit please?

xbugster:
will do later today and push

mtdowling:
Closing as this will need to be rebased and resubmitted as a single commit.

PR #1316, opened 2015-11-19
Use full name for response variable in example

bocharsky-bw:


bocharsky-bw:
@mtdowling Why does it closed? The `$res` variable name a bit confuse. It's could be consider as `result` not `response`.

PR #1318, opened 2015-11-20
Added convenience method to access a cookie by name

deepdivedylan:
I found this method useful to not have to manually iterate the cookie array.


tests/Cookie/CookieJarTest.php
```diff
@@ -45,6 +45,18 @@ public function testCreatesFromArray()
         $this->assertCount(2, $jar);
     }
 
+    public function testGetsCookiesByName()
+    {
+        $cookies = $this->getTestCookies();
+        foreach ($cookies as $cookie) {
+            $this->assertTrue($this->jar->setCookie($cookie));
```
mtdowling: No need for this assertion as that's not what's being tested.

src/Cookie/CookieJar.php
```diff
@@ -98,6 +98,25 @@ public static function shouldPersist(
...
+    {
+        $foundCookie = null;
+        foreach($this->cookies as $cookie) {
+            if(strtolower($cookie->getName()) === strtolower($name)) {
+                $foundCookie = $cookie;
+                break;
+            }
+        }
+
+        return($foundCookie);
```
mtdowling: no need for the `(` `)`

src/Cookie/CookieJar.php
```diff
@@ -98,6 +98,25 @@ public static function shouldPersist(
...
+     *
+     * @param string $name cookie name to search for
+     * @return SetCookie|null cookie that was found or null if not found
+     */
+    public function getCookieByName($name)
+    {
+        $foundCookie = null;
+        foreach($this->cookies as $cookie) {
+            if(strtolower($cookie->getName()) === strtolower($name)) {
+                $foundCookie = $cookie;
```
mtdowling: You can just return early here: `return $cookie`

src/Cookie/CookieJar.php
```diff
@@ -98,6 +98,25 @@ public static function shouldPersist(
...
 
+    /**
+     * Finds and returns the cookie based on the name
+     *
+     * @param string $name cookie name to search for
+     * @return SetCookie|null cookie that was found or null if not found
+     */
+    public function getCookieByName($name)
+    {
+        $foundCookie = null;
```
mtdowling: No need for a state variable like this. You can just return early.

src/Cookie/CookieJar.php
```diff
@@ -98,6 +98,25 @@ public static function shouldPersist(
...
+        foreach($this->cookies as $cookie) {
```
mtdowling: Please add a space between `foreach` and the opening `(`

src/Cookie/CookieJar.php
```diff
@@ -98,6 +98,25 @@ public static function shouldPersist(
...
+            if(strtolower($cookie->getName()) === strtolower($name)) {
```
mtdowling: Please add a space between `if` and the opening `(`.

I think `strcasecmp` would be nicer here as well.

src/Cookie/CookieJar.php
PhilippSchaffrath: just delete it then, since the method will always return null if nothing else is returned

sagikazarmark:
@dylan-mcdonald are you willing to wrap up this PR? Thanks

src/Cookie/CookieJar.php
```diff
@@ -86,6 +86,25 @@ public static function shouldPersist(
         return false;
     }
 
+    /**
+     * Finds and returns the cookie based on the name
+     *
+     * @param string $name cookie name to search for
+     * @return SetCookie|null cookie that was found or null if not found
+     */
+    public function getCookieByName($name)
```
kirkegaard: Wouldn't `getCookie` be a better name since it matches the setCookie method?

sagikazarmark: Not sure, setCookie has a completely different behaviour than the name suggests, so it might be confusing.

dylan-mcdonald:
@sagikazarmark Here's an update for the pull request.

dylan-mcdonald:
What is the status on this? Is there anything more I can do for you guys? Otherwise, let's go ahead an merge this. :smiley_cat:

src/Cookie/CookieJar.php
```diff
@@ -86,6 +86,21 @@ public static function shouldPersist(
...
+    {
+        foreach($this->cookies as $cookie) {
+            if(strcasecmp($cookie->getName(), $name) === 0) {
```
Nyholm: You could have a `SetCookie` with name = null and that would fail on HHVM. ([3v4l](https://3v4l.org/oLT0P)). That is an edge case but I believe we should cover it.

sagikazarmark: @dylan-mcdonald can you address that?

sagikazarmark:
Ping @dylan-mcdonald

src/Cookie/CookieJar.php
dylan-mcdonald: @sagikazarmark: Thanks for repinging me. I was out of the country. I can get to it later tonight. 😺

src/Cookie/CookieJar.php
```diff
@@ -89,10 +89,10 @@ public static function shouldPersist(
     /**
      * Finds and returns the cookie based on the name
      *
-     * @param string $name cookie name to search for
+     * @param string $name cookie name to search for (null by default)
      * @return SetCookie|null cookie that was found or null if not found
      */
-    public function getCookieByName($name)
+    public function getCookieByName($name = null)
```
sagikazarmark: Not sure if this is what @Nyholm pointed out. The problem is that when name is null, strcasecmp will fail on HHVM.

sagikazarmark:
Ping @dylan-mcdonald

sagikazarmark:
Ping @dylan-mcdonald

dylan-mcdonald:
OK. I think I misunderstood the last comments. I have updated the method to filter out null names.

sagikazarmark:
LGTM. @Nyholm ?

sagikazarmark:
It seems that passing null as the second parameter also fails under HHVM. Should we add a check if the name parameter is string, or just let people fail?

tests/Cookie/CookieJarTest.php
```diff
@@ -37,6 +37,19 @@ public function testCreatesFromArray()
...
+    {
+        $cookies = $this->getTestCookies();
+        foreach ($this->getTestCookies() as $cookie) {
+            $this->jar->setCookie($cookie);
+        }
+
+        $testCookie = $cookies[0];
+        $this->assertEquals($testCookie, $this->jar->getCookieByName($testCookie->getName()));
+        $this->assertNull($this->jar->getCookieByName("doesnotexist"));
+        $this->assertNull($this->jar->getCookieByName(null));
```
Nyholm: Is this a valid usage of this function? Even if the cookie name is `null` we would never return it with the current implementation. 

See my suggestion above.

src/Cookie/CookieJar.php
```diff
@@ -86,6 +86,21 @@ public static function shouldPersist(
         return false;
     }
 
+    /**
+     * Finds and returns the cookie based on the name
+     *
+     * @param string $name cookie name to search for (null by default)
```
Nyholm: Remove "(null by default)". We should not allow `$name` to be null.

tests/Cookie/CookieJarTest.php
Nyholm: Change to `$this->assertNull($this->jar->getCookieByName(''));`

dylan-mcdonald: Done. I pushed changes that just return null when `$name` is null.
Or do you prefer to throw an exception when this happens?

deepdivedylan:
I am returning to this thread to say I renamed myself from @dylan-mcdonald to @deepdivedylan. Also, what's the status on this PR? 😺

sagikazarmark:
LGTM. @Nyholm ?

deepdivedylan:
ping @Nyholm - how does this look? 😺

PR #1321, opened 2015-11-25
PSR-2

hannesvdvreken:


mtdowling:
I haven't checked recently, but I don't think PSR-2 has any guidance specifically around empty interfaces.

hannesvdvreken:
CS-fixer level psr-2 puts them like this. Opening brackets always go on the next line.

hannesvdvreken:
But it's your call, just proposing it here.

mtdowling:
Instead of a lot of one-off code formatting PRs, I'd be more open to discussing a more general PSR-2 update. Closing this.

hannesvdvreken:
Yeah, good idea. What would you be open to? A `.php_cs` file or a `scrutinizer.yml` file for checking and a big PR to fix all small inconsistencies? That might be hard to merge because of open PRs that will give merge conflicts then.

PR #1339, opened 2015-12-18

evictor:

```
Content-Type: text/plain
Set-Cookie: foo=bar
SET-COOKIE: baz=1
```

```
[
]
```


evictor:
Give me a moment to fix.

evictor:
Tests are passing locally so...

evictor:
OK I can reproduce failures.

evictor:
OK this should be good to go.

evictor:
Nvm actually I'm closing this for now. This is not so easy of a change since there are tests and other things depending on resultant case of headers.

evictor:
OK all tests passing and this is rebased on 5.3.

PR #1340, opened 2015-12-18

evictor:

```
Content-Type: text/plain
Set-Cookie: foo=bar
SET-COOKIE: baz=1
```

```
[
]
```


evictor:
Trying to select 5.3 as target but github going awry...

PR #1341, opened 2015-12-18

evictor:

```
Content-Type: text/plain
Set-Cookie: foo=bar
SET-COOKIE: baz=1
```

```
[
]
```

### DIFF
--- a/src/Message/AbstractMessage.php
+++ b/src/Message/AbstractMessage.php
@@ -106,7 +106,18 @@ abstract class AbstractMessage implements MessageInterface
     public function setHeaders(array $headers)
     {
         $this->headers = $this->headerNames = [];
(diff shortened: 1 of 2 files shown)


mtdowling:
Checking out the latest Guzzle 5.3, I'm not seeing the same issue:

```php
<?php

require 'vendor/autoload.php';

use GuzzleHttp\Message\Response;

$r = new Response(200, [
    'Set-Cookie' => ['bar', 'baz'],
    'SET-COOKIE' => ['bam', 'qux']
]);

echo $r;
```

Outputs:

```
HTTP/1.1 200 OK
SET-COOKIE: bar, baz, bam, qux

```

mtdowling:
Closing. Feel free to reopen if you can provide more details.

PR #1344, opened 2015-12-21
Fix example by adding backslash

JHGitty:


GrahamCampbell:
:-1:

GrahamCampbell:
Backslack not needed because the example is without a namespace.

GrahamCampbell:
It's incorrect to add a bashslash when not inside a namespace.

JHGitty:
But later in the same example there is a backslash used:
```php
$request = new \GuzzleHttp\Psr7\Request('GET', 'http://httpbin.org');
```

GrahamCampbell:
Then I think we should remove that backslash.

mtdowling:
Yeah I don't think this is necessary

PR #1350, opened 2015-12-28
Fix for PHP Fatal error in MockHandler.php

xopoc14:
The error appears when 'callable' is send as a string to 'queue' argument of MockHandler.
For example:
```php
$mock_handler = new MockHandler([
    '\MyNamespace\MyTest::mockResponse',
]);
```
We get:
PHP Fatal error:  Call to undefined function \MyNamespace\MyTest::mockResponse() in guzzlehttp/guzzle/src/Handler/MockHandler.php on line 77


ericduran:
Sweet. :+1:

PR #1352, opened 2015-12-30
Report JSON encoding errors

thewunder:
I'll do a separate PR for master if you like...

### DIFF
--- a/src/Message/MessageFactory.php
+++ b/src/Message/MessageFactory.php
@@ -343,7 +343,7 @@ class MessageFactory implements MessageFactoryInterface
             case 'json':
 
                 $json = json_encode($value);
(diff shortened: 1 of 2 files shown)


thewunder:
Hmm unaware that handling json errors in php 5.4 would be different.  If there is interest I can get it working in 5.4

src/Message/MessageFactory.php
```diff
@@ -342,7 +342,11 @@ protected function applyOptions(
 
             case 'json':
 
-                $request->setBody(Stream::factory(json_encode($value)));
+                $json = json_encode($value);
+                if(!$json) {
```
SofHad: Space after if

PR #1362, opened 2016-01-17
Support preserving host in headers for send function

lexakozakov:


mtdowling:
Guzzle should do this by default without any opt-in or ability to turn it off. If a user specifies a "Host" header, then Guzzle should always use that Host header.

PR #1367, opened 2016-01-19
Fix #1366 error when posting with form_params

### LINKED ISSUE #1366: HHVM in PHP7 strict mode fails to send any POST variables
When attempting to use form_params on HHVM in PHP7 strict mode Guzzle will send no post variables.

```
Warning: http_build_query() expects parameter 2 to be string, null given in /web/mizmoz-portal/vendor/guzzlehttp/guzzle/src/Client.php on line 294
```

ianchadwick:
HHVM return null when trying to build the query due to null being passed rather than an empty string.

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -291,7 +291,7 @@ class Client implements ClientInterface
                     . 'x-www-form-urlencoded requests, and the multipart '
                     . 'option to send multipart/form-data requests.');
             }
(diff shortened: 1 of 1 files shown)


jeskew:
Thanks!

moniruzzaman97m:
HHVM return null when trying to build the query due to null being passed rather than an empty string.

GrahamCampbell:
We don't support HHVM, sorry.

PR #1368, opened 2016-01-20
Fixed describe_type tests on PHP 7.0

Naktibalda:

PR #1369, opened 2016-01-21
Fixed describe_type tests on PHP 7.0

Naktibalda:


mtdowling:
Can you give a little context on what this does and why it's necessary?

GrahamCampbell:
@mtdowling Before: https://travis-ci.org/guzzle/guzzle/jobs/103712975. after https://travis-ci.org/guzzle/guzzle/jobs/103813457.

Naktibalda:
It was the only missing thing to get your tests pass on PHP 7.0

.travis.yml
```diff
@@ -12,13 +12,13 @@ before_script:
   - composer install --no-interaction --prefer-source --dev
   - ~/.nvm/nvm.sh install v0.6.14
   - ~/.nvm/nvm.sh run v0.6.14
+  - '[ "$TRAVIS_PHP_VERSION" != "7.0" ] || echo "xdebug.overload_var_dump = 1" >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini'
```
mtdowling: The logic here is very hard to follow. Can you update this to be the more straightforward if/then rather than if-not/else?

Shouldn't it also just set it to `0` when it's PHP 7?

```
[ "$TRAVIS_PHP_VERSION" == "7.0" ] && echo "xdebug.overload_var_dump = 0" >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini
```

/cc @jeskew

Naktibalda: I tried that in a previous PR: https://github.com/guzzle/guzzle/pull/1368
Build failed on all other versions because that line returns result code 1 in all other versions:
https://travis-ci.org/guzzle/guzzle/builds/103807788

|| ensures that 0 is returned in all versions.

Naktibalda: "Before" build failed because overloaded var_dump printed filename and line number.
Setting it to 1 was enough to fix it.

PR #1372, opened 2016-01-22
Changed  type hinting from array to ArrayAccess.

codbox:
Because of the type hinting am not able to pass an ArrayObject as a container. 

Passing an array is not possible for me, as i will have to pass it around by reference. 
and laravel IOC while resolving wont pass the paraments  by reference. 

the solution i found is to make the container an object ie: ArrayObject, as it will be passed by reference. 

But the type hinting was preventing me from doing so.

### DIFF
--- a/src/Middleware.php
+++ b/src/Middleware.php
@@ -76,7 +76,7 @@ final class Middleware
      *
      * @return callable Returns a function that accepts the next handler.
      */
-    public static function history(array &$container)
+    public static function history(\ArrayAccess &$container)
     {
         return function (callable $handler) use (&$container) {
             return function ($request, array $options) use ($handler, &$container) {


jeremeamia:
As you can see by the test failures in Travis, your proposed change is a breaking change. By changing it to \ArrayAccess, arrays will no longer work.

I you are having trouble with how the Laravel IoC container works with this object, then I suggest wrapping it with a custom object.

```php
<?php

namespace Your\Project;

use GuzzleHttp\Middleware;

class HistoryMiddleware
{
    private $nextHandler;



    public function __invoke(RequestInterface $request, array $options)
    {

    }
}

jeremeamia:
@codbox How about https://github.com/guzzle/guzzle/pull/1373 as an alternative. This should meet your needs and still be backward-compatible.

mtdowling:
Closing in favor of #1373

codbox:
Thanks, that works great.

PR #1373, opened 2016-01-22
More flexible history containers

jeremeamia:
Alternative to #1372.

Allows an array or array-accessible object to be used as the history container for the history middleware.


mtdowling:
Nice!

PR #1377, opened 2016-01-25
Use a LazyOpenStream instead of try_fopen r+

### LINKED ISSUE #1375: sink does not create new file
if i pass a file path to sink option in a request, the file is not created because it tries to fopen the file with r+

jeskew:
It looks like the stream handler [will immediately open a stream with an 'r+' flag](https://github.com/guzzle/guzzle/blob/5c12b74e9405c50fb0fd6ab831dd6e9df5560b1f/src/Handler/StreamHandler.php#L141), while the curl handler [will create a `LazyOpenStream` with a 'w+' flag](https://github.com/guzzle/guzzle/blob/5c12b74e9405c50fb0fd6ab831dd6e9df5560b1f/src/Handler/CurlFactory.php#L359). This will resolve #1375.

/cc @mtdowling @jeremeamia

### DIFF
--- a/src/Handler/StreamHandler.php
+++ b/src/Handler/StreamHandler.php
@@ -138,7 +138,7 @@ class StreamHandler
             : fopen('php://temp', 'r+');
 
         return is_string($sink)
-            ? new Psr7\Stream(Psr7\try_fopen($sink, 'r+'))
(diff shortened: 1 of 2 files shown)


jeremeamia:
That makes sense.

mtdowling:
:shipit:

PR #1381, opened 2016-01-27
fix NTLM auth support

Sebascii:


Sebascii:
added CURLAUTH_NTLM  http auth option

$client = new Client([
        'base_uri' => $url,
        'auth'    => array($user, $password 'ntlm')
    ]);

jeskew:
Is there any way to support NTLM that will work with the stream handler?

mtdowling:
Please update this PR with tests, documentation, and this would require throwing an exception in the StreamHandler because it does not support ntlm.

sagikazarmark:
@Sebascii  Should this be closed in favor of #1569?

sagikazarmark:
Thanks for your contribution, but I merged #1569 which implemented the same feature.

PR #1382, opened 2016-01-27
Add a constant holding base_uri config value to RequestOptions

vhraban:
It seems like all possible config options are defined in RequestOptions class. I have been using those constant when defining configs for the client in my code, but felt that the fact base_uri is not defined as constant is inconsistent. 
This PR would help to have less hard coded strings in the code that uses Guzzle


mtdowling:
`base_uri` is not a request option, it's a client constructor option.

PR #1388, opened 2016-01-30
I would expect to capture the same of a modified version of the request

hannesvdvreken:
Hi,

Q: why is it that you can't modify a request using the tap function to create a middleware?

--
please don't merge this, it breaks existing usages of the `tap` function.


jeremeamia:
You could use the [`mapRequest` middleware](https://github.com/guzzle/guzzle/blob/master/src/Middleware.php#L221-L236) instead.

hannesvdvreken:
Alright, thanks!

PR #1393, opened 2016-02-08
don't use eventName as parameter and variable

xabbuh:
Using `$eventName` for both as function parameter and in the `use` statement will fail on PHP 7.1 with a message like "Cannot use lexical variable $eventName as a parameter name" (see symfony/symfony-installer#237).

There also is a test in the test suite for PHP 7.1 ensuring that such a fatal error is triggered: https://github.com/php/php-src/blob/master/Zend/tests/closure_use_parameter_name.phpt

### DIFF
--- a/src/Event/Emitter.php
+++ b/src/Event/Emitter.php
@@ -43,8 +43,7 @@ class Emitter implements EmitterInterface
     public function once($eventName, callable $listener, $priority = 0)
     {
         $onceListener = function (
-            EventInterface $event,
-            $eventName
+            EventInterface $event
         ) use (&$onceListener, $eventName, $listener, $priority) {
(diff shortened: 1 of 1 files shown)


jeremeamia:
It seems like having the `$eventName` parameters in the original code was probably a mistake anyway.

cmeury:
This fix seems to be present in the 5.3.1 release only:

```bash
$ git tag --contains fcae91ff31de41e312fe113ec3acbcda31b2622e
5.3.1
```

Shouldn't it be also merged to `6.x`?

sagikazarmark:
Guzzle 6 is middleware based, there are no events anymore.

cmeury:
I see, thanks!

Issue #1405, opened 2016-02-17
Possibility to turn of quoting of cookies


dvaeversted:
I am currently using Guzzle for interfacing with a webservice, where i need a session cookie that contains a "=" sign.
Unfortunately this webservice does not support quoted cookie values, which is done in the `CookieJar::getCookieValue()` method.

Possible solutions:
1. Don't escape because of a = sign
2. Allow overwriting the getCookieValue() method, by modifying the call in `CookieHeader::withCookieHeader()` to utilize late static binding.

For now i had to completely re-implement the CookieJarInterface, for my use case to work.

mtdowling:
Hm... I'm open to changing the behavior here.

I wonder if we should be quoting at all. What does the cookie spec say? I wonder what other HTTP implementations do... For example, Firefox, Chrome, other PHP packages, etc.

dvaeversted:
While using the webservice through either Firefox or Chrome, neither of them escapes the cookie data.

The RFC does allow escaping, but i think the correct way would be to only escape if the server also sends it escaped.

mtdowling:
Thanks. I'm be open to changing it to match other prominent implementations.

PR #1406, opened 2016-02-17
Do not attempt to escape cookie values.

dvaeversted:
Trust that the cookie value sent by the server is escaped accordingly,
and do not attempt to escape the cookie by adding quotes around the value.

This is a pull request for: #1405 

I can see others have run into the same issue: http://tricksty.com/coding/remove-double-quotes-from-guzzle-cookies

### DIFF
--- a/src/Cookie/CookieJar.php
+++ b/src/Cookie/CookieJar.php
@@ -58,22 +58,10 @@ class CookieJar implements CookieJarInterface
     }
 
     /**
-     * Quote the cookie value if it is not already quoted and it contains
-     * problematic characters.
-     *
-     * @param string $value Value that may or may not need to be quoted
-     *
-     * @return string
+     * @deprecated
      */
     public static function getCookieValue($value)
     {
-        if (substr($value, 0, 1) !== '"' &&
-            substr($value, -1, 1) !== '"' &&
-            strpbrk($value, ';,=')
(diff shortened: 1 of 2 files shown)


src/Cookie/CookieJar.php
```diff
@@ -58,26 +58,6 @@ public static function fromArray(array $cookies, $domain)
...
 
     /**
-     * Quote the cookie value if it is not already quoted and it contains
-     * problematic characters.
-     *
-     * @param string $value Value that may or may not need to be quoted
-     *
-     * @return string
-     */
-    public static function getCookieValue($value)
```
mtdowling: Instead of removing this public method, let's mark it as deprecated, remove the rest of the docblock other than the `@deprecated` annotation, and have it just return the value as-is. That way anyone calling this directly wont break.

dvaeversted:
Updated commit according to comments.

mtdowling:
thanks!

carlosV2:
I've spend two days debugging my app, goutte and guzzle till I found this as the root of my problem :sob:

When is this going to be released? Thanks :grin:

avindra:
@carlosV2 : Released 20 hours ago

carlosV2:
@avindra thanks and thanks for reporting it :)

PR #1422, opened 2016-03-07
Rewind seekable request bodies before dispatching to cURL

### LINKED ISSUE #1335: Should CurlHandler rewind request body before consuming?
Here is a code describing a problem I got:
```php

$big_data = str_repeat('a', 1000000);

try {
    $body = stream_for($big_data);
    $signature = md5($body); // I have to read whole body to build signature...
    // $body->rewind();

    $client = new Client();

    $response = $client->post('http://httpbin.org/post', ["body" => $body]);

    echo json_decode($response->getBody())->data, PHP_EOL;
}
catch (Exception $e) {
    echo (string) $e->getMessage(), PHP_EOL;
}
```

This code results in curl timeout. Since I consumed body stream already and size of body more then 1000000, CurlHandler uses CURLOPT_READFUNCTION https://github.com/guzzle/guzzle/blob/2e5c595512de4885435eaff24b67066769b64b34/src/Handler/CurlFactory.php#L268 to read body. Since body is EOF currently, read returns empty string and curl just hangs until timeout reached.
I am ok with rewinding body myself, but in case when body length is less then 1000000, then stream will be casted to string which implicitly calls rewind on the stream making request magically work. 
StreamHandler also casting body to a string, so it works as well.

My guess here is that CurlHandler should rewind stream before consuming, since Content-length is calculated based on full stream length and in case we consumed chunk of the stream it would produce broken request.

jeskew:
This will resolve #1335.

/cc @mtdowling @jeremeamia

### DIFF
--- a/src/Handler/CurlFactory.php
+++ b/src/Handler/CurlFactory.php
@@ -265,6 +265,9 @@ class CurlFactory implements CurlFactoryInterface
                 $this->removeHeader('Content-Length', $conf);
             }
             $body = $request->getBody();
+            if ($body->isSeekable()) {
+                $body->rewind();
+            }
             $conf[CURLOPT_READFUNCTION] = function ($ch, $fd, $length) use ($body) {
                 return $body->read($length);
             };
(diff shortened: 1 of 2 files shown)


mtdowling:
Can you double check that all the AWS SDK for PHP tests are unaffected by this patch?

jeskew:
Integ and unit suites are still passing.

PR #1436, opened 2016-03-22
add curl sslversion option

qianfunian:


mtdowling:
We could only add `ssl_version` if it can be abstracted to work with other handlers (e.g., StreamHandler).  Otherwise, you can just add custom curl options to your request: http://docs.guzzlephp.org/en/latest/faq.html#how-can-i-add-custom-curl-options

qianfunian:
@mtdowling thanks

PR #1440, opened 2016-03-29
syntax fix

natostanco:


src/Handler/StreamHandler.php
```diff
@@ -284,7 +284,7 @@ function () use ($context, $params) {
 
         return $this->createResource(
             function () use ($request, &$http_response_header, $context) {
-                $resource = fopen($request->getUri(), 'r', null, $context);
+                $resource = fopen($request->getUri(), 'r', '', $context);
```
mtdowling: If this is to change, it should be changed to `false` not `''`.

natostanco: yes I used bool at the beginning, works the same, but an empty string preserving the type on the surface kind of felt more approriate

src/Client.php
```diff
@@ -347,7 +347,7 @@ private function applyOptions(RequestInterface $request, array &$options)
         if (isset($options['query'])) {
             $value = $options['query'];
             if (is_array($value)) {
-                $value = http_build_query($value, null, '&', PHP_QUERY_RFC3986);
+                $value = http_build_query($value, false, '&', PHP_QUERY_RFC3986);
```
mtdowling: If we're going to change this, then I think this should still be an empty string. See http://php.net/manual/en/function.http-build-query.php

natostanco: reverted

src/Handler/StreamHandler.php
mtdowling: The signature for fopen is: `resource fopen ( string $filename , string $mode [, bool $use_include_path = false [, resource $context ]] )`.

Using `''` for `bool $use_include_path = false` doesn't make sense. `false` would be a better change.

PR #1442, opened 2016-04-01
Updating documentation for async requests

kenske:
I took me many hours to realize I could use the settle function instead of unwrap if I wanted to get the responses regardless of their status.


jeremeamia:
Sounds like a good thing to add. Thanks.

PR #1443, opened 2016-04-01
php command

bil9000:
minor command missing, easy for some, however for newbie… fatality!


bil9000:
the composer bash line was missing thte php command, just a minor tweak. :)

bil9000:
fiing read me

PR #1449, opened 2016-04-12
Fixes #1448 Update Request-constructor to satisfy docblock.

### LINKED ISSUE #1448: Guzzle 5.3: Misleading docblock for body in Request-constructor
The docblock for `Request::__construc()` currently says that the 4th argument `$body` is mixed (see: https://github.com/guzzle/guzzle/blob/5.3/src/Message/Request.php#L31). As far as I can tell it can either be a StreamInterface or null. It would be helpful, e.g. for IDEs like PHPStorm if this is more accurate.
I would like to propose to either change the docblock to:

```
    /**
     * [...]
     * @param StreamInterface|null $body    Body to send with the request
     * [...]
     */
```

or change the constructor itself to call `Stream::factory($body) before setting the value`:

```
        if ($body !== null) {
            if (!($body instanceof StreamInterface)) {
                $body = Stream::factory($body);
            }
            $this->setBody($body);
        }
```

I can open a PR, but I wanted to make sure this change is actually appreciated and which of these is preferred.

dbrumann:
see ticket #1448

### DIFF
--- a/src/Message/Request.php
+++ b/src/Message/Request.php
@@ -3,6 +3,8 @@ namespace GuzzleHttp\Message;
 
 use GuzzleHttp\Collection;
 use GuzzleHttp\Event\HasEmitterTrait;
+use GuzzleHttp\Stream\Stream;
+use GuzzleHttp\Stream\StreamInterface;
 use GuzzleHttp\Subscriber\Prepare;
 use GuzzleHttp\Url;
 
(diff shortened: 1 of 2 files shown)


mtdowling:
If `Stream::factory()` is passed an instance of StreamInterface, it will return it as is. Can you remove the updated conditional?

See https://github.com/guzzle/streams/blob/master/src/Stream.php#L62

mtdowling:
Actually, this PR is pointless if you remove that addition... Sorry, this is not an issue.

PR #1453, opened 2016-04-19
defaultCredentials attribute added.

sangio90:
If credentials are required for many requests of the same Client instance, it is helpful not to specify them in each request.

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -113,7 +113,22 @@ class Client implements ClientInterface
 
     public function requestAsync($method, $uri = null, array $options = [])
     {
+        // Remove the option so that they are not doubly-applied.
         $options = $this->prepareDefaults($options);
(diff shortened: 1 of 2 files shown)


sangio90:
don't know why the test is failing. It was already failing when I first forked the project.

sangio90:
Added method createNewRequest to split the logic of the request creation from the async request.

mtdowling:
This change is not necessary because you can already set the `auth` request option when constructing the client, which will apply that option on every request.

sangio90:
Should I create a new pull request for the createNewRequest method?

sangio90:
BTW you can't set credentials after you created the client, credentials might be available some time after instantiating the client.

mtdowling:
In that case, you should use a middleware.

Having a RequestFactory that could be injected into a client would be great if you want to implement it.

PR #1459, opened 2016-04-25
Fix strange auth option handling

Tobion:


src/Client.php
```diff
@@ -325,13 +325,10 @@ private function applyOptions(RequestInterface $request, array &$options)
             unset($options['body']);
         }
 
-        if (!empty($options['auth'])) {
+        if (!empty($options['auth']) && is_array($options['auth'])) {
             $value = $options['auth'];
-            $type = is_array($value)
-                ? (isset($value[2]) ? strtolower($value[2]) : 'basic')
-                : $value;
-            $config['auth'] = $value;
```
Tobion: this var does not seem to be used at all

src/Client.php
```diff
@@ -325,13 +325,10 @@ private function applyOptions(RequestInterface $request, array &$options)
...
-            switch (strtolower($type)) {
```
Tobion: this was doing a `strtolower` a second time

src/Client.php
```diff
@@ -325,13 +325,10 @@ private function applyOptions(RequestInterface $request, array &$options)
             unset($options['body']);
         }
 
-        if (!empty($options['auth'])) {
+        if (!empty($options['auth']) && is_array($options['auth'])) {
             $value = $options['auth'];
-            $type = is_array($value)
```
Tobion: the `is_array` check here does not make sense as it can still fail a few line later at `base64_encode("$value[0]:$value[1]");` if it is not an array. So instead lets just skip the whole auth handling if is not an array.

Tobion:
Test failure seems unrelated

Issue #1419, opened 2016-03-05
What replaces client->setDefaultOption?


project707:
In our application we create a new `GuzzleHttp\Client` at a high class/service level and need to set options on that multiple places in child classes that inherit from that differently depending on what methods get called, but I can't figure out how that is supposed to be done to an instantiated client on the fly now (Guzzle 6.1.1).

I'm sure I am missing something really obvious in the docs/code since this is such a common use-case. The only method I see that seems like it should do that is `configureDefaults` which is private, and I know that  `setDefaultOption` went away.

mtdowling:
Clients are immutable in Guzzle 6, which means that you cannot change the defaults used by a client after it's created.

Maybe instead of trying to modify a client after it's created, you could utilize factories that create clients using the default values necessary for each different type required in your application.

project707:
From the perspective of someone needing to reuse something like dynamically generated proxy settings (and others specific to a logged in "session") for a set of requests, making clients immutable sounds like a fundamental flaw.

I get that psr7 means that request settings are immutable, but why does that carry over to client instances? Just curious.

mtdowling:
> From the perspective of someone needing to reuse something like dynamically generated proxy settings (and others specific to a logged in "session") for a set of requests, making clients immutable sounds like a fundamental flaw.

I don't see how you can draw that conclusion from your use case. Instead of dynamically modifying a client, just dynamically create clients.

Immutability is not necessary at the client level. It is not uncommon for middleware to be created that derives values based on state at a point in time. If that state changes, then the derived value must be updated accordingly, which means you either have to increase coupling between state changes and derived values or you cannot use cached/derived values in client collaborators.

Adding support for a MutableClient or something similar is something I'd consider (e.g. a separate interface and concrete class), but I do not want to change the ClientInterface or default Client implementation.

project707:
I see where you're coming from, but having thought we could rely on the Guzzle client to hold the "state", or maybe rather metadata, of a client for us (as it could up to this point) and having built a large modular application using it that way, you can probably understand a little frustration at that having changed so decidedly. The way things were, it required less code on our side and less of a necessity to handle the state of that metadata ourselves (i.e. no need to look at a service that holds some list of attributes given to the original client, but then for all I know, doing things this way may have also reduced the size or complexity (more likely) of the Guzzle codebase, which of course also holds great value.

[...]

PR #1462, opened 2016-04-26
Updated quickstart documentation

azamtav:
Added a line stating that clients are immutable


azamtav:
I couldn't find any documentation in the documents regarding clients being immutable. I found this closed issued which shed some light on it: https://github.com/guzzle/guzzle/issues/1419

Issue #1463, opened 2016-04-29
Bug in clients magic __call function


DaanBiesterbos:
The name of the PHP method is used as if it is a HTTP method.   The first argument is the name of the invoked function. The second argument are the actual parameters.  

```
public function __call($method, $args)
    {
        if (count($args) < 1) {
            throw new \InvalidArgumentException('Magic request methods require a URI and optional options array');
        }

        $uri = $args[0];
        $opts = isset($args[1]) ? $args[1] : [];

        return substr($method, -5) === 'Async'
            ? $this->requestAsync(substr($method, 0, -5), $uri, $opts)
            : $this->request($method, $uri, $opts);
    }
```

This is how it should be:
```
$method = $args[0];
$uri = $args[1];
$opts = $args[2];

```

jeskew:
`$method` will be the method called on a client. If, for example, you call `$client->get('/', [])`, then `__call` will be invoked with `get` bound to `$method` and `['/', []]` bound to `$args`. Please see [the PHP manual](http://php.net/manual/en/language.oop5.overloading.php#object.call) for a more in-depth explanation.

DaanBiesterbos:
Yes I know what __call does. But I do now understand what the intented purpose of the __call method is. If you upgrade from 5 to 6 the createRequest method os gone and will invoke the __call method. Perhaps it would be better to thrown an exception if the invoked method is not a http method. :-)

PR #1464, opened 2016-04-29
bugfix/ticket-1463

DaanBiesterbos:
Fix for issue described in ticket #1463

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -73,14 +73,17 @@ class Client implements ClientInterface
         $this->configureDefaults($config);
     }
 
-    public function __call($method, $args)
+    public function __call($phpMethod, $args)
     {
-        if (count($args) < 1) {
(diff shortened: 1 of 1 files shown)


src/Client.php
```diff
@@ -73,14 +73,17 @@ public function __construct(array $config = [])
         $this->configureDefaults($config);
     }
 
-    public function __call($method, $args)
+    public function __call($phpMethod, $args)
```
GrahamCampbell: this rename is unneeded, and should be reverted - the original name was better anywant

jeskew:
This doesn't look backwards compatible. What's the motivation for this change?

DaanBiesterbos:
Assumed bug is not an issue. Throwing an exception when method name is not a http method would be more appropriate.

PR #1468, opened 2016-05-05
[Travis-CI] Prefer dist

JHGitty:
Reason: This can speed up installs substantially on build servers and other use cases where you typically do not run updates of the vendors. It is also a way to circumvent problems with git if you do not have a proper setup.

Source: https://getcomposer.org/doc/03-cli.md#install

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ php:
 before_script:
   - curl --version
   - composer self-update
-  - composer install --no-interaction --prefer-source --dev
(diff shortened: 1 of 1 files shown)


jeskew:
Specifying `--prefer-source` prevents Travis from getting rate-limited by the GitHub API. `--prefer-source` pulls dependencies with `git clone`; `--prefer-dist` uses the GitHub API.

mtdowling:
Thanks @jeskew. Closing as I don't want our tests to be throttled.

PR #1474, opened 2016-05-11
Catch PHP7 \Throwable

fprochazka:
There are several "catch all" traps in the code, that catch `\Exception` and they should be catching `\Throwable`.
While it's not possible to simply replace it, since that would break compatibility with PHP 5.6, it still can be fixed pretty easily.

Little worse are the typehints, as I'm not sure how much you wanna typecheck that the object is in fact a exception.

### DIFF
--- a/src/Exception/ConnectException.php
+++ b/src/Exception/ConnectException.php
@@ -10,10 +10,17 @@ use Psr\Http\Message\RequestInterface;
  */
 class ConnectException extends RequestException
 {
+
+    /**
+     * @param string $message
+     * @param \Psr\Http\Message\RequestInterface $request
+     * @param \Exception|\Throwable $previous
+     * @param mixed[] $handlerContext
+     */
     public function __construct(
         $message,
         RequestInterface $request,
-        \Exception $previous = null,
+        $previous = null,
         array $handlerContext = []
     ) {
         parent::__construct($message, $request, null, $previous, $handlerContext);
--- a/src/Exception/RequestException.php
+++ b/src/Exception/RequestException.php
@@ -19,11 +19,18 @@ class RequestException extends TransferException
     /** @var array */
     private $handlerContext;
 
+    /**
+     * @param string $message
+     * @param \Psr\Http\Message\RequestInterface $request
+     * @param \Psr\Http\Message\ResponseInterface $response
+     * @param \Exception|\Throwable $previous
+     * @param mixed[] $handlerContext
+     */
     public function __construct(
         $message,
         RequestInterface $request,
         ResponseInterface $response = null,
-        \Exception $previous = null,
+        $previous = null,
         array $handlerContext = []
     ) {
         // Set the code of the exception if the response is set and not future.
--- a/src/Client.php
+++ b/src/Client.php
@@ -268,6 +268,8 @@ class Client implements ClientInterface
             return Promise\promise_for($handler($request, $options));
         } catch (\Exception $e) {
             return Promise\rejection_for($e);
+        } catch (\Throwable $e) {
+            return Promise\rejection_for($e);
         }
     }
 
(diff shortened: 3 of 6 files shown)


GrahamCampbell:
:-1: On this actually.

fprochazka:
@GrahamCampbell can you explain why?

GrahamCampbell:
Throwable would indicate that the actual application is broken. Surely we only want to catch Exceptions that occur because of a bad http response.

fprochazka:
That means you maybe don't wanna wrap such exceptions in `RequestException`, but there still are some places where you wanna know things broke so you don't get that object into an invalid state.

fprochazka:
Just to be clear - I didn't study the lib deeply enough to really understand where it makes sense so I added it everywhere and was hoping for a discussion about this.

mtdowling:
This is a good change, but I don't like all the duplication that has to be introduced. Is there a better way that other projects have been handling this?

GrahamCampbell:
Most projects just do the duplication. A common refactor I've seen has been to just call a function specifically for handling the catch.

So, instead of:

```php
try {
    foo();
} catch (Exception $e) {
    bar();
    baz();
    foo();
} catch (Throwable $e) {
    bar();
    baz();
    foo();
}
```

```php
try {
    foo();
} catch (Exception $e) {
    handle();
} catch (Throwable $e) {
    handle();
}

function handle()
{
    bar();
    baz();
    foo();
}
```

This extends to the OOP version too. I'm just too lazy to type out a class dfn and $this a few times.

src/Exception/RequestException.php
```diff
@@ -19,11 +19,18 @@ class RequestException extends TransferException
...
+     * @param \Psr\Http\Message\RequestInterface $request
+     * @param \Psr\Http\Message\ResponseInterface $response
+     * @param \Exception|\Throwable $previous
+     * @param mixed[] $handlerContext
+     */
     public function __construct(
         $message,
         RequestInterface $request,
         ResponseInterface $response = null,
-        \Exception $previous = null,
```
fprochazka: this should probably be reverted

src/Exception/ConnectException.php
```diff
@@ -10,10 +10,17 @@
...
+     * @param string $message
+     * @param \Psr\Http\Message\RequestInterface $request
+     * @param \Exception|\Throwable $previous
+     * @param mixed[] $handlerContext
+     */
     public function __construct(
         $message,
         RequestInterface $request,
-        \Exception $previous = null,
+        $previous = null,
```
fprochazka: this should probably be reverted

src/Exception/RequestException.php
```diff
@@ -19,11 +19,18 @@ class RequestException extends TransferException
...
+        $previous = null,
```
fprochazka: this should probably be reverted

fprochazka:
Do any of those places make sense to change? Or should I just close this and consider this solved?

src/Exception/ConnectException.php
mtdowling: I could be wrong, but I think it has to be a subclass of Exception here: http://php.net/manual/en/language.exceptions.extending.php. I don't think you can can create an exception with a previous exception that is Throwable and not an instance of Exception.

GrahamCampbell: @mtdowling An exception can have a previous Throwable, that;s totally fine.

fprochazka:
I've went throught the code again by myself and I came to conclusion, that it doesn't make sense to catch throwable in any of theese places.

fprochazka:
Anyway, now I have a good feeling that this is solved :)

Neirda24:
@fprochazka @GrahamCampbell : I don't understand why you would refuse this feature ? I have the following example: I call an API wich return `null` in the body (which I know is not fine but still json valid) and it throws an Error (I don't know which one yet). But Guzzle is trying to log it in the `MessageFormatter::format`. Because it is the wrong type it throws an other error which "hide" the previous one...

PR #1479, opened 2016-05-20
Added setConfigProperty to dynamically set configurations

sangio90:
this method might be used to dynamically add configurations or to swap/change headers between groups of multiple requests and avoid adding configs to each request.


mtdowling:
This allows the client to be mutable, which contradicts the design decision that was made to keep the client immutable. Perhaps we could add a mutable client implementation, but I don't want to add a single mutable method to an otherwise immutable client.

PR #1480, opened 2016-05-22
Remove URI default value

Tobion:
The URI default value makes no sense for many reasons:

1. `$client->request('GET')` makes no sense. Where does it get it from?
2. the shortcut methods actually require the URI `$client->get($uri)` (otherwise throw an exception) so this wasn't consistent at all.
3. #1374 only changed `ClientInterface::request` but not `ClientInterface::requestAsync` which is inconsistent
3. Request class also requires the URI: https://github.com/guzzle/psr7/blob/master/src/Request.php#L36

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -104,7 +104,7 @@ class Client implements ClientInterface
         return $this->sendAsync($request, $options)->wait();
     }
 
-    public function requestAsync($method, $uri, array $options = [])
+    public function requestAsync($method, $uri = '', array $options = [])
     {
         $options = $this->prepareDefaults($options);
         // Remove request modifying parameter because it can be done up-front.
@@ -123,7 +123,7 @@ class Client implements ClientInterface
         return $this->transfer($request, $options);
     }
 
-    public function request($method, $uri, array $options = [])
+    public function request($method, $uri = '', array $options = [])
     {
         $options[RequestOptions::SYNCHRONOUS] = true;
         return $this->requestAsync($method, $uri, $options)->wait();
@@ -142,7 +142,7 @@ class Client implements ClientInterface
             return $uri instanceof UriInterface ? $uri : new Psr7\Uri($uri);
         }
 
-        return Psr7\Uri::resolve($config['base_uri'], $uri);
+        return Psr7\Uri::resolve(Psr7\uri_for($config['base_uri']), $uri);
     }
 
     /**
--- a/src/Client.php
+++ b/src/Client.php
@@ -142,7 +142,7 @@ class Client implements ClientInterface
             return $uri instanceof UriInterface ? $uri : new Psr7\Uri($uri);
         }
 
-        return Psr7\Uri::resolve(Psr7\uri_for($config['base_uri']), $uri);
+        return Psr7\Uri::resolve($config['base_uri'], $uri);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/Client.php
```diff
@@ -142,7 +142,7 @@ private function buildUri($uri, array $config)
             return $uri instanceof UriInterface ? $uri : new Psr7\Uri($uri);
         }
 
-        return Psr7\Uri::resolve(Psr7\uri_for($config['base_uri']), $uri);
+        return Psr7\Uri::resolve($config['base_uri'], $uri);
```
Tobion: base_uri is already a UriInterface (see constructor)

src/Client.php
```diff
@@ -353,13 +358,6 @@ private function applyOptions(RequestInterface $request, array &$options)
             unset($options['query']);
         }
 
-        if (isset($options['json'])) {
```
Tobion: I moved this up to be consistent with multipart and form_params in setting the body option.

mtdowling:
If the client has a base URI, then the URI can be optional, which is the intent of allowing null. The request class requires a URI because it has not context of a base URI like the client does.

This also seems like a breaking change. If someone has been relying on a base URI, then this would break them.

Tobion:
I totally agree that allowing null or more precisely `''` makes sense. But my point is it should not be optional. The empty string URI is exactly meant for referencing the "same-document" (base URI).
So one should call `$client->request('GET', '')` instead of `$client->request('GET')`.

Calling it without an URI just does not make sense semantically because it's not specified where you want to get it from. Compare it with `<a href="">Link</a>` vs `<a>Link</a>`. The later does not make sense which is why it does not actually create a link in browsers (as I just tried and read in the HTML spec).

If you do not feel confortable to make this change for BC reasons, then I can exlude this change. But #1374 should IMO definitely be reverted before a release as it technically a BC break as well and not consistently changed for `requestAsync`.

src/Client.php
Tobion: I need to revert that I just realize because the base_uri can be overwritten dynamically through the options in the `request` method. So it's not guaranteed that the base_uri is still a UriInterface instance. I'll add a test for this.

Tobion:
I readded the default URI for the client implementation for BC. I just changed it from null to empty string which is more correct and consistent with `[uri_for](https://github.com/guzzle/psr7/blob/master/src/functions.php#L54)` that does not accept null and shouldn't break BC.

mtdowling:
Thanks!

DrDelay:
I agree with the changes made here, they make sense, and I don't like to be that guy, but in a project I'm contributing to this turned out to be a BC-breaking change.

Whilst the default value you added in 566d90d7cd17ff4917f65665a00429a54c831b55 made it BC for most "normal people" just using the *Client*, in this project we have [something](https://github.com/NicklasWallgren/PokemonGoAPI-PHP/blob/4e1a5c3f3ac004f3f4494f346aeca4282b487001/src/Clients/Proxies/ClientProxy.php) implementing the *ClientInterface*, therefore with versions before/after this has been merged an `Declaration of … must be compatible with GuzzleHttp\ClientInterface::request(…)` error arised.

CC: https://github.com/NicklasWallgren/PokemonGoAPI-PHP/issues/143

Tobion:
Sorry I didn't think about that. But it's indeed an edge case.
Btw, I didn't see the point of you custom Client implementation. The stack created in https://github.com/NicklasWallgren/PokemonGoAPI-PHP/blob/4e1a5c3f3ac004f3f4494f346aeca4282b487001/src/Clients/Proxies/ClientProxy.php#L142 is not even used at all. So something is pretty strange there.

PR #1492, opened 2016-06-10
Add backwards compatibility for older curl

dgallinari:
Even knowing that we have set minimum versions for PHP and curl, this simple fix will allow a bunch of older versions of curl library also work with Guzzle. I faced this issue myself (on a shared hosting) and this change made everything work.

By the way, CURLOPT_TIMEOUT_MS was "Added in cURL 7.16.2. Available since PHP 5.2.3.". 
See: http://php.net/manual/en/function.curl-setopt.php

### DIFF
--- a/src/Handler/CurlFactory.php
+++ b/src/Handler/CurlFactory.php
@@ -372,7 +372,11 @@ class CurlFactory implements CurlFactoryInterface
         }
 
         if (isset($options['timeout'])) {
-            $conf[CURLOPT_TIMEOUT_MS] = $options['timeout'] * 1000;
+            if (defined('CURLOPT_TIMEOUT_MS')) {
+                $conf[CURLOPT_TIMEOUT_MS] = $options['timeout'] * 1000;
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
>  Available since PHP 5.2.3.

Guzzle needs php 5.5. :-1:

mtdowling:
I don't think this is the only change necessary to support older versions of curl.

dgallinari:
I made the test myself. I will double check which version of curl I had on the shared hosting, but this simple fix was enough.

Actually, there are other options which could have a fallback. I will check them and add to this PR.

Btw, thanks for the heads up @mtdowling :+1:

mtdowling:
I don't think we need to support older versions of cURL in Guzzle, and I'm not keen on the additional maintenance burden of dealing with cURL edge cases on such old versions of cURL. There would need to be a much more thorough investigation into supporting older versions of cURL, and at this point, I don't think that's best for the project.

PR #1493, opened 2016-06-16
fix: corrects docblock for json_encode

gsdevme:
`$value` should be `mixed` not `string` see https://secure.php.net/manual/en/function.json-encode.php

### DIFF
--- a/src/functions.php
+++ b/src/functions.php
@@ -309,7 +309,7 @@ function json_decode($json, $assoc = false, $depth = 512, $options = 0)
 /**
  * Wrapper for JSON encoding that throws when an error occurs.
  *
- * @param string $value   The value being encoded
+ * @param mixed $value   The value being encoded
  * @param int    $options JSON encode option bitmask
  * @param int    $depth   Set the maximum depth. Must be greater than zero.
  *


GrahamCampbell:
This isn't correct.

GrahamCampbell:
Guzzle's version only wants to take a string, and it's allowed to do that.

gsdevme:
So when constructing a `GuzzleHttp\Psr7\Request` a user of this package might want to just use the guzzle json_encode though? you are therefore saying users should include a third party package such as  https://packagist.org/packages/webmozart/json to do our encoding...

I would say this function has an incorrect description **Wrapper for JSON encoding that throws when an error occurs.** you are not wrapping the functionality you are changing it.

Perhaps also be marked with `@internal` if you are refusing to accept this PR as it would not impact Guzzle one bit

Luciam91:
@GrahamCampbell why would a wrapper for a PHP method take a different argument from the method it's trying to wrap?

GrahamCampbell:
> why would a wrapper for a PHP method take a different argument signature from the method it's trying to wrap?

I see no reason why it shouldn't be allowed to.

GrahamCampbell:
We're saying our one only expects strings. We're allowed to say that.

gsdevme:
@GrahamCampbell I literally don't understand the point you are making; please answer the following question.

1.Within Guzzle itself its passing an array see https://github.com/guzzle/guzzle/blob/master/src/Cookie/FileCookieJar.php#L58 so where do you get the idea that it needs to be a string?

GrahamCampbell:
> you are not a member of Guzzle yet you say "we"

Yes, we, the PHP community.

GrahamCampbell:
The PHP community can decide to define their own functions as they please.

GrahamCampbell:
Not sure what was unclear about that?

gsdevme:
**Guzzle itself its passing an array see https://github.com/guzzle/guzzle/blob/master/src/Cookie/FileCookieJar.php#L58 so where do you get the idea that it needs to be a string?** @GrahamCampbell

GrahamCampbell:
Oh, I'm so sorry! I thought this was the json_decode function.

Perhaps we should set it to be `array` then. What sort of things is guzzle passing into there other than arrays?

gsdevme:
I've pointed this out three times @GrahamCampbell now you have finally realised.. I'm really not sure why its taken this long to understand I feel like you wanted to jump on this PR to increase your Github activity as I can see you have this posted on your website and it seems to matter to you?`https://gist.github.com/paulmillr/2657075/` whereas I wanted to do this PR to help make Guzzle better... Im not here for glory 

My main reason for this PR was scrutinizer-ci was flagging up an issue as I was passing an array into the function that apparently only accepts a string given its docblock. Im not here for glory Im here to prevent scrutinizer flagging this issue. 

Im not sure if its been a bad day or you are on some ego trip but my one tip as a somewhat young developer (4 years in the business) is there is always a better programmer than you so careful what you jump in and say you can be left looking a little silly if you are not fully aware of the codebase. 

As for jumping on this PR thats great but the core team must decide what the correct logic is, you seemed to pretty aggressively argue what is correct for Guzzle yet you are not part of Guzzle. Discussion is good for open source and should never be shutdown without taking into account all aspects. I felt like you was just overlooking and repeating your words without even reading my counter arguments.

gsdevme:
screenshot-github com 2016-06-16 23-21-33

halfer:
I am not taking sides, but it is appropriate that @GrahamCampbell gets [the right to reply to criticism](https://www.reddit.com/r/PHP/comments/4ofo0q/the_hardest_pr_ever_guzzle/).

eXtreme:
@gsdevme you are right but attacking @GrahamCampbell with such cannons like thread on reddit is just very... unprofessional. I would totally blame github for that because I was confused as well when started reading this thread and then visiting "Files changed" tab here in this PR. 

githubblame

You can clearly see (and completely incorrect) a hint that it is about `json_decode` function as @GrahamCampbell [thought in this comment](https://github.com/guzzle/guzzle/pull/1493#issuecomment-226625166). But of course it is about `json_encode` function but you cannot see this in the diff!

gsdevme:
The PR title.. Also his first and now deleted reply was "This isn't correct" with no further explanation, along with not being part of the project in any official capacity and with code referenced to prove my point I feel I matched his professionalism! 

Literally no reply even gave the impression he was reading my comments. 

Sorry if offensive or harm was caused though, naturally I wanted none of this I'm just a guy fixing a bug I spotted while writing code I didn't go looking for this 'fight'

eXtreme:
@gsdevme and he admitted he had made a mistake but you take a screenshot, post on reddit, make a shitthread.. It does not look like `I didn't go looking for this 'fight'`...

> The PR title..

Yeah, the title is large, but large enough to mistake encode from decode.. Look at the diff here, it says decode, it is incorrect but it is github's fault.

DavertMik:
@eXtreme
> You can clearly see (and completely incorrect) a hint that it is about json_decode function as

From which time docblocks go beyond the function declaration? 0.o
GitHub pointed to a previous function in a file. It would be confusing if it pointed to next...
And it is clear from the first message that it is somehow related to `json_encode` function.

I think the level of discussion transformed this PR to topic on reddit. Even you see that this is something obvious you should try to clarify your point. "This isn't corrent" is not a good argument. If @GrahamCampbell described his thoughts in more extended way than it would be much easier to find a cause and move forward.

gsdevme:
Reddit post was before he posted, timestamps can confirm that.

gsdevme:
There is two people at fault.. Trying to erase things on the internet does not work though. The discussions took an hour. 

My Reddit was to highlight the way in which the discussion happened and how badly we can approach peer reviews of code. As I feel like "this isn't correct" is the perhaps the worse response I've ever seen and to happen 1-2 minutes after my PR... I won't back down about that point.

eXtreme:
@DavertMik 
> From which time docblocks go beyond the function declaration? 0.o

As I said, it is just incorrect and **misleading** - you open a PR, take a glance at the diff, see  decode, see "string being an incorrect input" and you make a rash comment... Just try to step into GC shoes.. He made a mistake and apologized.

@gsdevme sorry, didn't know about that

DavertMik:
Yes. @gsdevme created a PR, he explained it with code and brief comment.
I'm pretty sure that "this is not correct" point require at least some argumentation regarding code change. 
And yes, this kills motivation to create issues or PRs if we (*PHP community*) have quick answers for everything.

saada:
Why don't we all just take a hike, a vacation, a nap, a massage, some exercise, and then come back and take another look at this "end of the world" PR, shall we? 😂

micbenner:
Honestly, I have considered contributing to an open source project and have tried to follow them. But I get really frustrated with arrogant people coming along posting things like "this isn't correct" without giving 10 seconds to act like a human being or explain themselves.

These people often defend themselves because "they are busy and need to go through lots of things" In my opinion this isn't an acceptable excuse to be a prick and scares potential contributors away.

Open source communities should be a relatively professional environment. I don't know what it is, ego tripping, race for the first response, a competition of who can be the busiest, whatever - a few think they are better than everyone else and ruin it for all.

gsdevme:
I don't need to step into anyone shoes, I explained my reasons case and point and he blindly ignored that is clear. 

A mistake is a mistake, no one is god like here, if the first reply I had written was read it should of been clear. He didn't ask for further information he stood grounds with his "this isn't correct" and yes I got bitter about it after an hour of back and fourth and linking evidence... So Reddit happened.

GrahamCampbell:
@gsdevme Sorry I misread the PR. Nobody's perfect. Hopefully we can still be friends. ;)

Exadra37:
@gsdevme do not waste your time here, @GrahamCampbell is known by is ruddiness and that is a reason why i never contributed more to Laravel, despite i am use it professionally everyday.

GrahamCampbell:
@Exadra37 Looks like I'm not the only one who's rude.

PR #1509, opened 2016-06-29
Fix uri without scheme fallback

Tobion:
Fixes guzzle/psr7#102

When URI has no scheme use http by default.

curl usually also uses http when no scheme is given. see https://curl.haxx.se/libcurl/c/CURLOPT_URL.html
but curl requires network-path URIs without a scheme to not start with 2 starting slashes which is actually wrong (probably done for user-friendliness, so called suffix references, see https://tools.ietf.org/html/rfc3986#section-4.5). so the fixes in psr7 1.3.1 actually broke curl handler for URIs without scheme. we solve this problem by applying the http default scheme to the URI beforehand. this also ensures that the stream handler works the same way. fopen() requires a scheme and has no fallback logic. so there was an inconsistency.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,14 +13,14 @@
         }
     ],
     "require": {
-        "php": ">=5.5.0",
-        "guzzlehttp/psr7": "~1.1",
-        "guzzlehttp/promises": "~1.0"
+        "php": ">=5.5",
+        "guzzlehttp/psr7": "^1.3.1",
+        "guzzlehttp/promises": "^1.0"
     },
     "require-dev": {
         "ext-curl": "*",
-        "phpunit/phpunit": "~4.0",
-        "psr/log": "~1.0"
+        "phpunit/phpunit": "^4.0",
+        "psr/log": "^1.0"
     },
     "autoload": {
         "files": ["src/functions_include.php"],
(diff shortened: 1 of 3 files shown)


Tobion:
Will fix tests. Tests failing because of old PHP version parse_url bugs: https://3v4l.org/V2AGL#version=5.5.21

mtdowling:
`InvalidArgumentException: Unable to parse URI: //127.0.0.1:8585/test`: For this error, what about updating the PSR7 library to detect that the URI has no scheme, add a scheme of http so that it parses correctly, then remove the scheme when returning the URI object? Would that be possible?

Tobion:
Yes that could work. The problem is that this only handles one of the parse_url bugs, see https://bugs.php.net/search.php?cmd=display&search_for=parse_url&x=0&y=0
And this bug was actually fixed in never versions. So IMO we either live with the bugs (which are usually only for rather uncommon URIs), or we can implement our URI parser (which is rather simple as the RFC 3986 already provides a regex for that). Or somebody with C skills could finally fix the PHP parse_url implementation.

Tobion:
@mtdowling fixed tests

PR #1510, opened 2016-06-29
Only read up to Content-Length in stream wrapper

mtdowling:
This commit updates the stream wrapper to only read up to the number of
bytes returned in the Content-Length header when draining a stream
synchronously.

@Yanacek @Tobion

### DIFF
--- a/src/Handler/StreamHandler.php
+++ b/src/Handler/StreamHandler.php
@@ -105,7 +105,12 @@ class StreamHandler
         $headers = \GuzzleHttp\headers_from_lines($hdrs);
         list ($stream, $headers) = $this->checkDecode($options, $headers, $stream);
         $stream = Psr7\stream_for($stream);
-        $sink = $this->createSink($stream, $options);
+        $sink = $stream;
+
+        if (strcasecmp('HEAD', $request->getMethod())) {
+            $sink = $this->createSink($stream, $options);
+        }
+
         $response = new Psr7\Response($status, $headers, $sink, $ver, $reason);
 
         if (isset($options['on_headers'])) {
@@ -118,6 +123,8 @@ class StreamHandler
             }
         }
 
+        // Do not drain when the request is a HEAD request because they have
+        // no body.
         if ($sink !== $stream) {
             $this->drain(
                 $stream,
--- a/src/Handler/StreamHandler.php
+++ b/src/Handler/StreamHandler.php
@@ -119,7 +119,11 @@ class StreamHandler
         }
 
         if ($sink !== $stream) {
-            $this->drain($stream, $sink);
+            $this->drain(
+                $stream,
+                $sink,
+                $response->getHeaderLine('Content-Length')
+            );
         }
 
         $this->invokeStats($options, $request, $startTime, $response, null);
@@ -181,13 +185,27 @@ class StreamHandler
      *
      * @param StreamInterface $source
      * @param StreamInterface $sink
+     * @param string          $contentLength Header specifying the amount of
+     *                                       data to read.
      *
      * @return StreamInterface
      * @throws \RuntimeException when the sink option is invalid.
      */
-    private function drain(StreamInterface $source, StreamInterface $sink)
-    {
-        Psr7\copy_to_stream($source, $sink);
+    private function drain(
+        StreamInterface $source,
+        StreamInterface $sink,
+        $contentLength
+    ) {
+        // If a content-length header is provided, then stop reading once
+        // that number of bytes has been read. This can prevent infinitely
+        // reading from a stream when dealing with servers that do not honor
+        // Connection: Close headers.
+        Psr7\copy_to_stream(
+            $source,
+            $sink,
+            strlen($contentLength) > 0 ? (int) $contentLength : -1
+        );
+
         $sink->seek(0);
         $source->close();
 
--- a/tests/Handler/StreamHandlerTest.php
+++ b/tests/Handler/StreamHandlerTest.php
@@ -160,6 +160,25 @@ class StreamHandlerTest extends \PHPUnit_Framework_TestCase
         fclose($stream);
     }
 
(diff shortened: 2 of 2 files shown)


mtdowling:
Note that the test failure is due to an unrelated change in a package Guzzle depends on. @Tobion is addressing this issue.

src/Handler/StreamHandler.php
```diff
@@ -119,7 +119,11 @@ private function createResponse(
         }
 
         if ($sink !== $stream) {
-            $this->drain($stream, $sink);
+            $this->drain(
+                $stream,
+                $sink,
+                $response->getHeaderLine("Content-Length")
```
Tobion: I think single quotes are used so far

src/Handler/StreamHandler.php
```diff
@@ -181,13 +185,26 @@ private function checkDecode(array $options, array $headers, $stream)
      *
      * @param StreamInterface $source
      * @param StreamInterface $sink
+     * @param null|int        $contentLength Amount of data to read.
```
Tobion: `getHeaderLine` returns a string in any case. so null is not used currently and int is not really true.

src/Handler/StreamHandler.php
Tobion: Maybe something like `is_numeric($response->getHeaderLine('Content-Length') ? (int) $response->getHeaderLine('Content-Length') : null`

mtdowling:
Good feedback, @Tobion. I've pushed up a commit to address the feedback.

src/Handler/StreamHandler.php
```diff
@@ -181,13 +185,27 @@ private function checkDecode(array $options, array $headers, $stream)
...
+        $contentLength
+    ) {
+        // If a content-length header is provided, then stop reading once
+        // that number of bytes has been read. This can prevent infinitely
+        // reading from a stream when dealing with servers that do not honor
+        // Connection: Close headers.
+        Psr7\copy_to_stream(
+            $source,
+            $sink,
+            strlen($contentLength) > 0 ? (int) $contentLength : -1
```
Tobion: If Content-Length is not numeric, it will cast to 0. Not sure if that is wanted.

_test.php
```diff
@@ -0,0 +1,74 @@
...
+
+$stack = new HandlerStack();
+$formatter = new MessageFormatter();
+$stack->setHandler(\GuzzleHttp\choose_handler());
+$stack->push(Middleware::httpErrors());
+$client = new Client(['handler' => $stack,]);
+
+function overviewRequestFactory($page = 1)
+{
+    return new Request('GET', sprintf('https://kat.cr/movies/%d/', $page));
```
Tobion: what is this file for?

Tobion:
Responses to HEAD requests can sent a Content-Length header that is not actually the size of the response body:

> in the case of the HEAD method, the size of the entity-body that would have been sent had the request been a GET.

Is it handled somewhere that certain status codes or HEAD responses must not have a body?

_test.php
mtdowling: Doh. Removed.

src/Handler/StreamHandler.php
mtdowling: Actually, that's probably ideal. It would prevent us from being able to read any data, which is better than reading until a timeout.

Tobion:
Looking at the implementation of copy_to_stream ( https://github.com/guzzle/psr7/blob/master/src/functions.php#L385 ) it seems that it will require the amount of memory that is given with the maxLength size. So if the response is 1 GB, it will also require 1 GB of memory to copy the stream. This doesn't seem right. Shouldn't it read the stream in batches just as it is done when `$maxLen === -1`?

mtdowling:
Yeah, this is not taking HEAD requests into account. I'll push a fix.

As for copy_to_stream, you're right. The implementation could be improved.

mtdowling:
No longer draining the stream for HEAD requests. The poor maxLen implementation can be addressed separately on the PSR-7 repo. Rebased on master, so hopefully the tests will pass now.

Tobion:
`createSink` still fopens a stream even for HEAD requests. This doesn't seem necessary.

mtdowling:
Nice catch. I pushed a commit to no longer create a new sink when sending a HEAD request.

src/Handler/StreamHandler.php
```diff
@@ -118,8 +124,14 @@ private function createResponse(
             }
         }
 
-        if ($sink !== $stream) {
-            $this->drain($stream, $sink);
+        // Do not drain when the request is a HEAD request because they have
+        // no body.
+        if ($hasResponseBody && $sink !== $stream) {
```
Tobion: `$hasResponseBody` seems useless as the second condition is never true in this case anyway

mtdowling: True. Fixed.

src/Handler/StreamHandler.php
```diff
@@ -105,7 +105,13 @@ private function createResponse(
         $headers = \GuzzleHttp\headers_from_lines($hdrs);
         list ($stream, $headers) = $this->checkDecode($options, $headers, $stream);
         $stream = Psr7\stream_for($stream);
-        $sink = $this->createSink($stream, $options);
+        $sink = $stream;
+
+        $hasResponseBody = strcasecmp("HEAD", $request->getMethod());
```
mtdowling: Doh... All my recent Java development is becoming habit.

Tobion:
:+1:

src/Handler/StreamHandler.php
mattzuba: An actual contentLength of 0, or one that is cast to 0 seems to be causing issues with fread in the Psr7 package, see issue #1539.

I hit this fread warning today when calling some AWS API commands which caused my Laravel app to bomb out.

PR #1511, opened 2016-06-30
Equal to identical

cheprasov:
In PHP any string, that does not started from number is equal (==) 0.
Example:
```php
var_dump(0 == '+'); // true
var_dump(0 == '*'); // true
var_dump(0 == '&'); // true
var_dump(0 == 'foo'); // true

var_dump(0 != '&'); // false
```
To avoid this behavior is better to use identical operator (===)
Also, === is faster than ==


Tobion:
:+1:

PR #1513, opened 2016-07-01
Fixed a PHP namespace

javiereguiluz:


GrahamCampbell:
The example is already fine without this surely? The example is not namespaced.

javiereguiluz:
@GrahamCampbell if line 21 is correct, then should we update the line 33? Thanks.

GrahamCampbell:
Looks like line 33 has an unnecessary slash too.

Best leave it to @mtdowling to decide what he wants though. :)

Tobion:
:+1: good this way

PR #1514, opened 2016-07-01
ignore URI fragment when dereferencing

Tobion:
The URI fragment should be ignored when dereferencing a URI and is not meant to be sent.

> Fragment identifiers have a special role in information retrieval
   systems as the primary form of client-side indirect referencing,
   allowing an author to specifically identify aspects of an existing
   resource that are only indirectly provided by the resource owner.  As
   such, the fragment identifier is not used in the scheme-specific
   processing of a URI; instead, the fragment identifier is separated
   from the rest of the URI prior to a dereference, and thus the
   identifying information within the **fragment itself is dereferenced
   solely by the user agent, regardless of the URI scheme**.

https://tools.ietf.org/html/rfc3986#section-3.5

### DIFF
--- a/src/Handler/CurlFactory.php
+++ b/src/Handler/CurlFactory.php
@@ -194,7 +194,7 @@ class CurlFactory implements CurlFactoryInterface
         $conf = [
             '_headers'             => $easy->request->getHeaders(),
             CURLOPT_CUSTOMREQUEST  => $easy->request->getMethod(),
-            CURLOPT_URL            => (string) $easy->request->getUri(),
+            CURLOPT_URL            => (string) $easy->request->getUri()->withFragment(''),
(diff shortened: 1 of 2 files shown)


src/Handler/CurlFactory.php
```diff
@@ -495,12 +495,14 @@ private static function retryFailedRewind(
 
     private function createHeaderFn(EasyHandle $easy)
     {
-        if (!isset($easy->options['on_headers'])) {
```
Tobion: This was only rewritten so phpstorm understands that `$onHeaders` can be a callable or null and does not mark this with an error.

Tobion:
I don't see where `StreamHandler::add_debug` is used at all?

PR #1519, opened 2016-07-06
increase length of response

mroslyak:
would like to increase response from 120 to 240.


mtdowling:
This is just a summary. You can get the full response by calling $e->getResponse()->getBody().

PR #1524, opened 2016-07-13
Fix documentation about default connect_timeout

mweibel:
According to [CurlFactory::getDefaultConf()](https://github.com/guzzle/guzzle/blob/master/src/Handler/CurlFactory.php#L200) the default connect timeout is 150s instead of the written 0s.

### DIFF
--- a/docs/request-options.rst
+++ b/docs/request-options.rst
@@ -256,9 +256,9 @@ connect_timeout
 ---------------
 
 :Summary: Float describing the number of seconds to wait while trying to connect
-        to a server. Use ``0`` to wait indefinitely (the default behavior).
+        to a server. Use ``0`` to wait indefinitely.
 :Types: float
-:Default: ``0``
+:Default: ``150``
 :Constant: ``GuzzleHttp\RequestOptions::CONNECT_TIMEOUT``
 
 .. code-block:: php


Tobion:
@mweibel seems to be right here.

> Use 0 to wait indefinitely

This doesn't seem to be working as well.
https://curl.haxx.se/libcurl/c/CURLOPT_CONNECTTIMEOUT_MS.html has a default value of 300000. And it explicitily says

> Set to zero to switch to the default built-in connection timeout - 300 seconds

So 0 does not mean indefinitely for connect_timeout. But for timeout it seems to be the case: https://curl.haxx.se/libcurl/c/CURLOPT_TIMEOUT_MS.html

But timeout of 0 also does not seem to work for the StreamHandler.
According to http://php.net/manual/de/context.http.php

> By default the default_socket_timeout php.ini setting is used.

which is 60 by default. So the behavior is inconsistent bewtween the curl handler and the stream handler.

GrahamCampbell:
What's the status of this PR?

GrahamCampbell:
It looks like the default is actually 300 seconds, in curl, if no option is provided, as @Tobion says.

Nyholm:
Closing due inactivity.

Issue #1528, opened 2016-07-15
Wrong documentation for allow_redirects


mstrom:
Documentation has those default values 

[
    'max'             => 5,
    'strict'          => false,
    'referer'         => **true**,
    'protocols'       => ['http', 'https'],
    'track_redirects' => false
]

But in code (https://github.com/guzzle/guzzle/blob/master/src/RedirectMiddleware.php#L26)

$defaultSettings = [
        'max'             => 5,
        'protocols'       => ['http', 'https'],
        'strict'          => false,
        'referer'         => **false**,
        'track_redirects' => false,
    ];

So 'referer' is mismatch

PR #1529, opened 2016-07-16
Fixing issue #1528 - default value of allow_redirects.referer

mstrom:
Issue #1528

### DIFF
--- a/docs/request-options.rst
+++ b/docs/request-options.rst
@@ -30,7 +30,7 @@ allow_redirects
         [
             'max'             => 5,
             'strict'          => false,
-            'referer'         => true,
+            'referer'         => false,
             'protocols'       => ['http', 'https'],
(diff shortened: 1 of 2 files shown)


mtdowling:
Thanks for the PR!

In this case, I think I'd actually rather update the documentation to reflect the code rather than the code to reflect the docs.

mstrom:
@mtdowling are you sure? Curl has referer option (CURLOPT_AUTOREFERER) set to true on default

http://php.net/manual/en/function.curl-setopt.php

mtdowling:
Yeah, curl doesn't have this default behavior from what I can tell. I'm also hesitant to start sending this data with redirects when we haven't always done that.

mstrom:
@mtdowling I changed documentation please review

mtdowling:
thanks!

PR #1544, opened 2016-08-03
Fix issue with dublicating providing eventName into listener

Enelar:
Issue causing on php 7.0.1, making impossible to run any request.


GrahamCampbell:
Guzzle 4 isn't supported I think?

mtdowling:
Correct. Guzzle 4 is dead. Upgrading to Guzzle 5 (which is basically dead too) should take only a couple minutes though.

PR #1552, opened 2016-08-05
MutableClient first implementation

sangio90:
Referring to the pr #1479.
This is a very simple implementation of a mutable version of the Client, the class extends the Client base class.
By now this class only exposes one public method that does the trick, it can be improved in the future.
The only change in the main class is the visibility of the 'config' attribute, which should be declared protected to allow this.

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -25,7 +25,7 @@ use Psr\Http\Message\ResponseInterface;
 class Client implements ClientInterface
 {
     /** @var array Default request options */
-    private $config;
+    protected $config;
 
     /**
      * Clients accept an array of constructor parameters.
--- a/src/MutableClient.php
+++ b/src/MutableClient.php
@@ -1,0 +1,20 @@
+<?php
+
+namespace GuzzleHttp;
+
+/**
+ * Class MutableClient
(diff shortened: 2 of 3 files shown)


jeskew:
Making a private property protected is not a backwards compatible change, as any class that extended `GuzzleHttp\Client` and had its own private `$config` property would cause a fatal error. See [this fiddle](https://3v4l.org/TWkiQ) for an example.

Rather than altering the `$config` property and extending Guzzle's core client class, this PR could instead provide a mutable implementation of `GuzzleHttp\ClientInterface` that delegates to immutable instances of `GuzzleHttp\Client` under the hood.

GrahamCampbell:
👎 for supporting mutability. It was made immutable by design on purpose surely?

sagikazarmark:
I am also not a huge fan of mutability. It allows bad design in your code.

sagikazarmark:
I am going to close this from one week from now.

sagikazarmark:
I am closing this as it goes against the design decision of the library. @jeskew recommended a a better solution, but honestly, I don't really want to see a mutable client in Guzzle.

PR #1565, opened 2016-08-29
Only add scheme when host is present

pkruithof:
See guzzle/psr7#114

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -145,7 +145,7 @@ class Client implements ClientInterface
             $uri = Psr7\Uri::resolve(Psr7\uri_for($config['base_uri']), $uri);
         }
 
-        return $uri->getScheme() === '' ? $uri->withScheme('http') : $uri;
+        return $uri->getScheme() === '' && $uri->getHost() !== '' ? $uri->withScheme('http') : $uri;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


tests/ClientTest.php
```diff
@@ -641,4 +641,16 @@ public function testHttpDefaultSchemeIfUriHasNone()
...
         $this->assertSame('http://example.org/test', (string) $mockHandler->getLastRequest()->getUri());
     }
+
+    public function testOnlyAddSchemeWhenHostIsPresent()
+    {
+        $mockHandler = new MockHandler([new Response()]);
+        $client = new Client(['handler'  => $mockHandler]);
+
+        $client->request('GET', 'baz');
+        $this->assertEquals(
```
Tobion: please use the same code as above: assertSame with string cast

src/Client.php
```diff
@@ -145,7 +145,11 @@ private function buildUri($uri, array $config)
             $uri = Psr7\Uri::resolve(Psr7\uri_for($config['base_uri']), $uri);
         }
 
-        return $uri->getScheme() === '' ? $uri->withScheme('http') : $uri;
+        if ($uri->getScheme() === '' && !empty($uri->getHost())) {
```
Tobion: `$uri->getHost() !== ''`

Tobion: and please keep the ternary operator

Tobion:
:+1:

PR #1569, opened 2016-08-30
Added curl ntlm auth support

miguelbalboa:


miguelbalboa:
Same as #1381 but with doc and test.

I'm wondering if StreamHandler::createStream is the right place to through exception if we try to use NTLM with PHP Stream.

Something like:

```php
        // Microsoft NTLM authentication only supported with curl handler
        if (isset($options['auth'])) {
            if (is_array($options['auth']) && isset($options['auth'][2])) {
                if ('ntlm' == $options['auth'][2]) {
                    throw new \InvalidArgumentException('Microsoft NTLM authentication only supported with curl handler');
                }
            }
        }
```

Or in Client::applyOptions ?

sagikazarmark:
Thanks for your contribution.

Can you please squash your commits (and rebase from master)?

> Merge branch 'master' of https://github.com/guzzle/guzzle

This shouldn't be there

> I'm wondering if StreamHandler::createStream is the right place to through exception if we try to use NTLM with PHP Stream.

Since in `Client::applyOptions` you don't know about the handlers, I would rather do it in `StreamHandler::createStream`. Options are normalized there anyway (eg. on_headers callback is checked there).

I would also move all the conditions into one if: same amount of logical expressions to evaluate, but less branching and more readable code.

Do you think you can finish this @miguelbalboa ?

miguelbalboa:
@sagikazarmark 
squashed and added NTLM Exception in ``StreamHandler::createStream``.

rudott:
@sagikazarmark do you know when this will be merged?

sagikazarmark:
I don't have an exact ETA yet. Current aim is to get 6.3 around end of February.

sagikazarmark:
Thanks for your work on this, sorry for the painfully long wait.

garethellis36:
FYI [the docs](http://docs.guzzlephp.org/en/latest/request-options.html#auth) indicate that this feature is currently supported, but I just wasted half an hour debugging before I found that NTLM auth isn't supported in `Client.php` in 6.2.3.

PR #1562, opened 2016-08-30
Retry-After Header in Guzzle 6

scheb:
I'm currently migrating a project from Guzzle 5 to 6. In 5 i was using https://github.com/guzzle/retry-subscriber, which was able to read the Retry-After header from the response and delay the retry accordingly. Now with Guzzle 6 middleware, I don't have access to the Response in the delay function.

What's the proposed solution to do retries which respect the Retry-After header? Am I missing something?


scheb:
Can you please review? Thanks!

PR #1570, opened 2016-08-31
Obfuscates URI if there is an username and a password present

digitalcrab:
Lets assume that you include username and password into the URI. Currently, if the exception accrued the error message looks like so:

```
Client error: `POST http://username:secrect@hostname.com/somePath` resulted in a `400 Bad Request` response: ......,  (truncated...)
```

It lead to security issue because password is shown like it is.

This commit will hide the password (replace it with `***`).

### DIFF
--- a/src/Exception/RequestException.php
+++ b/src/Exception/RequestException.php
@@ -4,6 +4,7 @@ namespace GuzzleHttp\Exception;
 use Psr\Http\Message\RequestInterface;
 use Psr\Http\Message\ResponseInterface;
 use GuzzleHttp\Promise\PromiseInterface;
+use Psr\Http\Message\UriInterface;
 
 /**
  * HTTP Request exception
@@ -89,12 +90,15 @@ class RequestException extends TransferException
             $className = __CLASS__;
         }
 
+        $uri = $request->getUri();
+        $uri = static::obfuscateUri($uri);
+
         // Server Error: `GET /` resulted in a `404 Not Found` response:
         // <html> ... (truncated)
         $message = sprintf(
             '%s: `%s` resulted in a `%s` response',
             $label,
-            $request->getMethod() . ' ' . $request->getUri(),
+            $request->getMethod() . ' ' . $uri,
             $response->getStatusCode() . ' ' . $response->getReasonPhrase()
         );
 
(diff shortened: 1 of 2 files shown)


sagikazarmark:
Although I like the idea, this could be a problem when you want to handle an exception and use the original data inside. Not sure what's the best option though. Maybe we should make this configurable?

digitalcrab:
@sagikazarmark how you gonna use original data in this case? parse exception message to find the URI? I don't think that this is a case.

sagikazarmark:
Sorry, I missed that the obfuscated URI is only used for the exception message.

src/Exception/RequestException.php
```diff
@@ -142,6 +146,24 @@ public static function getResponseBodySummary(ResponseInterface $response)
     }
 
     /**
+     * Obfuscates URI if there is an username and a password present
+     *
+     * @param UriInterface $uri
+     *
+     * @return UriInterface
+     */
+    public static function obfuscateUri($uri)
```
sagikazarmark: Do we need this as public static? I would say make it private and test only the behaviour of the exception without exposing this as a behaviour too.

tests/Exception/RequestExceptionTest.php
```diff
@@ -152,4 +152,19 @@ public function testCanProvideHandlerContext()
         $e = new RequestException('foo', $r, null, null, ['bar' => 'baz']);
         $this->assertEquals(['bar' => 'baz'], $e->getHandlerContext());
     }
+
+    public function testObfuscateUrlWithUsername()
+    {
+        $r = new Request('GET', 'http://username@www.oo.com');
+        $e = RequestException::create($r, new Response(500));
+        $this->assertContains('username@', $e->getMessage());
```
sagikazarmark: I would say you should test the whole returned URL here to make sure the rest of the URL is not modified.

tests/Exception/RequestExceptionTest.php
```diff
@@ -152,4 +152,19 @@ public function testCanProvideHandlerContext()
...
+    }
+
+    public function testObfuscateUrlWithUsernameAndPassword()
+    {
+        $r = new Request('GET', 'http://user:password@www.oo.com');
+        $e = RequestException::create($r, new Response(500));
+        $this->assertNotContains('user:password', $e->getMessage());
```
sagikazarmark: Testing if something is not in the URL can be omitted if you test the whole URL here too.

src/Exception/RequestException.php
```diff
@@ -142,6 +146,24 @@ public static function getResponseBodySummary(ResponseInterface $response)
     }
 
     /**
+     * Obfuscates URI if there is an username and a password present
+     *
+     * @param UriInterface $uri
+     *
+     * @return UriInterface
+     */
+    private static function obfuscateUri($uri)
```
sagikazarmark: No need static either IMHO.

digitalcrab: the call of this function happens earlier than we have exception object. So, it has to be a static function.

sagikazarmark: Grrrrr, github diff. Fair enough.

sagikazarmark:
Thanks @digitalcrab

PR #1575, opened 2016-09-03
Removed ugly code - Converted if/elses to switches

vijayant123:
Hello! This is a quick fix to some ugly nested if-elses that were buggering me.

Also putting the methods and requests URI in switch cases allows anyone to add his/her own custom URI to the server and test it.

Regards,
Vijayant

### DIFF
--- a/tests/server.js
+++ b/tests/server.js
@@ -125,60 +125,93 @@ var GuzzleServer = function(port, log) {
   };
 
   var controlRequest = function(request, req, res) {
-    if (req.url == '/guzzle-server/perf') {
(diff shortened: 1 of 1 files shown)


sagikazarmark:
Thanks for you r contribution.

Personally I don't think this is more readable. Also, switches are slower than ifs, so this is bad from a performance optimization point of view.

sagikazarmark:
Closing as it adds no real benefits. Thank you for the contribution though.

vijayant123:
Okay! Thanks BTW.

Issue #1539, opened 2016-07-25
"fread(): Length parameter must be greater than 0" after updating to 6.2.1


Jalle19:
I'm getting this on some requests, but not all. All requests are similar but still only some fail so I'm not sure what is going on.

```
PHP Warning:  fread(): Length parameter must be greater than 0 in /home/negge/Projects/vagrant-registry-generator/vendor/guzzlehttp/psr7/src/Stream.php on line 211
PHP Stack trace:
PHP   1. {main}() /home/negge/Projects/vagrant-registry-generator/bin/vagrant-registry-generator:0
PHP   2. Symfony\Component\Console\Application->run() /home/negge/Projects/vagrant-registry-generator/bin/vagrant-registry-generator:10
PHP   3. Symfony\Component\Console\Application->doRun() /home/negge/Projects/vagrant-registry-generator/vendor/symfony/console/Application.php:117
PHP   4. Symfony\Component\Console\Application->doRunCommand() /home/negge/Projects/vagrant-registry-generator/vendor/symfony/console/Application.php:186
PHP   5. Symfony\Component\Console\Command\Command->run() /home/negge/Projects/vagrant-registry-generator/vendor/symfony/console/Application.php:815
PHP   6. Jalle19\VagrantRegistryGenerator\VagrantRegistryGenerator->execute() /home/negge/Projects/vagrant-registry-generator/vendor/symfony/console/Command/Command.php:256
PHP   7. Jalle19\VagrantRegistryGenerator\Registry\Writer->write() /home/negge/Projects/vagrant-registry-generator/src/VagrantRegistryGenerator.php:66
PHP   8. Jalle19\VagrantRegistryGenerator\Registry\Writer->writeManifest() /home/negge/Projects/vagrant-registry-generator/src/Registry/Writer.php:74
PHP   9. League\Flysystem\Filesystem->put() /home/negge/Projects/vagrant-registry-generator/src/Registry/Writer.php:106
PHP  10. League\Flysystem\AwsS3v3\AwsS3Adapter->update() /home/negge/Projects/vagrant-registry-generator/vendor/league/flysystem/src/Filesystem.php:98
PHP  11. League\Flysystem\AwsS3v3\AwsS3Adapter->upload() /home/negge/Projects/vagrant-registry-generator/vendor/league/flysystem-aws-s3-v3/src/AwsS3Adapter.php:120
[...]

PR #1577, opened 2016-09-05
handle drain case where content-length is the literal string zero

michaelbutler:
# What

Fixes issue https://github.com/guzzle/guzzle/issues/1539

```
PHP Warning:  fread(): Length parameter must be greater than 0
```

If the Content-Length header is zero (`0`) a PHP Warning will be generated because the string length of `0` is 1, which ends up passing the integer `0` to `fread`. The intention should be passing `-1` to `fread` -- reading the entire stream.

I added a unit test for this case, which if you copy the unit test to master you'll see that the test suite will fail. With this small fix, the unit test suite passes and no PHP Warning is generated.

# Why

A `Content-Length: 0` header could happen, such as with Amazon AWS. 

# Test

Run the test suite: `vendor/bin/phpunit tests/` and everything should pass.

### DIFF
--- a/src/Handler/StreamHandler.php
+++ b/src/Handler/StreamHandler.php
@@ -210,7 +210,7 @@ class StreamHandler
         Psr7\copy_to_stream(
             $source,
             $sink,
-            strlen($contentLength) > 0 ? (int) $contentLength : -1
+            (strlen($contentLength) > 0 && (int) $contentLength > 0) ? (int) $contentLength : -1
         );
 
         $sink->seek(0);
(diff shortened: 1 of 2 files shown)


michaelbutler:
@mtdowling it didn't look like anyone had did this yet, so here it is.

sagikazarmark:
This looks good to me

src/Handler/StreamHandler.php
```diff
@@ -210,7 +210,7 @@ private function drain(
         Psr7\copy_to_stream(
             $source,
             $sink,
-            strlen($contentLength) > 0 ? (int) $contentLength : -1
+            (strlen($contentLength) > 0 && (int) $contentLength > 0) ? (int) $contentLength : -1
```
Nyholm: Wouldn't `empty($contentLength)` do the same?

`$contentLength` comes from `getHeaderLine` which will return a string with the content length or an empty string.

sagikazarmark: It would, but it's more error prone IMO, because you would still need to check if it's a numeric value.

Nyholm:
Thank you for this PR. It is great that you added the unit test. 

I'm :+1:

PR #1580, opened 2016-09-07
Handle on_headers option in MockHandler

### LINKED ISSUE #1536: MockHandler does not call on_headers callback
Using the MockHandler you cannot test the `on_headers` callback as it does not call it at all.

Codeklopper:
Should fix #1536

### DIFF
--- a/src/Handler/MockHandler.php
+++ b/src/Handler/MockHandler.php
@@ -1,6 +1,7 @@
 <?php
 namespace GuzzleHttp\Handler;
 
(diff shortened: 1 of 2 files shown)


flashek:
+1

flashek:
BTW Do You have idea how to mock `request` method utilising `on_headers` in unit tests? I tried to do this today, but I lost :|

sagikazarmark:
Thank you @Codeklopper

PR #1585, opened 2016-09-14
fix milliseconds timeout problem

bai471051357:
"If libcurl is built to use the standard system name resolver, that portion of the transfer will still use full-second resolution for timeouts with a minimum timeout allowed of one second."

The problem is that on (Li|U)nix, when libcurl uses the standard name resolver, a SIGALRM is raised during name resolution which libcurl thinks is the timeout alarm.

The solution is to disable signals using CURLOPT_NOSIGNAL.

### DIFF
--- a/src/Handler/CurlFactory.php
+++ b/src/Handler/CurlFactory.php
@@ -379,6 +379,13 @@ class CurlFactory implements CurlFactoryInterface
             $conf[CURLOPT_CONNECTTIMEOUT_MS] = $options['connect_timeout'] * 1000;
         }
 
+        if (strtoupper(substr(PHP_OS, 0, 3)) !== 'WIN') {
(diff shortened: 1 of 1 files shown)


bai471051357:
please read my pull request

sagikazarmark:
Give us some time please, this is not something we can validate via static analysis of the code, we will have to do some research as well. Thank you very much for your PR, will be reviewed shortly.

bai471051357:
if you set timeout < 1s,  curl return timeout error directly。

sagikazarmark:
Thanks for your contribution. This is a duplicate of #1718, which was created after this one, but it seems to be more precise, so I am going to merge it soon. Thanks for your efforts though and sorry for the long wait.

PR #1592, opened 2016-09-19
Make retry with different request possible

xjchengo:
It often needs change request a little when retrying a request. For example, `access_token` is expired so request failed. When retrying this request, we need to use new `access_token`.  But `Psr\Http\Message\RequestInterface` is immutable. This PR makes retry with different request possible.

### DIFF
--- a/src/RetryMiddleware.php
+++ b/src/RetryMiddleware.php
@@ -6,8 +6,8 @@ use Psr\Http\Message\RequestInterface;
 use Psr\Http\Message\ResponseInterface;
 
 /**
- * Middleware that retries requests based on the boolean result of
- * invoking the provided "decider" function.
+ * Middleware that retries requests based on the result of invoking
+ * the provided "decider" function.
  */
 class RetryMiddleware
 {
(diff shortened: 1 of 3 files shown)


sagikazarmark:
Hm, this looks interesting, wonder why I missed it. I will check if I can include it in the 6.3 release.

sagikazarmark:
Can you please rebase your branch? Also, can you please provide some tests?

src/RetryMiddleware.php
```diff
@@ -74,14 +74,17 @@ public function __invoke(RequestInterface $request, array $options)
...
-            if (!call_user_func(
+            $result = call_user_func(
                 $this->decider,
                 $options['retries'],
                 $req,
                 $value,
                 null
-            )) {
+            );
+            if (!$result) {
```
dzubchik: Why should we do type-casting? As described in phpdoc, decider should return false to reject next retry. Why don't just check $result===false?

src/RetryMiddleware.php
```diff
@@ -74,14 +74,17 @@ public function __invoke(RequestInterface $request, array $options)
...
                 return $value;
+            } elseif ($result instanceof RequestInterface) {
```
dzubchik: And previous statement is terminated, so we can omit **else**.

src/RetryMiddleware.php
xjchengo: Previous code did type-casting. Using `$result===false`  breaks backward compatibility.

src/RetryMiddleware.php
xjchengo: ok

xjchengo:
Tests have been added.

sagikazarmark:
The implementation looks good to me. Can you please also add a note about this feature to the middleware's docblock?

Nyholm:
I like this. But as Mark says, the doc block needs to be updated.

Nyholm:
Also, make a comment on why we cannot use strict comparison.

Nyholm:
I think so. Could you rebase this?

Nyholm:
Closing due inactivity.

Nyholm:
Feel free to reopen this PR if someone wants to work on it

SSEsmaeeli:
Any progress regarding this feature?

PR #1595, opened 2016-09-21
don't use eventName as parameter and variable

LawnGnome:
This cherry picks @xabbuh's fix from #1393 to the 4.x branch, thereby avoiding `Fatal error: Cannot use lexical variable $eventName as a parameter name` when running Guzzle 4 on PHP 7.1.

### DIFF
--- a/src/Event/Emitter.php
+++ b/src/Event/Emitter.php
@@ -44,8 +44,7 @@ class Emitter implements EmitterInterface
     public function once($eventName, callable $listener, $priority = 0)
     {
         $onceListener = function (
-            EventInterface $event,
-            $eventName
+            EventInterface $event
         ) use (&$onceListener, $eventName, $listener, $priority) {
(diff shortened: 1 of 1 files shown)


sagikazarmark:
Can you please take a look at the failing tests?

sagikazarmark:
All right, I checked the 4.x branch and the tests are indeed broken on 5.5 and 5.6 (and hhvm).

Seems to me that it is related to `php://temp` streams based on this: https://bugs.php.net/bug.php?id=68948

In theory this is broken once and for all, unless we use a different type of stream.

Anyway, I think this one should be merged and a PATCH should be tagged, then 4.x should be frozen once and for all.

WDYT @guzzle/maintainers ?

sagikazarmark:
Okay, merging this for now, tag will follow if no problems arise.

sagikazarmark:
Thanks

Issue #1605, opened 2016-10-01
Warning generated when 404 response with "Content-Length: 0" received


mpdude:
I am communicating with an API that might return 404 responses and sets `Content-Length: 0`.

This causes a warning in `\GuzzleHttp\Handler\StreamHandler::drain` when attempting to read `0` bytes from the stream.

I think the method should read
```php
    private function drain(
        StreamInterface $source,
        StreamInterface $sink,
        $contentLength
    ) {
        $length = strlen($contentLength) > 0 ? (int) $contentLength : -1;
        // If a content-length header is provided, then stop reading once
        // that number of bytes has been read. This can prevent infinitely
        // reading from a stream when dealing with servers that do not honor
        // Connection: Close headers.
        if ($length > 0) {
            Psr7\copy_to_stream(
                $source,
                $sink,
                $length
            );
        }

        $sink->seek(0);
        $source->close();

        return $sink;
    }
```

michaelbutler:
are you sure you're using the latest guzzle version? this was fixed in https://github.com/guzzle/guzzle/pull/1577

mpdude:
Indeed. Missed that as it has not yet been released.

Thanks!

PR #1606, opened 2016-10-01
Fix for #1605

mpdude:


michaelbutler:
1. please provide a description for your PR
2. please check if this is really needed, i think when the content length becomes `-1` that means "copy all to stream" and your change doesn't do this.

mpdude:
Closing as this has already been addressed in #1577.

michaelbutler:
Also note that it's in dev master but hasn't been released yet. You can temporarily downgrade to an earlier version to get around it too.

PR #1608, opened 2016-10-02
Added cofiguration option to force client to use either ipv4 or ipv6 protocol (curl)

paveljanda:


src/Handler/CurlFactory.php
```diff
@@ -375,6 +375,14 @@ private function applyHandlerOptions(EasyHandle $easy, array &$conf)
             $conf[CURLOPT_TIMEOUT_MS] = $options['timeout'] * 1000;
         }
 
+        if (isset($options['force_ipv4'])) {
+            if (defined('CURLOPT_IPRESOLVE') && defined('CURL_IPRESOLVE_V4')) {
```
sagikazarmark: From which PHP version is this feature available? Is it PHP version dependent at all? For me it looks a little bit weird that we check the feature and then refer to the PHP version. I would rather check the PHP version (if that's still necessary) or say this curl feature is not available. WDYT?

src/Handler/CurlFactory.php
```diff
@@ -375,6 +375,14 @@ private function applyHandlerOptions(EasyHandle $easy, array &$conf)
...
+                $conf[CURLOPT_IPRESOLVE] = CURL_IPRESOLVE_V4;
+            } else {
+                throw new \RuntimeException('Can not force client to use IPV4 protocol on php version ' . phpversion());
```
sagikazarmark: Can we do the check in a reverse order? If the feature is not available throw an exception else just continue running. (You don't even need the else branch)

src/Handler/CurlFactory.php
paveljanda: Id should not be dependent on any PHP version. It just won't work when the curl library is not present.

src/Handler/CurlFactory.php
paveljanda: Roger that.

paveljanda:
@sagikazarmark Is it all right now?

src/Handler/CurlFactory.php
sagikazarmark: So if this feature comes with every version of curl and php, why do you need to check it's existence? Just two lines bellow another curl constant is used and it doesn't check curl existence, not to mention that it is CURL factory so it might probably be clear by this time if curl is present or not.

paveljanda: Sure. Should I remove the check?

paveljanda: Removed.

sagikazarmark:
Can you please also add some documentation?

https://github.com/guzzle/guzzle/blob/master/docs/request-options.rst

paveljanda:
There it is.

docs/request-options.rst
```diff
@@ -975,6 +975,27 @@ timeout
     // PHP Fatal error:  Uncaught exception 'GuzzleHttp\Exception\RequestException'
 
 
+.. force_ipv4-option:
+
+force_ipv4
+---------------
```
sagikazarmark: Can you please trim dashes to the length of the option?

docs/request-options.rst
```diff
@@ -975,6 +975,27 @@ timeout
...
+
+:Summary: Set to true if you want the HTTP handlers to use only ipv4 protocol.
+:Types: bool
+:Default: false
+:Constant: ``GuzzleHttp\RequestOptions::FORCE_IPV4``
+
+.. code-block:: php
+
+    // Force ipv4 protocol
+    $client->request('GET', '/foo', [GuzzleHttp\RequestOptions::FORCE_IPV4 => TRUE]);
```
sagikazarmark: Please make "true" lowercase

sagikazarmark: Also the standard is that you use the option name as string, not the constant

paveljanda: 1, Done.
2, Done, but i strongly recommend to use the constant, not the string.

sagikazarmark: Noted and agreed, but for the sake of consistency, we leave the string there

sagikazarmark:
@guzzle/maintainers please add some input here

mtdowling:
Could this be more general -- like force_ip_resolve => v4?

paveljanda:
@mtdowling Sure, great idea. Updated.

Nyholm:
Are we happy with the changes here?

sagikazarmark:
I am, it's also in the "Ready to merge" section in the Contribution project.

sagikazarmark:
Thanks @paveljanda

PR #1611, opened 2016-10-05
Add 'read_timeout' option

### LINKED ISSUE #1516: Reading on streamed body doesn't have a timeout
When reading from a streamed body, it is not possible to timeout. As a result, the client can get stuck trying to read the body. It can happen in many occasions, for instance if the remote end crashes or just forgets about you, the client will try to read from the socket forever (c.f. half-open tcp).

Example:

```
$client = new Client([
    'timeout' => 300,
    'stream' => true,
]);

$response = $client->get('http://example.com/');

$body = $response->getBody()->detach();

while (!feof($body)) {
    // this can hang forever
    $line = fgets($body);
}
```

I would be great if Guzzle allowed to specify a read timeout. I didn't find a way to apply a timeout myself.

arnaud-lb:
This adds a 'read_timeout' option, allowing the user to specify a read timeout for reads on the body of streamed requests.

Support is added only to the StreamHandler, since only this handler seems to support stream requests.

This fixes #1516

### DIFF
--- a/src/Handler/StreamHandler.php
+++ b/src/Handler/StreamHandler.php
@@ -308,9 +308,17 @@ class StreamHandler
         );
 
         return $this->createResource(
-            function () use ($request, &$http_response_header, $context) {
+            function () use ($request, &$http_response_header, $context, $options) {
                 $resource = fopen((string) $request->getUri()->withFragment(''), 'r', null, $context);
                 $this->lastHeaders = $http_response_header;
+
+                if (isset($options['read_timeout'])) {
+                    $readTimeout = $options['read_timeout'];
+                    $sec = (int) $readTimeout;
+                    $usec = ($readTimeout - $sec) * 100000;
+                    stream_set_timeout($resource, $sec, $usec);
+                }
+
                 return $resource;
             }
         );
--- a/src/RequestOptions.php
+++ b/src/RequestOptions.php
@@ -238,6 +238,12 @@ final class RequestOptions
     const TIMEOUT = 'timeout';
 
     /**
+     * read_timeout: (float, default=default_socket_timeout ini setting) Float describing
+     * the body read timeout, for stream requests.
+     */
+    const READ_TIMEOUT = 'read_timeout';
+
+    /**
      * version: (float) Specifies the HTTP protocol version to attempt to use.
      */
     const VERSION = 'version';
--- a/tests/Handler/StreamHandlerTest.php
+++ b/tests/Handler/StreamHandlerTest.php
@@ -7,6 +7,7 @@ use GuzzleHttp\Psr7;
 use GuzzleHttp\Psr7\Request;
 use GuzzleHttp\Psr7\Response;
 use GuzzleHttp\Psr7\FnStream;
+use GuzzleHttp\RequestOptions;
 use GuzzleHttp\Tests\Server;
 use GuzzleHttp\TransferStats;
 use Psr\Http\Message\ResponseInterface;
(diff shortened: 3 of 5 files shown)


src/Handler/StreamHandler.php
```diff
@@ -308,9 +308,17 @@ function () use ($context, $params) {
         );
 
         return $this->createResource(
-            function () use ($request, &$http_response_header, $context) {
+            function () use ($request, &$http_response_header, $context, $options) {
                 $resource = fopen((string) $request->getUri()->withFragment(''), 'r', null, $context);
                 $this->lastHeaders = $http_response_header;
+
+                if (isset($options['read_timeout'])) {
```
jeskew: This could use the RequestOptions constant defined in this PR.

src/Handler/StreamHandler.php
```diff
@@ -308,9 +308,17 @@ function () use ($context, $params) {
...
+                    $readTimeout = $options['read_timeout'];
+                    $sec = (int) $readTimeout;
+                    $usec = ($readTimeout - $sec) * 1000000;
```
jeskew: Wouldn't this always be 0? If a user supplies a readTimeout of `1`, then `$usec` will be set to `(1 - 1) * 1000000`.

src/Handler/StreamHandler.php
arnaud-lb: Other uses of the $options array in this class don't make use of RequestOptions constants, so I tried to keep it consistent.

arnaud-lb:
@jeskew: Thank you for the review. I updated the PR accordingly.

src/Handler/StreamHandler.php
jeskew: You're right; it's better to keep this consistent.

tests/Handler/StreamHandlerTest.php
```diff
@@ -654,4 +655,25 @@ public function testDrainsResponseAndReadsAllContentWhenContentLengthIsZero()
...
+            ]
+        )->wait();
+        $this->assertEquals(200, $response->getStatusCode());
+        $this->assertEquals('OK', $response->getReasonPhrase());
+        $body = $response->getBody()->detach();
+        $line = fgets($body);
+        $this->assertEquals("sleeping 60 seconds ...\n", $line);
+        $line = fgets($body);
+        $this->assertFalse($line);
+        $this->assertFalse(feof($body));
```
jeskew: According to the PHP docs, this value will be true if the stream has timed out. They don't mention any known exceptions.

src/RequestOptions.php
```diff
@@ -238,6 +238,12 @@
     const TIMEOUT = 'timeout';
 
     /**
+     * read_timeout: (float, default=default_socket_timeout) Float describing
```
sagikazarmark: Maybe mention that it's configured in PHP ini?

tests/server.js
```diff
@@ -155,6 +155,15 @@ var GuzzleServer = function(port, log) {
...
         res.end(body);
+      } else if (req.url == '/guzzle-server/read-timeout') {
+        if (that.log) {
+          console.log('Sleeping');
+        }
+        res.writeHead(200, 'OK');
+        res.write("sleeping 60 seconds ...\n");
+        setTimeout(function () {
+          res.end("slept 60 seconds\n");
+        }, 10000);
```
sagikazarmark: Is this 60 seconds?

arnaud-lb:
@sagikazarmark Thank you for the review, I've updated the PR accordingly

tests/server.js
```diff
@@ -155,6 +155,15 @@ var GuzzleServer = function(port, log) {
...
         res.end(body);
+      } else if (req.url == '/guzzle-server/read-timeout') {
+        if (that.log) {
+          console.log('Sleeping');
+        }
+        res.writeHead(200, 'OK');
+        res.write("sleeping 60 seconds ...\n");
+        setTimeout(function () {
+          res.end("slept 60 seconds\n");
+        }, 60*1000);
```
sagikazarmark: I am still not 100% sure what's happening here: it says 60 seconds, but in the test the read timeout is only one.

arnaud-lb: The server takes 60 seconds to send end of the response, and the client has a 1 second read timeout. So the client times out while reading the response.

arnaud-lb:
@sagikazarmark: Awesome, added documentation

src/Handler/StreamHandler.php
```diff
@@ -308,9 +308,17 @@ function () use ($context, $params) {
...
-            function () use ($request, &$http_response_header, $context) {
+            function () use ($request, &$http_response_header, $context, $options) {
                 $resource = fopen((string) $request->getUri()->withFragment(''), 'r', null, $context);
                 $this->lastHeaders = $http_response_header;
+
+                if (isset($options['read_timeout'])) {
+                    $readTimeout = $options['read_timeout'];
+                    $sec = (int) $readTimeout;
+                    $usec = ($readTimeout - $sec) * 100000;
+                    stream_set_timeout($resource, $sec, $usec);
```
sagikazarmark: To be honest I wonder if it would be better to pass the timeout to `stream_context_create`. It supports float values so it wouldn't require the magic above. One could even argue that the same functionality is already available via the undocumented `stream_context` option. WDYT @arnaud-lb ?

arnaud-lb: If it makes the code clearer, I'm all for it :) Which stream context option are you thinking about ? The "http.timeout" context option is used to implement Guzzle's "timeout" option, and I believe that it affects only the connection + reading the headers.

sagikazarmark: Hm, indeed, you are right, missed that.

sagikazarmark:
Thanks for working on this, sorry for the long wait.

PR #1612, opened 2016-10-05
Minor refactoring

mortenhauberg:


src/Client.php
```diff
@@ -190,13 +190,10 @@ private function configureDefaults(array $config)
...
-            // Add the User-Agent header if one was not already set.
-            foreach (array_keys($this->config['headers']) as $name) {
-                if (strtolower($name) === 'user-agent') {
-                    return;
-                }
-            }
+        }
+
+        $lowercasedHeaders = array_change_key_case($this->config['headers'], CASE_LOWER);
+        if (!isset($lowercasedHeaders['user-agent'])) {
```
sagikazarmark: I would still leave this in an else branch. This now runs nevertheless of the previous if, while it should only run if the previous condition is false.

src/Handler/CurlFactory.php
```diff
@@ -227,12 +227,13 @@ private function applyMethod(EasyHandle $easy, array &$conf)
         }
 
         $method = $easy->request->getMethod();
-        if ($method === 'PUT' || $method === 'POST') {
+        if (in_array($method, ['PUT', 'POST'], true)
```
sagikazarmark: I don't think this increases readability. What's the reasoning behind?

src/UriTemplate.php
```diff
@@ -131,9 +131,7 @@ private function expandMatch(array $matches)
 
                     if (!$isNestedArray) {
                         $var = rawurlencode($var);
-                        if ($parsed['operator'] === '+' ||
-                            $parsed['operator'] === '#'
-                        ) {
+                        if (in_array($parsed['operator'], ['+', '#'], true)) {
```
sagikazarmark: Same here: don't think it is better readable, even if shorter.

src/UriTemplate.php
```diff
@@ -185,7 +183,7 @@ private function expandMatch(array $matches)
                     $variable = substr($variable, 0, $value['position']);
                 }
                 $expanded = rawurlencode($variable);
-                if ($parsed['operator'] === '+' || $parsed['operator'] === '#') {
+                if (in_array($parsed['operator'], ['+', '#'], true)) {
```
sagikazarmark: And here too

src/Client.php
mortenhauberg: 👍🏼

src/Handler/CurlFactory.php
mortenhauberg: In my mind it does increase the readability, but I'll change it :)

src/UriTemplate.php
mortenhauberg: I'll revert :)

src/UriTemplate.php
mortenhauberg: 👍🏼

src/Handler/CurlFactory.php
sagikazarmark: Option one: "If the method is PUT or POST then..."
Option two: "if the method is one of PUT or POST then..."

For me the first one is logically more simple as we want a literally logical answer to this question. That's how I read it. :)

src/Handler/CurlFactory.php
```diff
@@ -227,12 +227,12 @@ private function applyMethod(EasyHandle $easy, array &$conf)
...
+        if (($method === 'PUT' || $method === 'POST') && !$easy->request->hasHeader('Content-Length')) {
             // See http://tools.ietf.org/html/rfc7230#section-3.3.2
-            if (!$easy->request->hasHeader('Content-Length')) {
-                $conf[CURLOPT_HTTPHEADER][] = 'Content-Length: 0';
-            }
-        } elseif ($method === 'HEAD') {
+            $conf[CURLOPT_HTTPHEADER][] = 'Content-Length: 0';
+        }
+
+        if ($method === 'HEAD') {
```
sagikazarmark: I am still confused about this: If the method is PUT or POST, I don't want to evaluate this second condition, so elseif is better. But in that case, this needs to be reverted too. Sorry.

mortenhauberg: Sure 👍🏼

src/Handler/CurlFactory.php
```diff
@@ -227,11 +227,9 @@ private function applyMethod(EasyHandle $easy, array &$conf)
         }
 
         $method = $easy->request->getMethod();
-        if ($method === 'PUT' || $method === 'POST') {
+        if (($method === 'PUT' || $method === 'POST') && !$easy->request->hasHeader('Content-Length')) {
```
sagikazarmark: Sorry, I meant this thing too. Right now if the method is PUT or POST, but the second condition evaluates to false the elseif condition is still evaluated.

mortenhauberg: You're totally right

sagikazarmark:
Sorry for asking you to revert most of your changes. :(

sagikazarmark:
And thank you very much

mortenhauberg:
No problem. And thank you!

gmponos:
Hello,

What I was wondering about this MR is if these changes have any minor impact in performance. I haven't done any performance test on this MR but from what I know array_filter is slower than foreach and `is_null` is slower from `=== null`

Most of the changes I saw on this MR where mainly about making the code more readable. I guess for a library like guzzle performance should be most important rather than readability.

sources: 
http://www.levijackson.net/are-array_-functions-faster-than-loops/
http://stackoverflow.com/a/4662685/4158811

Probably you can find other sources too.

sagikazarmark:
Generally I agree with you. The reason why I accepted this patch is that (correct me if I am wrong) these are only called upon construction time once, so it should not cause any performance hit.

gmponos:
Hmm.. I didn't notice that. Yes you are correct.. It is called only on construction... So yes, most probably the difference will be insignificance.

sagikazarmark:
Actually I was wrong in one of the cases: it's called in every subsequent request. It's premature optimization anyway, so I am reverting the PR.

PR #1613, opened 2016-10-06
hhvm type fixes

natostanco:
like #1366


Nyholm:
Could you provide a tests that shows that these issues are fixed when running at HHVM?

sagikazarmark:
Hm, not sure if we need tests for that, those nulls seems simply incorrect to me as the documentation clearly says string. A rebase would be necessary though.

sagikazarmark:
@natostanco may I ask you why you closed this?

PR #1614, opened 2016-10-07
Follow location on status code 3xx, 201,202

avido:
Instead of only following Location if http-status is 3xx. 
Also following the location if status response in 201 (Created) Or 202 (Accepted)


mtdowling:
I'm not sure this is the expected behavior developers would expect. Correct me if I'm wrong, but 201 created provides a location header that tells you where you can get the created resource. I don't it's meant to instruct clients to automatically follow that location.

avido:
Hmm valid point..
You can close the merge request ;)

PR #1621, opened 2016-10-18
Don't include summary in exception message when body is empty

cnanney:
Exception messages end with `:\n\n` when there is no response body.


sagikazarmark:
Thank you @cnanney

PR #1622, opened 2016-10-19
Quickstart: note that HTTP_PROXY only works on CLI

jorissteyn:
Due to "httpoxy" the environment variable HTTP_PROXY is not trusted on non-cli processes. This should be mentioned in the quickstart guide.


sagikazarmark:
Do you think it would make sense to add some context there? Like https://httpoxy.org/

jorissteyn:
Makes sense. I've added a commit - what do you think?

sagikazarmark:
Awesome, thank you very much.

PR #1626, opened 2016-10-21
Created more client exceptions

gmponos:
I have added some more Exceptions for 4xx http errors. I haven't wrote any Tests. If the changes seems fine to you then I will proceed with the unit tests and update any documentation.

More specific I added the following Exceptions

```
 400 = BadRequestException
 401 = UnauthorizedException
 403 = ForbiddenException
 404 = NotFoundException
 405 = MethodNotAllowedException
 406 = NotAcceptableException
 409 = ConflictException
 410 = GoneException
 429 = TooManyRequestsException
```

I believe they are the most common cases.

### DIFF
--- a/src/Exception/BadRequestException.php
+++ b/src/Exception/BadRequestException.php
@@ -1,0 +1,7 @@
+<?php
+namespace GuzzleHttp\Exception;
+
+/**
+ * Exception when an HTTP error occurs (400 error)
+ */
+class BadRequestException extends ClientException  {}
--- a/src/Exception/ConflictException.php
+++ b/src/Exception/ConflictException.php
@@ -1,0 +1,7 @@
+<?php
+namespace GuzzleHttp\Exception;
+
+/**
+ * Exception when an HTTP error occurs (409 error)
+ */
+class ConflictException extends ClientException  {}
--- a/src/Exception/ForbiddenException.php
+++ b/src/Exception/ForbiddenException.php
@@ -1,0 +1,7 @@
+<?php
+namespace GuzzleHttp\Exception;
+
+/**
+ * Exception when an HTTP error occurs (403 error)
+ */
+class ForbiddenException extends ClientException  {}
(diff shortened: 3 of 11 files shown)


gmponos:
Yes... It does add more maintenance and if it is ever needed to delete/change/refactor a class of them then it would be more easy to have a breaking change and I can see your point.

What I wanted to achieve is instead of having something like this:

``` php
    try {
        $client->get('/something');
    } catch(RequestException $e){
        if ($e->getCode() === 404){
            $this->doSomething();
        } else {
           $this->doSomethingElse();
       }
    }
    return false;
```

I think this it is better structured:
``` php
    try {
        $client->get('/something');
    } catch (NotFoundException $e){
        $this->doSomething();
    } catch (RequestException $e) {
       $this->doSomethingElse();
    }
    return false;
```

I am also curious about some feedback for this PR. So, yes... leave it open. Thank you.

Konafets:
I like this. It gives some meaning to the status codes.

Nyholm:
Thank you for this PR. 
I've discussed this back and forth many times. Though it seams nice there is no real reason for this when **catching** an exception. A third party library that uses Guzzle will not do anything different in the error handling if the response is a 404 or 405 exception. 

If an application author, however, is interested in a specific exception, I see no issue doing:
```php
try {
    $client->get('/something');
} catch(RequestException $e){
    $code = $e->getCode()
    if ($code === 404 || $code === 405){
        $this->checkTheHttpRequestLine();
    } elseif ($code > 410) {
       $this->doSomethingElse();
    } else {  
       $this->other();
    }
}
```

The benefits for this PR are small/none and it adds many more exception classes.

ABM-Dan:
This might be better served by having the various codes available as constants instead of separate paper-thin classes.

sagikazarmark:
Here you are: https://github.com/shrikeh/teapot

sagikazarmark:
Thanks for the contribution @gmponos, but currently I don't see this acceptable for Guzzle, I am sorry.

gmponos:
No problem.

PR #1629, opened 2016-10-22
Replace concat

gmponos:
It's not a big thing but I don't see the reason why there should be a concatenation since there is already an sprintf function. Unless it's for some readability reason.


sagikazarmark:
Thanks @gmponos

PR #1635, opened 2016-10-27
Added missing square bracket

j92:
Added a missing square bracket to the headers section in the request options page.


sagikazarmark:
Thanks @j92

PR #1639, opened 2016-11-05
add http request auth config keys

abacaphiliac:
add http request auth config keys, so that `array_merge` operations do not produce unexpected runtime issues.

i'm using a framework that uses a configuration merge tool, similar in behavior to `array_merge`. when my local overrides are merged into the default configuration, my resulting guzzle auth config contains a duplicate set of numerically-indexed credentials, which causes unexpected behavior in my http client.

my suggested change is backwards-compatible. we can continue to use numerically-indexed auth config, or users can use my new `username`, `password`, and `type` keys that can be easily overridden/merged.

thoughts?


sagikazarmark:
Thanks for the contribution @abacaphiliac 

This indeed looks interesting and would be nice to have as a feature. My concern that supporting both ways could cause a slight confusion in people.

I would consider deprecating the "old way" since - as you pointed out - it does not really play well with configuration management.

Anyway I like your PR, waiting for other voices.

abacaphiliac:
@sagikazarmark thanks for the feedback. i agree that supporting both is ambiguous, and difficult to document. ideally there would only be one way, but switching from numerically indexed to a config hash would definitely be a breaking change. do you have suggestions for deprecation other than comments/docs?

src/Client.php
```diff
@@ -339,7 +339,16 @@ private function applyOptions(RequestInterface $request, array &$options)
...
             $value = $options['auth'];
-            $type = isset($value[2]) ? strtolower($value[2]) : 'basic';
+            $type = 'basic';
+            if (isset($value['type'])) {
+                $type = strtolower($value['type']);
+            } else if (isset($value[2])) {
+                $type = strtolower($value[2]);
+            }
+            $value = [
+                array_key_exists('username', $value) ? $value['username'] : $value[0],
```
sagikazarmark: Wonder if isset would be better (it's faster). Do we allow null values here?

sagikazarmark:
Well, if we decide to go this way then we have to go with your solution at first, make numeric indexed array deprecated and removed support in the next major release.

src/Client.php
abacaphiliac: i'd like to maintain behavioral consistency, and the previous implementation does not restrict null. i think null is acceptable.

abacaphiliac:
@sagikazarmark i took care of your requests. i also squashed the commits, which seems to have upset Github's review tool. i'll keep that in mind for next time haha.

vaclavvanik:
Any news on this?

abacaphiliac:
i have resolved conflicts. this is a pretty old request now, but the added tests still pass.

Nyholm:
Im sorry this has stalled since forever. Im hesitant that this is needed. 

I'll close this PR for now as we haven't merged it in 4 years and there is no indication/activity/request for it. 

Don't take this as a permanent "no, never". It is more me cleaning up really old stuff. Feel free to open an issue with this if you want to. (Or to open the same PR again)

PR #1642, opened 2016-11-06
Deprecate BadResponseException instantiation without a response

### LINKED ISSUE #1631: Should the constructor method allow to create a BadResponseException without a response
At the moment the `BadResponseException` does not have a `__constructor` method. So it uses the parent __constructor which is:

``` php
    public function __construct(
        $message,
        RequestInterface $request,
        \Exception $previous = null,
        array $handlerContext = []
    ) {
    ...
    }
```

The above allows to create a `BadResponseException` with no response. All BadResponseExceptions are for 4xx and 5xx. I guess that you always have a response on these codes.

By allowing the response to be null, the programmer has to check if there is a response in order to be sure.

``` php
try{
    $client->get('/myurl');
} catch (BadResponseException $e) {
     if($e->hasResponse()){
         $e->getResponse()->getStatusCode();
     }
}
```

So what I suggest is that the BadResponseException to have a constructor disallowing someone to create an Exception like this without a response.

``` php
class BadResponseException extends RequestException
{

    public function __construct(
        $message,
        RequestInterface $request,
        ResponseInterface $response,
        \Exception $previous = null,
        array $handlerContext = []
    ) {
        parent::__construct($message, $request, $response, $previous, $handlerContext);
    }
}
```

If you agree I can make the PR.

gmponos:
Closes #1631 
I created the PR anyway :smile: 

**Edit**:
I guess this is a Breaking Change? I don't know if it will be accepted but it feels more correct to me this way.

### DIFF
--- a/src/Exception/BadResponseException.php
+++ b/src/Exception/BadResponseException.php
@@ -1,7 +1,27 @@
 <?php
 namespace GuzzleHttp\Exception;
 
+use Psr\Http\Message\RequestInterface;
+use Psr\Http\Message\ResponseInterface;
+
 /**
  * Exception when an HTTP error occurs (4xx or 5xx error)
  */
-class BadResponseException extends RequestException {}
+class BadResponseException extends RequestException
+{
+    public function __construct(
+        $message,
+        RequestInterface $request,
+        ResponseInterface $response = null,
+        \Exception $previous = null,
+        array $handlerContext = []
+    ) {
+        if (null === $response) {
+            @trigger_error(
+                'Instantiating the ' . __CLASS__ . ' class without a Response is deprecated since version 6.3 and will be removed in 7.0.',
+                E_USER_DEPRECATED
+            );
+        }
+        parent::__construct($message, $request, $response, $previous, $handlerContext);
+    }
+}


sagikazarmark:
Okay, I am willing to accept this patch, but since this is a BC break, we will have to go through the "normal" deprecation process, so please:

- add back a default null value in the constructor
- before invoking the parent constructor, please add a check for response being null and if it is, trigger an error, like this: `@trigger_error('Instantiating the '.__CLASS__.' class is deprecated since version 6.3 and will be removed in 7.0.', E_USER_DEPRECATED);`

gmponos:
Isn't this message better?

    'Instantiating the ' . __CLASS__ . ' class with null Response is deprecated since version 6.3 and will be removed in 7.0.'

sagikazarmark:
> Isn't this message better?

OMG 👊 

What I wanted to say is "without a Response"

gmponos:
Done!

src/Exception/BadResponseException.php
```diff
@@ -1,7 +1,28 @@
...
+use Psr\Http\Message\RequestInterface;
+use Psr\Http\Message\ResponseInterface;
+
 /**
  * Exception when an HTTP error occurs (4xx or 5xx error)
  */
-class BadResponseException extends RequestException {}
+class BadResponseException extends RequestException
+{
+
```
sagikazarmark: Can you please remove this extra line?

gmponos: Done.!

sagikazarmark:
You were fast! If you don't mind, I keep this from merging for a little while. Since it includes deprecation it needs to go into 6.3. I am now going through the open issues and PRs to see if something requires urgent fix and as such a PATCH version for the 6.2.x series, but after that I will merge it.

Thanks for your contribution!

gmponos:
@sagikazarmark yea am awake... 
I think it would be better also to squash the commits for clean history. What do you think?

sagikazarmark:
You can, but if you don't I will do it anyway. Github now supports squashing commits when merging. 😉 Quite useful.

gmponos:
>  If you don't mind, I keep this from merging for a little while.

Ofcourse I don't mind.

> You can, but if you don't I will do it anyway

Did it anyway.

sagikazarmark:
Thanks for your contribution

PR #1648, opened 2016-11-10
added missing `->send()` in docs

OskarStark:
otherwise the request is never executed


sagikazarmark:
This was true for Guzzle 3, in Guzzle 6 it sends the request.

OskarStark:
You are right, I got it afterwards 😄

PR #1649, opened 2016-11-10
Update .gitattributes

Aliance:
Before if I exec `composer require guzzlehttp/guzzle` I would have see this:

So the `.travis.yml` file appears.

This pull request solves it.

I think the appearance of empty directories `build`, `docs` and `tests` also could be fixed.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,6 +1,6 @@
-build/*          export-ignore
-docs/*           export-ignore
-tests/*          export-ignore
+build            export-ignore
+docs             export-ignore
+tests            export-ignore
 .editorconfig    export-ignore
 .gitattributes   export-ignore
 .gitignore       export-ignore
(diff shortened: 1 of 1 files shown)


kamilsk:
> I think the appearance of empty directories build, docs and tests also could be fixed.
Yes, just replace `docs/*` -> `/docs`, etc.

You can check yourself by `git archive` command

Aliance:
Yeap, that's [right](http://stackoverflow.com/questions/21832151/git-archive-export-ignore-ignoring-directories)!

Aliance:
@mtdowling review this request please!

kamilsk:
I see last activity https://github.com/guzzle/guzzle/commits/master and I think it would be better to ask @sagikazarmark.

sagikazarmark:
I will take a look.

sagikazarmark:
Thanks

PR #1658, opened 2016-11-15
fix sample code in literal to work

DQNEO:
I know it is just a sample code and  the parameters `['user', 'pass']` are dummy.

But it is fact that this code does not work as it is.

```
PHP Fatal error:  Uncaught GuzzleHttp\Exception\ClientException: Client error: `GET https://api.github.com/user` resulted in a `401 Unauthorized` response:
{
  "message": "Bad credentials",
  "documentation_url": "https://developer.github.com/v3"
}
```

this PR fix the problem.

I think it is very important to show codes which work fine as it is, especially for beginners.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,15 +19,13 @@ trivial to integrate with web services.
 
 ```php
 $client = new \GuzzleHttp\Client();
-$res = $client->request('GET', 'https://api.github.com/user', [
-    'auth' => ['user', 'pass']
-]);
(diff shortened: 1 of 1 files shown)


sagikazarmark:
Wow, thanks @DQNEO. Indeed getting started codes should preferably be functional ones. I am just merging this as a quick fix skipping the review period.

Thanks.

PR #1659, opened 2016-11-15
Support force_ip_resolve in StreamHandler

c960657:
The option `force_ip_resolve` that was added in #1608 is only supported for `CurlHandler`.

This patch adds support for `force_ip_resolve` to `StreamHandler`.


src/Handler/StreamHandler.php
```diff
@@ -286,7 +288,7 @@ private function createStream(RequestInterface $request, array $options)
             foreach ($options as $key => $value) {
                 $method = "add_{$key}";
                 if (isset($methods[$method])) {
-                    $this->{$method}($request, $context, $value, $params);
+                    $this->{$method}($request, $context, $value, $params, $uri);
```
sagikazarmark: Is it really necessary to add the URI as a parameter? It's accessible from the request, isn't it?

src/Handler/StreamHandler.php
```diff
@@ -277,6 +278,7 @@ private function createStream(RequestInterface $request, array $options)
 
         $params = [];
         $context = $this->getDefaultContext($request, $options);
+        $uri = $request->getUri()->withFragment('');
```
sagikazarmark: Why is this better than the original code?

src/Handler/StreamHandler.php
```diff
@@ -350,7 +352,7 @@ private function getDefaultContext(RequestInterface $request)
         return $context;
     }
 
-    private function add_proxy(RequestInterface $request, &$options, $value, &$params)
+    private function add_proxy(RequestInterface $request, &$options, $value, &$params, UriInterface &$uri)
```
sagikazarmark: Objects are passed by reference

src/Handler/StreamHandler.php
sagikazarmark: In fact it's not a good idea IMO as the reference for this object will not be the same as you expected.

src/Handler/StreamHandler.php
```diff
@@ -428,7 +430,24 @@ function ($code, $a, $b, $c, $transferred, $total) use ($value) {
...
 
-    private function add_debug(RequestInterface $request, &$options, $value, &$params)
+    private function add_force_ip_resolve(RequestInterface $request, &$options, $value, &$params, UriInterface &$uri)
+    {
+        if ('v4' === $value) {
+            $records = dns_get_record($uri->getHost(), DNS_A);
+            if (!isset($records[0]['ip'])) {
+                throw new ConnectException("Could not resolve IPv4 address for host '{$uri->getHost()}'", $request);
+            }
+            $uri = $uri->withHost($records[0]['ip']);
```
sagikazarmark: Sadly this is also not going to work as you expect. HTTP Messages are immutable, which means if you "modify" one them you will get a new object. The new URI will not be present in the request object. What you can do is saving the resolved IP address in the options array and passing it to the URI, then the URI to the request somewhere in the above code.

src/Handler/StreamHandler.php
c960657: True, but URIs are immutable, so I have to change the value by reference.

src/Handler/StreamHandler.php
c960657: I'm not sure what you mean. I am not modifying the HTTP message here, only $uri that is passed by reference.

src/Handler/StreamHandler.php
c960657: It is, but I chose to preserve the original URI in the request in case other callbacks, middleware etc. need it. The modified $url is not the actual URL being request but a version where the hostname has been replaced with an IP address, i.e. it lacks information about HTTP virtual host. This version is only used to "trick" fopen() into binding to the proper IP.

src/Handler/StreamHandler.php
c960657: See my comment below on why I created a separate `$uri` variable outside the callback.

c960657:
If you have IPv6 connectivity, you can test the code using this script:
```
$c = new \GuzzleHttp\Client();
print (string) $c->get('http://icanhazip.com/s', ['force_ip_resolve' => 'v4'])->getBody();
print (string) $c->get('http://icanhazip.com/s', ['force_ip_resolve' => 'v6'])->getBody();
```

src/Handler/StreamHandler.php
sagikazarmark: Ah, okay, you are totally right. For some reasons I thought you are trying to set the host in the original request. Even then, adding the URI to all methods bellow seems to be too much. How about a `resolveUri` or `resolveAddress` or something similar method? It would make things easier IMO.

c960657: Done.

src/Handler/StreamHandler.php
```diff
@@ -301,21 +302,44 @@ private function createStream(RequestInterface $request, array $options)
...
         );
     }
 
+    private function resolveUri(UriInterface $uri, array $options)
+    {
+        if (isset($options['force_ip_resolve'])) {
+            if ('v4' === $options['force_ip_resolve']) {
+                $records = dns_get_record($uri->getHost(), DNS_A);
+                if (!isset($records[0]['ip'])) {
+                    throw new ConnectException("Could not resolve IPv4 address for host '{$uri->getHost()}'", $request);
```
sagikazarmark: Can you please use sprintf instead? I am not really fan of this kind of resolution.

src/Handler/StreamHandler.php
```diff
@@ -301,21 +302,44 @@ private function createStream(RequestInterface $request, array $options)
...
-                $resource = fopen((string) $request->getUri()->withFragment(''), 'r', null, $context);
+            function () use ($uri, &$http_response_header, $context) {
+                $resource = fopen((string) $uri, 'r', null, $context);
                 $this->lastHeaders = $http_response_header;
                 return $resource;
             }
         );
     }
 
+    private function resolveUri(UriInterface $uri, array $options)
```
sagikazarmark: Maybe call it resolveIp or resolveAddress? Since that's what we do here, isn't it?

c960657: Done. I chose the name `resolveHost` (we are resolving the hostname to an IP address).

src/Handler/StreamHandler.php
c960657: Done

sagikazarmark:
Thanks for your work and patience. 6.3 is on it's way

PR #1662, opened 2016-11-17
Typo

dle-sweethearts:
base_url actually is called base_uri

### DIFF
--- a/docs/clients.rst
+++ b/docs/clients.rst
@@ -14,23 +14,23 @@ Creating a client
 The constructor of a client accepts an associative array of configuration
 options.
 
-base_url
+base_uri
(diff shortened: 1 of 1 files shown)


Nyholm:
Thank you for this contribution. But Im afraid you are incorrect. 
It is called `base_url` in 5.3. See https://github.com/guzzle/guzzle/blob/5.3/src/Client.php#L252

It is in 6.0 it is called `base_uri`. See https://github.com/guzzle/guzzle/blob/6.2.2/src/Client.php#L69

dle-sweethearts:
I was using the wrong Documentation the whole time.. Sorry 😣

Nyholm:
No worries. I really appreciate your contribution.

PR #1675, opened 2016-12-05
5.3: be compatible with PHP 7

pschultz:


src/Client.php
```diff
@@ -182,6 +182,13 @@ public function send(RequestInterface $request)
                 return new FutureResponse(new RejectedPromise($e));
             }
             throw RequestException::wrapException($trans->request, $e);
+        } catch (\TypeError $error) {
```
sagikazarmark: Why just `TypeError` and not `Error` or `Throwable`. Haven't looked to deep, so you might have better insights ATM.

src/Utils.php
```diff
@@ -131,6 +131,10 @@ public static function uriTemplate($template, array $variables)
      */
     public static function jsonDecode($json, $assoc = false, $depth = 512, $options = 0)
     {
+        if ($json === '' || $json === null) {
```
sagikazarmark: Just for curiosity: this has nothing to do with PHP 7 compatibility, right? AFAIK json_decode would just return null anyway, so this might be a performance gain. I guess HHVM would fail if it's null, do it's compatibility thing there.

pschultz: Both PHP 5 and 7 return null, but in PHP 7 a subsequent json_last_error returns JSON_ERROR_SYNTAX, where 5 returns JSON_ERROR_NONE. See https://3v4l.org/FAO0Z.

src/Client.php
pschultz: I don't recall. Presumably I just ran our test suite and added a catch block for our specific failure. I don't see an immediate reason to not catch Throwable, so I'll update the PR.

pschultz: I did it to satisfy [ClientTest::testEnsuresResponseIsPresentAfterSending](https://github.com/guzzle/guzzle/blob/3b9826a2/tests/ClientTest.php#L392), which failed with

> 1) GuzzleHttp\Tests\ClientTest::testEnsuresResponseIsPresentAfterSending
Failed asserting that exception of type "TypeError" matches expected exception "\GuzzleHttp\Exception\RequestException".

On second thought that is probably a flawed test. 

The PHP docs state that "Error is the base class for all internal PHP errors." The subclasses are DivisionByZeroError, AssertionError, ParseError, and TypeError. All of those indicate incorrect code, not exceptional runtime behaviour (and are certainly not RequestExceptions). Converting them to Exceptions changes the semantics. Error and Exception are in different type hierarchies so they can be handled differently.

So even though it is wrong, wrapping TypeError is required to stay backwards compatible. Wrapping all Errors would be more wrong.

sagikazarmark:
Thanks. I will release this as part of 5.3.2 soon.

davidkmenta:
@sagikazarmark could you please release (tag) the version 5.3.2?

sagikazarmark:
@davidkmenta Done

Issue #1683, opened 2016-12-13
Add support for CURLOPT_CAPATH


notxarb:
`CaBundle::getSystemCaRootBundlePath()` returns either a path to the bundle or the to a directory containing the certificates. This is assuming that what is returned is a bundled certificate file. cURL doesn't like it when CURLOPT_CAINFO is a folder, and it errors. CURLOPT_CAPATH should be set instead.

sagikazarmark:
Fixed in #1684

sagikazarmark:
Released in [6.3.0](https://github.com/guzzle/guzzle/releases/tag/6.3.0)

PR #1684, opened 2016-12-13
Bugfix/curlopt capath

notxarb:
This should resolve issue #1683


geradt:
When will this fix be committed and merged with master (released)?

I am having an issue right now that this solution would resolve.

sagikazarmark:
If you need an immediate solution you can copy the propsed factory to your code base and use it in your curl handler.

This PR will be reviewd shortly.

geradt:
Hi Just checking in on this fix again. Any chance it can go in the next push? Thank you again in advance.

sagikazarmark:
@geradt  i will try to make it in the next (6.3) release, but I would like to check if the file/dir checks are correct first.

geradt:
Thank you! The sooner you can get this in the better. I appreciate it. Do you have any kind of target date in mind? I see the 6.3 milestone but it doesn't have a due date.

Thanks!

sagikazarmark:
I would like to get it done asap, but time is a factor I have to take into account sadly.

sagikazarmark:
Thanks for your patience and work. 6.3 is not so far.

geradt:
Thanks sagikazarmark!! Looking forward to it!

PR #1686, opened 2016-12-15
Fix count being called on null

kelunik:
PHP 7.2 requires arguments to count being an array or an instance of Countable.

This property when not initialized, is neither of these, it's null.


sagikazarmark:
Thanks

PR #1687, opened 2016-12-15
Make CurlMultiHandler extandable

kluevandrew:
I need to override tick() method in reactphp app, and don't want to copy all class? please use protected access for properties and methods


sagikazarmark:
Sorry, but I am not sure I like this. What are your concerns about copying the class?

sagikazarmark:
Sorry, but this is not something I would like to see in Guzzle. Please copy the class if you really need it.

PR #1704, opened 2016-12-27
Update request-options.rst

baijunyao:
$request->getHeader('Content-Type'); is array;
$request->getBody() ; is Object;  
$request->getBody()->getContents(); is string;  //"{"foo":"bar"}"

### DIFF
--- a/docs/request-options.rst
+++ b/docs/request-options.rst
@@ -557,9 +557,10 @@ over the wire.
     $clientHandler = $client->getConfig('handler');
     // Create a middleware that echoes parts of the request.
     $tapMiddleware = Middleware::tap(function ($request) {
-        echo $request->getHeader('Content-Type');
+        $contentType = $request->getHeader('Content-Type');
+        echo $contentType[0];
         // application/json
-        echo $request->getBody();
+        echo $request->getBody()->getContents();
         // {"foo":"bar"}
     });
 


docs/request-options.rst
```diff
@@ -557,9 +557,10 @@ over the wire.
     $clientHandler = $client->getConfig('handler');
     // Create a middleware that echoes parts of the request.
     $tapMiddleware = Middleware::tap(function ($request) {
-        echo $request->getHeader('Content-Type');
+        $contentType = $request->getHeader('Content-Type');
+        echo $contentType[0];
         // application/json
-        echo $request->getBody();
+        echo $request->getBody()->getContents();
```
Nyholm: `getContents` will read the "rest of the body". Which is not always what you want. 

`echo $request->getBody();` will work because `getBody` returns a stream that have a `__toString` function.

docs/request-options.rst
```diff
@@ -557,9 +557,10 @@ over the wire.
     $clientHandler = $client->getConfig('handler');
     // Create a middleware that echoes parts of the request.
     $tapMiddleware = Middleware::tap(function ($request) {
-        echo $request->getHeader('Content-Type');
+        $contentType = $request->getHeader('Content-Type');
+        echo $contentType[0];
```
Nyholm: Remove this. It will be an error if there is no Content-Type

docs/request-options.rst
```diff
@@ -557,9 +557,10 @@ over the wire.
     $clientHandler = $client->getConfig('handler');
     // Create a middleware that echoes parts of the request.
     $tapMiddleware = Middleware::tap(function ($request) {
-        echo $request->getHeader('Content-Type');
+        $contentType = $request->getHeader('Content-Type');
```
Nyholm: Change this line to:

`echo $request->getHeaderLine('Content-Type');`

baijunyao:
Oh, I see, Thank you.

Nyholm:
Why did you close this? There is an issue that needs to be fixed here: 

```php
// This line:
echo $request->getHeader('Content-Type');

// Should be replaced by:
echo $request->getHeaderLine('Content-Type');

```

Issue #1710, opened 2016-12-31
Redirect status code history?


mallardduck:
_I've searched and didn't seem to find anything on this, if I'm missing something let me know._

I noticed that #1183 added a feature for redirect history, but it only tracks the URL locations not the HTTP status codes. I'm wondering if it would be possible to add history for status codes as well? Would a PR for a solution be accepted if submitted?

I was thinking of a few ways to try and pull it off, but the easiest seems like adding it on to the current RedirectMiddleware provided with Guzzle. It could just be added an additional header similar to the current one being added; maybe something like:

```
    const STATUS_HISTORY_HEADER = 'X-Guzzle-Redirect-Status-History';


    private function withTracking(PromiseInterface $promise, $uri)
    {
        return $promise->then(
            function (ResponseInterface $response) use ($uri) {
                // Note that we are pushing to the front of the list as this
                // would be an earlier response than what is currently present
                // in the history header.
                $historyHeader = $response->getHeader(self::HISTORY_HEADER);
                $statusHeader = $response->getHeader(self::STATUS_HISTORY_HEADER);
                array_unshift($historyHeader, $uri);
                array_unshift($statusHeader, $response->getStatusCode());
                return $response->withHeader(self::HISTORY_HEADER, $historyHeader)->withHeader(self::STATUS_HISTORY_HEADER, $statusHeader);
            }
        );
    }

```

mallardduck:
Update: Actually it looks like this method wouldn't end up getting proper status codes. I've tested it out and it seems like you end up with all 200's.

sagikazarmark:
That sounds a little bit odd. All those status codes should be 3xx.

mallardduck:
I think I figured out my issue; I seem to have it working now that I'm passing the status into the withTracking methoid like the uri. I'll submit a PR soon!

sagikazarmark:
What's the point in preserving HTTP Status codes? What's the added information (besides it's a redirection)?

mallardduck:
I see the main benefit being that by tracking the status codes we can use Guzzle in a similar way to using Curl. Using `curl -IL` can get me useful information on how a request gets from point A to point B. Redirect paths are not always as simple as A->B, when it gets complicated and looks like A->C->E->D->B then you benefit from having status codes.

[...]

PR #1711, opened 2016-12-31
Track redirect HTTP status codes

mallardduck:
Using a similar technique to the current Redirect History feature this PR adds support for tracking the HTTP status codes in a new: 'X-Guzzle-Redirect-Status-History' header.

This PR was created in response to #1710

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -131,17 +131,16 @@ setting the ``expect`` request option to ``false``:
     $client = new GuzzleHttp\Client(['expect' => false]);
 
 How can I track a redirected requests HTTP codes?
-============================================
+=================================================
 
-You can enable the `track_redirects` option which will enable Guzzle to track
-any redirected URI and status code. Each redirected URI will be stored in the
-``X-Guzzle-Redirect-History`` header and each Response Code will be stored in
+You can enable tracking of redirected URLs and the status codes via the
+`track_redirects` option. Each redirected URI will be stored in the
+``X-Guzzle-Redirect-History`` header and each status code will be stored in
 the ``X-Guzzle-Redirect-Status-History`` header.
 
-The ``X-Guzzle-Redirect-History`` header will exclude the initial request's URI
-and the ``X-Guzzle-Redirect-Status-History`` header will exclude the final
-response code. With this in mind you should easily be able to track a requests
-full redirect path.
+The URL history header will exclude the initial request's URI and the
+status code history header will exclude the final status code. With this
+in mind you should easily be able to track a request's full redirect path.
 
 For example, let's say you need to track redirects and provide both results
 together in a single report:
@@ -167,13 +166,11 @@ together in a single report:
     $redirectUriHistory = $response->getHeader('X-Guzzle-Redirect-History'); // retrieve Redirects URI history
     $redirectCodeHistory = $response->getHeader('X-Guzzle-Redirect-Status-History'); // retrieve Redirects HTTP Status history
 
-    // Add the initial URI requested to the URI history
-    $reversedUriHistory = array_reverse($redirectUriHistory); // First we reverse the array items
-    array_push($reversedUriHistory, $initialRequest); // then we add the initial URL on the end
-    $redirectUriHistory = array_reverse($reversedUriHistory); // Overwrite original variable with new list
+    // Add the initial URI requested to the (beginning of) URI history
+    array_unshift($reversedUriHistory, $initialRequest); // then we add the initial URL on the end
 
     // Add the final HTTP status code to the HTTP response history
-    array_push($redirectCodeHistory, $response->getStatusCode()); // Add final response status to code array
+    array_push($redirectCodeHistory, $response->getStatusCode()); // Add final status code to code array
 
     // (Optional) Combine the items of each array into a single result set
     $fullRedirectReport = [];
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -129,3 +129,54 @@ setting the ``expect`` request option to ``false``:
 
     // Disable the expect header on all client requests
     $client = new GuzzleHttp\Client(['expect' => false]);
+
+How can I track a redirected requests HTTP codes?
+============================================
+
+You can enable the `track_redirects` option which will enable Guzzle to track
+any redirected URI and status code. Each redirected URI will be stored in the
+``X-Guzzle-Redirect-History`` header and each Response Code will be stored in
+the ``X-Guzzle-Redirect-Status-History`` header.
+
+The ``X-Guzzle-Redirect-History`` header will exclude the initial request's URI
+and the ``X-Guzzle-Redirect-Status-History`` header will exclude the final
+response code. With this in mind you should easily be able to track a requests
+full redirect path.
+
+For example, let's say you need to track redirects and provide both results
+together in a single report:
+
+.. code-block:: php
+
+    // First you configure Guzzle with redirect tracking and make a request
+    $client = new Client([
+                RequestOptions::CONNECT_TIMEOUT => 10,
+                RequestOptions::TIMEOUT => 10,
+                RequestOptions::COOKIES => true,
+                RequestOptions::ALLOW_REDIRECTS => [
+                    'max'             => 10,        // allow at most 10 redirects.
+                    'strict'          => true,      // use "strict" RFC compliant redirects.
+                    'referer'         => true,      // add a Referer header
+                    'track_redirects' => true
+                ]
+            ]);
+    $initialRequest = '/redirect/3'; // Store the request URI for later use
+    $response = $client->request('GET', $initialRequest); // Make your request
+
+    // Retrieve both Redirect History headers
+    $redirectUriHistory = $response->getHeader('X-Guzzle-Redirect-History'); // retrieve Redirects URI history
+    $redirectCodeHistory = $response->getHeader('X-Guzzle-Redirect-Status-History'); // retrieve Redirects HTTP Status history
+
+    // Add the initial URI requested to the URI history
+    $reversedUriHistory = array_reverse($redirectUriHistory); // First we reverse the array items
+    array_push($reversedUriHistory, $initialRequest); // then we add the initial URL on the end
+    $redirectUriHistory = array_reverse($reversedUriHistory); // Overwrite original variable with new list
+
+    // Add the final HTTP status code to the HTTP response history
+    array_push($redirectCodeHistory, $response->getStatusCode()); // Add final response status to code array
+
+    // (Optional) Combine the items of each array into a single result set
+    $fullRedirectReport = [];
+    foreach ($redirectUriHistory as $key => $value) {
+        $fullRedirectReport[$key] = ['location' => $value, 'code' => $redirectCodeHistory[$key]];
+    }
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -130,17 +130,16 @@ setting the ``expect`` request option to ``false``:
     // Disable the expect header on all client requests
     $client = new GuzzleHttp\Client(['expect' => false]);
 
-How can I track a redirected requests HTTP codes?
-=================================================
+How can I track a redirected requests?
+======================================
 
-You can enable tracking of redirected URLs and the status codes via the
-`track_redirects` option. Each redirected URI will be stored in the
-``X-Guzzle-Redirect-History`` header and each status code will be stored in
-the ``X-Guzzle-Redirect-Status-History`` header.
+You can enable tracking of redirected URIs and status codes via the
+`track_redirects` option. Each redirected URI and status code will be stored in the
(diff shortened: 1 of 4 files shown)


src/RedirectMiddleware.php
```diff
@@ -108,23 +110,26 @@ public function checkRedirect(
         if (!empty($options['allow_redirects']['track_redirects'])) {
             return $this->withTracking(
                 $promise,
-                (string) $nextRequest->getUri()
+                (string) $nextRequest->getUri(),
+                (int) $response->getStatusCode()
```
sagikazarmark: Is there any added value in casting it to int? It will become string anyway. If ou just followed the pattern: the URI needs to be casted so that the URI object becomes string.

src/RedirectMiddleware.php
```diff
@@ -108,23 +110,26 @@ public function checkRedirect(
...
                 // would be an earlier response than what is currently present
                 // in the history header.
-                $header = $response->getHeader(self::HISTORY_HEADER);
-                array_unshift($header, $uri);
-                return $response->withHeader(self::HISTORY_HEADER, $header);
+                $historyHeader = $response->getHeader(self::HISTORY_HEADER);
+                $statusHeader = $response->getHeader(self::STATUS_HISTORY_HEADER);
+                array_unshift($historyHeader, $uri);
+                array_unshift($statusHeader, $statusCode);
+                return $response->withHeader(self::HISTORY_HEADER, $historyHeader)->withHeader(self::STATUS_HISTORY_HEADER, $statusHeader);
```
sagikazarmark: Can you split this line to multiple ones?

sagikazarmark:
Okay, you bought me.

Can you please add some tests?

mallardduck:
Sure thing, I can totally add tests!
I'll work up some tests tonight and get an updated PR sometime over the weekend.

src/RedirectMiddleware.php
```diff
@@ -108,23 +110,27 @@ public function checkRedirect(
         if (!empty($options['allow_redirects']['track_redirects'])) {
             return $this->withTracking(
                 $promise,
-                (string) $nextRequest->getUri()
+                (string) $nextRequest->getUri(),
+                (string) $response->getStatusCode()
```
sagikazarmark: I am failing to see why we need to do any casting here. As said `getUri` returns an object which MUST be casted to string, but the status code is scalar.

sagikazarmark:
I have one last comment, apart from that good to go.

src/RedirectMiddleware.php
mallardduck: Sorry about that, I guess I was a little confused initially. Totally get your point now tho.

Update PR incoming....

sagikazarmark:
Okay, the code looks good to me, there is only one final thought: Right now we store status code, and next URL pairs, which means the status code belongs to the previous request, which returned the appropriate new location.

Not sure if it can be confusing or we should document it (since the middlewares are not documented, somewhere in the code).

WDYT?

mallardduck:
That's an excellent point!

As you've suggested I think adding comments in the code could be a good option. Alternatively I was thinking this could be added to the FAQ doc?

I'm not sure how common of a question "Why can't I track redirects with Guzzle?" actually is, but maybe this information could fit there? If not that exact question maybe one similar to it?

I also just realized that I should definitely update the [request-options](https://github.com/guzzle/guzzle/blob/14fa40e26857c8f406d7a0c7d577705085fb6950/docs/request-options.rst) file as it contains info on the currently provided header.

----

I'll update that file and add some comments to the code; I may even include my example of how I track the full redirect path. From there we can evaluate if that's enough information!

mallardduck:
I've updated the relevant request-options section in the documents, please take a look and let me know your thoughts!

I've also included an example in the FAQ file that's useful for users looking to consume both headers as a single item. This can be used to merge HTTP status codes with their respective request URIs.

docs/faq.rst
```diff
@@ -129,3 +129,54 @@ setting the ``expect`` request option to ``false``:
 
     // Disable the expect header on all client requests
     $client = new GuzzleHttp\Client(['expect' => false]);
+
+How can I track a redirected requests HTTP codes?
+============================================
```
sagikazarmark: Should be as long as the title

docs/faq.rst
```diff
@@ -129,3 +129,54 @@ setting the ``expect`` request option to ``false``:
...
+
+You can enable the `track_redirects` option which will enable Guzzle to track
```
sagikazarmark: This first sentence with double "enabled" sounds a little bit weird...

docs/faq.rst
```diff
@@ -129,3 +129,54 @@ setting the ``expect`` request option to ``false``:
...
+any redirected URI and status code. Each redirected URI will be stored in the
+``X-Guzzle-Redirect-History`` header and each Response Code will be stored in
```
sagikazarmark: Let's say status code instead of Response Code

docs/faq.rst
sagikazarmark: Maybe say "You can enable tracking the redirected URL and the status code via the ... option"

docs/faq.rst
```diff
@@ -129,3 +129,54 @@ setting the ``expect`` request option to ``false``:
...
+the ``X-Guzzle-Redirect-Status-History`` header.
+
+The ``X-Guzzle-Redirect-History`` header will exclude the initial request's URI
```
sagikazarmark: Instead of the header names we could say "the URL history" and "the status code history". Just to avoid duplication. But I am open to other suggestions.

docs/faq.rst
```diff
@@ -129,3 +129,54 @@ setting the ``expect`` request option to ``false``:
...
+and the ``X-Guzzle-Redirect-Status-History`` header will exclude the final
+response code. With this in mind you should easily be able to track a requests
```
sagikazarmark: status code

sagikazarmark: track a request's

docs/faq.rst
```diff
@@ -129,3 +129,54 @@ setting the ``expect`` request option to ``false``:
...
+            ]);
+    $initialRequest = '/redirect/3'; // Store the request URI for later use
+    $response = $client->request('GET', $initialRequest); // Make your request
+
+    // Retrieve both Redirect History headers
+    $redirectUriHistory = $response->getHeader('X-Guzzle-Redirect-History'); // retrieve Redirects URI history
+    $redirectCodeHistory = $response->getHeader('X-Guzzle-Redirect-Status-History'); // retrieve Redirects HTTP Status history
+
+    // Add the initial URI requested to the URI history
+    $reversedUriHistory = array_reverse($redirectUriHistory); // First we reverse the array items
```
sagikazarmark: How about using array_unshift instead?

docs/faq.rst
```diff
@@ -129,3 +129,54 @@ setting the ``expect`` request option to ``false``:
...
+    array_push($reversedUriHistory, $initialRequest); // then we add the initial URL on the end
+    $redirectUriHistory = array_reverse($reversedUriHistory); // Overwrite original variable with new list
+
+    // Add the final HTTP status code to the HTTP response history
+    array_push($redirectCodeHistory, $response->getStatusCode()); // Add final response status to code array
+
+    // (Optional) Combine the items of each array into a single result set
+    $fullRedirectReport = [];
+    foreach ($redirectUriHistory as $key => $value) {
+        $fullRedirectReport[$key] = ['location' => $value, 'code' => $redirectCodeHistory[$key]];
```
sagikazarmark: What should be the first item here? IMO it should be the first URL if any, but currently it's the last.

docs/request-options.rst
```diff
@@ -70,9 +70,15 @@ pairs:
   is encountered. The callable is invoked with the original request and the
   redirect response that was received. Any return value from the on_redirect
   function is ignored.
-- track_redirects: (bool) When set to ``true``, each redirected URI encountered
-  will be tracked in the ``X-Guzzle-Redirect-History`` header in the order in
-  which the redirects were encountered.
+- track_redirects: (bool) When set to ``true``, each redirected URI and Response
```
sagikazarmark: status code

docs/request-options.rst
```diff
@@ -70,9 +70,15 @@ pairs:
...
+  Code encountered will be tracked in the ``X-Guzzle-Redirect-History`` and 
+  ``X-Guzzle-Redirect-Status-History`` headers respectively. All URIs and
+  Response Codes will be stored in the order in which the redirects were
```
sagikazarmark: status code

docs/request-options.rst
```diff
@@ -70,9 +70,15 @@ pairs:
...
+  encountered.
+
+  Note: When tracking redirects the ``X-Guzzle-Redirect-History`` header will
+  exclude the initial request's URI and the ``X-Guzzle-Redirect-Status-History``
+  header will exclude the final response code.
```
sagikazarmark: status code

docs/faq.rst
mallardduck: This sounds good to me and seems to make it easier to read.

docs/faq.rst
mallardduck: xD yep should have thought about that.

lately I've been using collections a lot and converted that logic into this. (there's no unshift for the collecitons i've been using)

docs/faq.rst
mallardduck: I guess I'm a little unsure on this to be honest. I read the comments in the middleware about using array_unshift there due to ordering of codes....however when I execute this code I am provided the same order of redirects as I see in curl.

Using the faq code with this patch I am provided with:

```
[{"location":"https:\/\/grandmascookieblog.com\/cloud.php","code":302},{"location":"http:\/\/www.liquidweb.com\/cloudsites\/","code":301},{"location":"https:\/\/www.liquidweb.com\/cloudsites\/","code":200}]
```
You can test with this URL: https://grandmascookieblog.com/cloud.php

Via curl I can see:

```
[ guzzle (track_redirect_codes)]$ curl -IL https://grandmascookieblog.com/cloud.php
HTTP/1.1 302 Found
Date: Thu, 12 Jan 2017 23:59:52 GMT
Server: Apache
Location: http://www.liquidweb.com/cloudsites/
Cache-Control: max-age=0
Expires: Thu, 12 Jan 2017 23:59:52 GMT
X-Cache: MISS from grandmascookieblog.com
Content-Type: text/html; charset=UTF-8

HTTP/1.1 301 Moved Permanently
Server: nginx/1.10.2
Date: Thu, 12 Jan 2017 23:59:53 GMT
Content-Type: text/html
Content-Length: 185
Connection: keep-alive
Location: https://www.liquidweb.com/cloudsites/
X-Frame-Options: SAMEORIGIN

HTTP/1.1 200 OK
Server: nginx/1.10.2
Date: Thu, 12 Jan 2017 23:59:53 GMT
Content-Type: text/html; charset=ISO-8859-1
Content-Length: 38913
Connection: keep-alive
Set-Cookie: production_www_sessid=1223adaee9e822dff41d9be79dd793a2; path=/
X-ua-compatible: IE=edge,chrome=1
X-Frame-Options: SAMEORIGIN
```

docs/faq.rst
mallardduck: Didn't realize that, I'll get that updated!

docs/faq.rst
```diff
@@ -129,3 +129,51 @@ setting the ``expect`` request option to ``false``:
...
+            ]);
+    $initialRequest = '/redirect/3'; // Store the request URI for later use
+    $response = $client->request('GET', $initialRequest); // Make your request
+
+    // Retrieve both Redirect History headers
+    $redirectUriHistory = $response->getHeader('X-Guzzle-Redirect-History'); // retrieve Redirects URI history
+    $redirectCodeHistory = $response->getHeader('X-Guzzle-Redirect-Status-History'); // retrieve Redirects HTTP Status history
+
+    // Add the initial URI requested to the (beginning of) URI history
+    array_unshift($reversedUriHistory, $initialRequest); // then we add the initial URL on the end
```
sagikazarmark: Is `$reversedUriHistory` correct here?

docs/faq.rst
sagikazarmark: Hm, looking at the code again, it should be fine now.

sagikazarmark:
@mallardduck I would like to release this in 6.3 in a few weeks. Can you check the remaining comments? Thanks

mallardduck:
@sagikazarmark absolutely! I was hoping to get this done today  at work however I ran out of time. I'll work on this over the weekend and shoot an updated PR your way.

I really appreciate the followup on this and wanted to let you know it's been awesome to collaborate and give back.

sagikazarmark:
@mallardduck thanks, I think this is ready. I changed some small wordings, also changed the FAQ example (removed some request options) to make it simpler. I think those weren't important for this feature anyway. If they were, please let me know, so we can put them back.

PR #1713, opened 2017-01-03
Require react/promise:2.2 or above

Nyholm:
We use react/promise but we have never specified it in composer.json. We do currently not support react/promise:2.0 or react/promise:2.1

This will fix https://github.com/guzzle/RingPHP/issues/39

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,8 @@
     ],
     "require": {
         "php": ">=5.4.0",
-        "guzzlehttp/ringphp": "^1.1"
+        "guzzlehttp/ringphp": "^1.1",
+        "react/promise": "^2.2"
     },
     "require-dev": {
         "ext-curl": "*",


sagikazarmark:
Shouldn't it be in RingPHP?

Nyholm:
No, It should be react/promise. react/promise has been an indirect dependency via guzzlehttp/ringphp.

guzzlehttp/ringphp supports react/promise:~2.0 but this package need ^2.2. 

Situations like this is why Im not a big fan on indirect dependencies. I think it is better to specify this dependency here and with proper version constraints. 

The alternative solution would be to bump to react/promise:^2.2 on guzzlehttp/ringphp (but there is no reson) and then require guzzlehttp/ringphp:new here.

sagikazarmark:
What does "indirect" dependency means? Isn't it rather a conflict problem?

Nyholm:
With indirect i mean

A -> B -> C

If A is using C there is a dependency. If A do not specify the dependency "since it is already loaded by B" I call C an **indirect** dependency of A. Maybe not the best choice of words but I hope you understand.

jayS-de:
The Pool class has a direct coupling to the React/Promise library. See [Pool.php#L12-L15](https://github.com/guzzle/guzzle/blob/955957d7/src/Pool.php#L12-L15) But Guzzle doesn't list this as required dependency as itself and relies on the fact that RingPHP requires the promises. The refactoring in the use of a deprecated function (see https://github.com/guzzle/guzzle/commit/955957d) now uses an implementation detail of react/promise v2.2 which may not be correctly resolved by the constraints defined in the composer.json as it is shown in #1714

sagikazarmark:
Alright, I think enough said.

Nyholm:
Thank you for merging.

PR #1715, opened 2017-01-04
Update Client.php

jaylathia:
Adding preg_replace to replace square braces of array indices, generated by http_build_query function.

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -357,6 +357,7 @@ class Client implements ClientInterface
             $value = $options['query'];
             if (is_array($value)) {
                 $value = http_build_query($value, null, '&', PHP_QUERY_RFC3986);
+                $value = preg_replace('/%5B(?:[0-9]|[1-9][0-9]+)%5D=/', '=', $value);
             }
             if (!is_string($value)) {
                 throw new \InvalidArgumentException('query must be a string or array');


Nyholm:
Thank you for this PR. I do not understand why you would like to replace them. Can you please explain your scenario. 

Aslo, if this PR ever going to get merged there has to be tests.

jaylathia:
The scenario being converting 
`?variable[0]=value1&variable[1]=value2&variable[2]=value3`
to
`?variable=value1&variable=value2&variable=value3`

Nyholm:
Yes, I can see that your patch converts the query like that.

`?variable=value1&variable=value2&variable=value3` is not a valid format for a query string. That is why I asked why you would like this change.

jayS-de:
It is a valid query string. But it depends on the library/language you are using if this could be correctly parsed.

Most PHP applications expect the array notation. In other languages its very common to combine parameters of the same name. The URI implementation is able to parse this correctly.

sagikazarmark:
Putting this onto my close in a few days list.

mtdowling:
This is a breaking change. You can provide the query option as a string to use any form of serialization you'd like.

sagikazarmark:
Closing as:

- breaking change
- alternative solution proposed
- invalid

PR #1718, opened 2017-01-06
fix when curl set time less than one second causes libcurl to timeout…

tom916:
http://php.net/manual/en/function.curl-setopt.php

>If you want cURL to timeout in less than one second, you can use CURLOPT_TIMEOUT_MS, although >there is a bug/"feature"  on "Unix-like systems" that causes libcurl to timeout immediately if the value is >< 1000 ms with the error "cURL Error (28): Timeout was reached".  The explanation for this behavior is:

### DIFF
--- a/src/Handler/CurlFactory.php
+++ b/src/Handler/CurlFactory.php
@@ -370,9 +370,9 @@ class CurlFactory implements CurlFactoryInterface
             $conf[CURLOPT_FILE] = fopen('php://temp', 'w+');
             $easy->sink = Psr7\stream_for($conf[CURLOPT_FILE]);
         }
-        $lessThan1s = false;
+        $timeoutRequiresNoSignal = false;
         if (isset($options['timeout'])) {
-            $options['timeout'] < 1 && $lessThan1s = true;
+            $options['timeout'] < 1 && $timeoutRequiresNoSignal = true;
             $conf[CURLOPT_TIMEOUT_MS] = $options['timeout'] * 1000;
         }
 
@@ -386,11 +386,12 @@ class CurlFactory implements CurlFactoryInterface
         }
 
         if (isset($options['connect_timeout'])) {
-            $options['connect_timeout'] < 1 && $lessThan1s = true;
+            $options['connect_timeout'] < 1 && $timeoutRequiresNoSignal = true;
             $conf[CURLOPT_CONNECTTIMEOUT_MS] = $options['connect_timeout'] * 1000;
         }
-        if ($lessThan1s) {
-            $conf[CURLOPT_NOSIGNAL] = 1;
+
+        if ($timeoutRequiresNoSignal) {
+            $conf[CURLOPT_NOSIGNAL] = true;
         }
 
         if (isset($options['proxy'])) {
--- a/src/Handler/CurlFactory.php
+++ b/src/Handler/CurlFactory.php
@@ -390,7 +390,7 @@ class CurlFactory implements CurlFactoryInterface
             $conf[CURLOPT_CONNECTTIMEOUT_MS] = $options['connect_timeout'] * 1000;
         }
 
-        if ($timeoutRequiresNoSignal) {
+        if ($timeoutRequiresNoSignal && strtoupper(substr(PHP_OS, 0, 3)) !== 'WIN') {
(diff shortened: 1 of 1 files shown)


src/Handler/CurlFactory.php
```diff
@@ -370,8 +370,9 @@ private function applyHandlerOptions(EasyHandle $easy, array &$conf)
             $conf[CURLOPT_FILE] = fopen('php://temp', 'w+');
             $easy->sink = Psr7\stream_for($conf[CURLOPT_FILE]);
         }
-
+        $lessThan1s = false;
```
mtdowling: can we call this, `$timeoutRequiresNoSignal` or something more specific? Can you also add a comment that describes why we are doing this and link to the original comment at http://php.net/manual/en/function.curl-setopt.php#104597

src/Handler/CurlFactory.php
```diff
@@ -385,8 +386,12 @@ private function applyHandlerOptions(EasyHandle $easy, array &$conf)
         }
 
         if (isset($options['connect_timeout'])) {
+            $options['connect_timeout'] < 1 && $lessThan1s = true;
             $conf[CURLOPT_CONNECTTIMEOUT_MS] = $options['connect_timeout'] * 1000;
         }
+        if ($lessThan1s) {
+            $conf[CURLOPT_NOSIGNAL] = 1;
```
mtdowling: This should be `true` and not `1`.

sagikazarmark:
This seems to be  duplication of #1585

src/Handler/CurlFactory.php
tom916: @mtdowling  I mean in the url ,there are very detailed explanation

tom916:
@mtdowling 

I add the test case for old version

```
<?php
namespace GuzzleHttp\Tests;
use GuzzleHttp\Client;
use GuzzleHttp\Psr7\Request;
class MillSecondTest extends \PHPUnit_Framework_TestCase
{

    public function testSetTimeoutSmallThanOneSecond()
    {
        $client = new Client();
        $requestOptions = [
                'connect_timeout' => 0.1,
                'timeout' => 0.1
        ];

        try {
            $request = new Request('GET', "http://foo.com");
            $client->send($request, $requestOptions);
        } catch (\Exception $e) {
            $this->assertEquals(true, preg_match("#resolve via SIGALRM method#", $e->getMessage()));
        }
    }
}

```

src/Handler/CurlFactory.php
```diff
@@ -370,8 +370,9 @@ private function applyHandlerOptions(EasyHandle $easy, array &$conf)
             $conf[CURLOPT_FILE] = fopen('php://temp', 'w+');
             $easy->sink = Psr7\stream_for($conf[CURLOPT_FILE]);
         }
-
+        $timeoutRequiresNoSignal = false;
         if (isset($options['timeout'])) {
+            $options['timeout'] < 1 && $timeoutRequiresNoSignal = true;
```
sagikazarmark: How about doing `$timeoutRequiresNoSignal |= $options['timeout'] < 1` instead?

src/Handler/CurlFactory.php
```diff
@@ -385,9 +386,14 @@ private function applyHandlerOptions(EasyHandle $easy, array &$conf)
         }
 
         if (isset($options['connect_timeout'])) {
+            $options['connect_timeout'] < 1 && $timeoutRequiresNoSignal = true;
```
sagikazarmark: Same here

src/Handler/CurlFactory.php
tom916: @sagikazarmark 
 be better 👍

sagikazarmark:
I am actually in favor of this instead of #1585.

sagikazarmark:
For the record: here is the full explanation: http://php.net/manual/en/function.curl-setopt.php#104597

sagikazarmark:
@mtdowling this looks good to me. Are you satisfied with your requested changes?

sagikazarmark:
Ping @mtdowling

mtdowling:
:shipit:

sagikazarmark:
Thanks @tom916

PR #1721, opened 2017-01-09
[PrepareBodyMiddleware] Always include the Content-Length if there's a body

duncan3dc:
I've been working with a web service provider and their server was returning status `500` for all requests, this is because they accept a body with a `GET` request, but require the `Content-Length` header.

While this is a violation of the HTTP spec, it took a long time to notice the issue.
Once I realised the cause I thought about having Guzzle throw an exception when a body is provided with `GET`, but that would prevent Guzzle from being used with some web services, so I then thought about this approach

### DIFF
--- a/tests/PrepareBodyMiddlewareTest.php
+++ b/tests/PrepareBodyMiddlewareTest.php
@@ -13,11 +13,30 @@ use Psr\Http\Message\RequestInterface;
 
 class PrepareBodyMiddlewareTest extends \PHPUnit_Framework_TestCase
 {
-    public function testAddsContentLengthWhenMissingAndPossible()
+
+    public function methodProvider()
+    {
+        $methods = ['GET', 'PUT', 'POST'];
+        $bodies = ['Test', ''];
+        foreach ($methods as $method) {
+            foreach ($bodies as $body) {
+                yield [$method, $body];
+            }
+        }
+    }
+    /**
+     * @dataProvider methodProvider
+     */
+    public function testAddsContentLengthWhenMissingAndPossible($method, $body)
     {
         $h = new MockHandler([
-            function (RequestInterface $request) {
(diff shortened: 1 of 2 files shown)


sagikazarmark:
I think this has been reported before and the general recommendation was to create a separate middleware.

Although this turns out to be a real problem for a bunch of people, I don't think this solution is a good idea, as we would add Content-Length headers even when not necessary.

I think it would be okay to make it configurable or have a separate middleware in Guzzle's repository, but we can't make it enabled by default.

I don't know if there would be any consequences, but I would really like to avoid them.

duncan3dc:
@sagikazarmark There are many solutions, this particular provider accepts `POST` (as this request should be), this PR is more about the problem of the default Guzzle settings hiding the cause of the problem.

It could be argued that it's not Guzzle's responsibility to highlight errors with other systems, but it's a bit odd that Guzzle says you can't have a body with `GET` so doesn't send a `Content-Length` header, but does still send the body

sagikazarmark:
Hm, alright, this doesn't cause any trouble if there is no body, which should be the case with GET/HEAD (even TRACE?) requests normally.

tests/PrepareBodyMiddlewareTest.php
```diff
@@ -73,6 +73,24 @@ function (RequestInterface $request) {
         $this->assertEquals(200, $response->getStatusCode());
     }
 
+    public function testAddsContentLengthForGET()
```
sagikazarmark: I think we could reuse the existing test (maybe with a data provider) and instead add test for the case when there is no body (Content-Length should not be added then).

duncan3dc: I've reused the existing test now, and added tests for methods with and without bodies

sagikazarmark:
Thanks for your work, this will be part of the 6.3 release.

PR #1726, opened 2017-01-16
Fix query option documentation

doalex:
Hi,
I hope the behavior I documented is right, please validate :)

### DIFF
--- a/docs/request-options.rst
+++ b/docs/request-options.rst
@@ -810,7 +810,7 @@ query
     // Send a GET request to /get?foo=bar
     $client->request('GET', '/get', ['query' => ['foo' => 'bar']]);
 
-Query strings specified in the ``query`` option will overwrite a query string
+Query strings specified in the ``query`` option will overwrite all query string
 values supplied in the URI of a request.
 
 .. code-block:: php


Nyholm:
Thank you for this PR. 

I think you have misunderstood the intent. The line should say: 

> "... overwrite a query string value supplied in the URI of a request."


When running the following: 
```php
$url = 'http://www.example.com?foo=123&test=456';
$client->request('GET', $url, ['query' => ['foo' => 'bar']]);
```
You will end up requesting the url `http://www.example.com?foo=bar&test=456`

doalex:
@Nyholm strange, in my case it overwrites all parameters:
```php
$url = 'http://www.example.com?foo=123&test=456';
$client->request('GET', $url, ['query' => ['foo' => 'bar']]);
```
results in `GET /?foo=bar HTTP/1.1`, an empty query array even discards all parameters.

I noticed this behavior change while migrating from guzzle 5 to guzzle 6, which is a pain.
I'am using guzzle `6.22`.

Nyholm:
You are correct @doalex, it is me who is living in the past.

PR #1729, opened 2017-01-23
psr/log should be also in suggested packages

gmponos:
Since you have a [middleware](https://github.com/guzzle/guzzle/blob/master/src/Middleware.php#L185) in Middleware class that is using the LoggerInterface and this class is bundled with the package and it is not just for testing then I believe that psr/log should not be in require-dev.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -33,6 +33,9 @@
             "GuzzleHttp\\Tests\\": "tests/"
         }
     },
+    "suggest": {
+        "psr/log": "This package holds all interfaces/classes/traits related to PSR-3 in order to use Middleware::log()"
+    },
     "extra": {
         "branch-alias": {
             "dev-master": "6.2-dev"
--- a/composer.json
+++ b/composer.json
@@ -34,7 +34,7 @@
         }
     },
     "suggest": {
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
👎

GrahamCampbell:
This package is not **required** to use guzzle.

sagikazarmark:
I agree with @GrahamCampbell , the log middleware is an optional feature. I could imagine it being added to suggest section.

GrahamCampbell:
Yeh, suggest is the correct place for this.

gmponos:
> I agree with @GrahamCampbell , the log middleware is an optional feature. I could imagine it being added to suggest section.

I aggree also. I will make the change.

composer.json
```diff
@@ -33,6 +33,9 @@
             "GuzzleHttp\\Tests\\": "tests/"
         }
     },
+    "suggest": {
+        "psr/log": "This package holds all interfaces/classes/traits related to PSR-3 in order to use Middleware::log()"
```
sagikazarmark: I would prefer something shorter which simply states why we suggest this package (like "Required for using the Log middleware".

gmponos: OMG I had totally forgotten about this. Done.

sagikazarmark:
Thanks

PR #1746, opened 2017-02-18
[Refactor] pass handler explicitly

DQNEO:
# Background
Reading the codes, I found that `handler` is a essential element in guzzle.

# Reason to change
Meanwhile its appearance is too modest.
This PR helps readers find what is important in the `transfer()` method.

# Backword compatibility
This PR does not change the external behavior of guzzle.
It only changes the error message when non callable value is passed on construction.
(But after PR 1745 is applied,  this PR has no side effect at all)


Nyholm:
Im not sure what benefits this would bring. I rather add comments for the reader.

PR #1753, opened 2017-02-23
Fix deprecations with guzzle/psr7 version 1.4

Tobion:
According to https://github.com/guzzle/psr7/releases/tag/1.4.0

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,7 @@
     ],
     "require": {
         "php": ">=5.5",
-        "guzzlehttp/psr7": "^1.3.1",
+        "guzzlehttp/psr7": "^1.4",
         "guzzlehttp/promises": "^1.0"
     },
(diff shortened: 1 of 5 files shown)


sagikazarmark:
Thanks @Tobion 

I would probably wait for guzzle/psr7#138 to get to a resolution before merging. What do you think?

Also, should this go into a patch release or can be part of 6.3 which is imminent, but probably still needs a few weeks.

sagikazarmark:
Although it seems like the changes in this PR are not affected by the BC break being discussed in the aformentioned issue, are they?

Tobion:
This PR is independent of https://github.com/guzzle/psr7/issues/138 and can be merged regardless.

sagikazarmark:
This will be merged and released as 6.2.3 soon. The base needs to be changed though.

PR #1760, opened 2017-02-28
Fix deprecations with guzzle/psr7 version 1.4

sagikazarmark:
Cherry picked #1753 to an earlier version of guzzle so it can be tagged as a patch version, also merged into the latest master. Did it so we can immediately release it (6.3 is no blocker).

@Tobion can you please review?


Tobion:
👍

PR #1772, opened 2017-03-10
Add a method to get a cookie by cookie name

owwo:
I added a method to simply get it from the `CookieJar` by the name of the cookie.


sagikazarmark:
Duplicate of #1318

I am more in favor of the style of that PR, but there are pending requests there. If you could wrap those up, it would be nice.

sagikazarmark:
Oh, infect those are already fixed.

owwo:
Ok

PR #1778, opened 2017-03-14
Test cookie handling when cookie path is empty.

gtoboso:
Using CookieJar class I have detected that matchesPath method on SetCookie class sometimes fails when withCookieHeader method is called. 
The reason is that the cookie path attribute sent from the upstream server is an empty string.

I have created a test to check different cookie path possibilities according to the https://tools.ietf.org/html/rfc6265#section-5.1.4

### DIFF
--- a/src/Cookie/CookieJar.php
+++ b/src/Cookie/CookieJar.php
@@ -216,9 +216,39 @@ class CookieJar implements CookieJarInterface
                 if (!$sc->getDomain()) {
                     $sc->setDomain($request->getUri()->getHost());
                 }
+                if (0 !== strpos($sc->getPath(), '/')) {
+                    $sc->setPath($this->getCookiePathFromRequest($request));
+                }
                 $this->setCookie($sc);
             }
         }
+    }
+
+    /**
+     * Computes cookie path following RFC 6265 section 5.1.4
+     *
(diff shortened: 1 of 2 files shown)


src/Cookie/CookieJar.php
```diff
@@ -216,11 +216,41 @@ public function extractCookies(
...
+        if (''  === $uriPath) {
+            return '/';
+        }
+        if (0 !== strpos($uriPath, '/')) {
+            return '/';
+        }
+        if ('/' === $uriPath) {
+            return '/';
+        }
+        if (0 === $lastStashPos = strrpos($uriPath, '/')) {
```
ajgarlag: This var should be named `$lastSlashPos`

src/Cookie/CookieJar.php
```diff
@@ -216,11 +216,41 @@ public function extractCookies(
...
+            return '/';
+        }
+
+        return substr($uriPath, 0, $lastStashPos);
```
ajgarlag: This var should be named `$lastSlashPos`

sagikazarmark:
Thanks for your contribution, I will review it shortly.

Also, thanks for the review @ajgarlag

PR #1779, opened 2017-03-14
[WIP] Fix stream size after decoding content

ajgarlag:
When curl decodes the original body contents, the returned response stream has a fixed size of `0`, but it should be `null` or the real body size.


oscarotero:
Thanks for this fix, it is important.

I have a PSR-15 middleware to implement a proxy using guzzle.
The response body size returns `0`, and because this value is cached, I have to do this to refresh the cache and recalculate the correct stream size:

https://github.com/middlewares/proxy/blob/master/src/Proxy.php#L102

ajgarlag:
This is not really a fix, but a test to demonstrate the problem. I discovered it building a proxy too (a PSR-15 delegate). To avoid it, I've disabled content decoding and wrote a middleware to decode the response body.

Not sure how to really fix it in guzzle. Any hint is welcome.

Nyholm:
This has been in WIP for about 3 years. I'll close this PR now =)

PR #1781, opened 2017-03-17
Add functionality that allow redirect with Auth credentials

StanislavPivovarov:


sagikazarmark:
Hm, this is interesting. I would like to see some more expressive name for the option, like `remove_auth` or even better `keep_auth`. This would also need some documentation with exclamation marks that this could be serious security problem.

Actually, thinking about security, not sure if this is a good idea after all. Apart from your use case, I consider this as bad practice.

sagikazarmark:
I have been thinking about this a lot and I decided not to accept this change. The main reason is security: I consider this a serious security flaw and even if it can be justified in some cases, this would lead to more confusion and security issues than add any gain.

I would suggest you adding authorization headers *after* the redirect plugin in your chain.

Thanks for your work anyway, sorry for the negative answer.

PR #1792, opened 2017-03-30
Fix #1790 proxy should not end with /

### LINKED ISSUE #1790: Test failure with 7.1.4RC1
FYI, detected by Fedora QA
https://apps.fedoraproject.org/koschei/package/php-guzzlehttp-guzzle6?collection=f27

The "Failed to parse address "127.0.0.1:8126/" error seems ugly...

Can you please have a look, so we can try to avoid BC break in upcoming 7.1.4.

remicollet:
@sgolemon  can you please have a look, related to fix from https://bugs.php.net/74216

Perhaps better to ignore this ending / in php side to avoid BC

But, probably,  fixing here make sense in all case

### DIFF
--- a/src/Handler/StreamHandler.php
+++ b/src/Handler/StreamHandler.php
@@ -381,7 +381,7 @@ class StreamHandler
                         $value['no']
                     )
                 ) {
-                    $options['http']['proxy'] = $value[$scheme];
+                    $options['http']['proxy'] = rtrim($value[$scheme], '/');
                 }
             }
         }
--- a/tests/Handler/StreamHandlerTest.php
+++ b/tests/Handler/StreamHandlerTest.php
@@ -272,7 +272,7 @@ class StreamHandlerTest extends \PHPUnit_Framework_TestCase
         $url = str_replace('http', 'tcp', Server::$url);
         $res = $this->getSendResult(['proxy' => ['http' => $url]]);
         $opts = stream_context_get_options($res->getBody()->detach());
-        $this->assertEquals($url, $opts['http']['proxy']);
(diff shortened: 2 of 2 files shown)


remicollet:
See https://github.com/php/php-src/pull/2443

remicollet:
Can we get some feedback for project owner about the issue ? (ping @sagikazarmark)

RC time is only 2 weeks, and I'd like to see some consensus on this before 7.1.4

(this PR is only there to point where the failure is, perhaps not perfect, but discussion really needed)

mtdowling:
This seems reasonable to me. @sagikazarmark what do you think?

remicollet:
So, seems to be an agreement to not fix in PHP side, as bad data should not be accepted.

Question is: should we only fix the test, or also the library (add the trim) to be error tolerant.

sagikazarmark:
I see the problem and I think this is rather a potential quick fix, but not a solution to it.

I guess the question here is what a valid value is for a proxy. The slash at the end is clearly not, at least not for me. I would say allowing it is a bug, which should be fixed by either validating the value and throwing exceptions, or parsing it as a URI and silently reconstructing the value using only schema, host and port segments. Both sound risky from a BC break point of view, so maybe we should just merge this and properly fix in a follow up issue.

BTW this fix should be added as a 6.2.x patch (that's the current stable release line).

sagikazarmark:
@remicollet I am going to accept #1811 for now as it contains more PHP 7.2 related test fixes. For now I decided to solve it only in the tests and add a fix later if necessary.

Thanks for reporting the issue.

PR #1795, opened 2017-03-31
Fix/transfer encoding fix for get method

smecnarowski:
When CURLOPT_UPLOAD is sent to curl_setopt method it's resulting with 'Transfer-uncoding: chunked' which rather should not be present in GET or HEAD requests as in section 4.4 of https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html

### DIFF
--- a/src/Handler/CurlFactory.php
+++ b/src/Handler/CurlFactory.php
@@ -21,6 +21,9 @@ class CurlFactory implements CurlFactoryInterface
     /** @var int Total number of idle handles to keep in cache */
     private $maxHandles;
 
+    /** @var array */
+    private static $skipMethods = ['GET' => true, 'HEAD' => true];
+
     /**
      * @param int $maxHandles Maximum number of idle handles.
      */
@@ -258,7 +261,7 @@ class CurlFactory implements CurlFactoryInterface
             // Don't duplicate the Content-Length header
             $this->removeHeader('Content-Length', $conf);
             $this->removeHeader('Transfer-Encoding', $conf);
-        } else {
+        } elseif (!isset(self::$skipMethods[$request->getMethod()])) {
             $conf[CURLOPT_UPLOAD] = true;
             if ($size !== null) {
                 $conf[CURLOPT_INFILESIZE] = $size;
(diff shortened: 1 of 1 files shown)


sagikazarmark:
While I generally agree with this, GET and HEAD requests CAN have a body.

smecnarowski:
Sorry i put wrong paragraph. What I was thinking about was written in section 4.3:

> The presence of a message-body in a request is signaled by the inclusion of a Content-Length or Transfer-Encoding header field in the request's message-headers. A message-body MUST NOT be included in a request if the specification of the request method (section 5.1.1) does not allow sending an entity-body in requests. A server SHOULD read and forward a message-body on any request; if the request method does not include defined semantics for an entity-body, then the message-body SHOULD be ignored when handling the request.

I needed to connect IIS server behind ARR proxy and there i found this problem. The second one is very restrictive about above MUST-NOT be included and throws 502 error.

sagikazarmark:
Okay. Since we have another PR which tries to loosen these rules within Guzzle, I would like to cross-check with that one to see how it would interfere with this one before merging. But if there is a conflict between sending and not sending body in a GET/HEAD request, I would like Guzzle to stick to the standard and what some servers do in ninja mode.

smecnarowski:
OK - Thank you

smecnarowski:
@sagikazarmark And how it looks like ?

Nyholm:
Thank you for this PR. I do understand your use case but Im not sure the solution is the best one. As Mark said, GET request may have a body. 

In side `applyBody` we cannot really tell if there is a body or not. However, if it is a HEAD request, no body is allowed and we should `unset($conf[CURLOPT_UPLOAD])` in `applyMethod``. (Like we do with CURLOPT_READFUNCTION). 

To solve your issue for GET request you may set a Content-Length header to 0.

sagikazarmark:
I think the aformentioned PR is already merged and will be released as 6.3

I guess merging this PR would go quite the opposite direction, so I am not sure at this point we can merge it as well.

smecnarowski:
Sorry for late answer. For my point of view and real live problem I would like to share one on Stackoverflow user analyze. https://stackoverflow.com/a/15656884

PR #1809, opened 2017-04-13
Fix counting null variable (PHP 7.2)

scaytrase:


sagikazarmark:
Thanks, looks great. There are a few more places though where I could imagine this fix:

- static and normal constructor default value should be an empty array as well
- queue empty checks should use `empty` instead of just evaluating it as a boolean value

Can you do these fixes as well?

scaytrase:
For more complex approach we should just add nightly php to travis test matrix (with allowed failures for some time) and see the failures. I can try it

scaytrase:
Not sure about changing signatures as soon as it might provide any kind of BC breaks, but hardening checks sounds good

sagikazarmark:
> For more complex approach we should just add nightly php to travis test matrix (with allowed failures for some time) and see the failures.

Agreed

> Not sure about changing signatures as soon as it might provide any kind of BC breaks, but hardening checks sounds good

Manually passing null to an array typehinted parameter is not possible in the first place, so I think changing the default value is safe in this case.

scaytrase:
> Manually passing null to an array typehinted parameter is not possible in the first place, so I think changing the default value is safe in this case.

What about inheritance and `parent` calls? Most of classes are not final thus could be extended.

The following example fails on 7.1:

```php

class A
{
    public function __construct(array $val = [])
    {
        var_dump($val);
    }
}

class B extends A
{
    public function __construct(array $val = null)
    {
        parent::__construct($val);
    }
}

$b = new B();
```

```
PHP Fatal error:  Uncaught TypeError: Argument 1 passed to A::__construct() must be of the type array, null given
```

So changing signature should be considered BC-break

scaytrase:
Closing in favor of #1811

sagikazarmark:
Hm, you are right :(

I thought this would fail too: https://3v4l.org/DWTTP

PR #1810, opened 2017-04-13
Simple function refactoring

MarekBielecki:


sagikazarmark:
Thanks for the contribution. I am not entirely sure what's the point though. How does it make this function better?

MarekBielecki:
Function now is less complex I think. I'm just starting an adventure with your code. Nothing will happen if You reject this request :)

sagikazarmark:
Sorry, I don't think it makes anything less complex, quite the opposite actually. Thanks for the effort though.

PR #1811, opened 2017-04-14
Add PHP 7.2 testing

scaytrase:
* Include PHP 7.2 to travis matrix
* Fix TCP proxy url (trailing slash parsing fails on PHP 7.2)
* Update to phpunit `^4.0 || ^5.0` (4.0 is no longer supported, but required for testing against PHP 5.5)
* Add tests for countables

At the moment no more failures found for 7.2 and tests are green. Further patches are matter of test coverage

Incorporates #1809 
cc. @sagikazarmark

### DIFF
--- a/tests/Cookie/CookieJarTest.php
+++ b/tests/Cookie/CookieJarTest.php
@@ -39,7 +39,7 @@ class CookieJarTest extends \PHPUnit_Framework_TestCase
 
     public function testEmptyJarIsCountable()
     {
-        self::assertCount(0, new CookieJar());
+        $this->assertCount(0, new CookieJar());
     }
 
     public function testGetsCookiesByName()
--- a/tests/Cookie/CookieJarTest.php
+++ b/tests/Cookie/CookieJarTest.php
@@ -37,6 +37,11 @@ class CookieJarTest extends \PHPUnit_Framework_TestCase
         $this->assertCount(2, $jar);
     }
 
+    public function testEmptyJarIsCountable()
+    {
+        self::assertCount(0, new CookieJar());
+    }
+
     public function testGetsCookiesByName()
     {
         $cookies = $this->getTestCookies();
--- a/tests/Handler/StreamHandlerTest.php
+++ b/tests/Handler/StreamHandlerTest.php
@@ -270,6 +270,7 @@ class StreamHandlerTest extends \PHPUnit_Framework_TestCase
     public function testAddsProxyByProtocol()
     {
         $url = str_replace('http', 'tcp', Server::$url);
+        $url = rtrim($url, '/');
         $res = $this->getSendResult(['proxy' => ['http' => $url]]);
         $opts = stream_context_get_options($res->getBody()->detach());
         $this->assertEquals($url, $opts['http']['proxy']);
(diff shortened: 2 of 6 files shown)


sagikazarmark:
We already have a proposed fix for the proxy thing in #1792

It's not decided yet what should be the final solution though.

scaytrase:
@sagikazarmark I'm not fixing it, but just "hiding" the problem with testing against *correct* URI. 

If\When you decide to fix this in a proper way you choose you can add more tests against slash-trailed URIs to check that  behavior is correct

tests/Cookie/CookieJarTest.php
```diff
@@ -37,6 +37,11 @@ public function testCreatesFromArray()
         $this->assertCount(2, $jar);
     }
 
+    public function testEmptyJarIsCountable()
+    {
+        self::assertCount(0, new CookieJar());
```
Aliance: Why `self::` not `$this->`. I think code should be uniform.

Aliance: May be `$this->assertInstanceOf` will be better?

scaytrase: Accepted `$this->`, will fix this. 

Did not understand about `instanceOf`. This test explicitly invokes the `count()` method to be sure, that internals of `CookieJar` is countable from construction with 7.2

Aliance: As I understand, you want to check the existence of `count` method. Am I right? But actually you check that `count == 0`. So I'm suggesting you to check implementing of the `\Countable` interface.

scaytrase: No, I wan't to actually invoke the `count` method. Because PHP 7.2 throws a warning when trying to count a non-countable things like `null`, which could be done internally (i.e counting internal property which was not initialized) . So I want to check that `CookieJar` does not throws the warning when calling count directly after constructing. 

https://wiki.php.net/rfc/counting_non_countables

The same thing I did with the `MockHandler`, but we have to fix it (see array default)

Aliance: You wrote:
> So I want to check that CookieJar does not throws the warning when calling count

But actually you
> you check that `count == 0`

So nothing changes in my mind if you will check implementing of the \Countable interface^ because it guarantees that calling `count` won't fail.

But it's kind of a holywar... =)

scaytrase: Having `count` method implemented does not guarantees that id does not fail. Short example:

```php
class foo implements \Countable {
   private $things;

   public function count() { return count($this->things); }

   public function push($element) { $this->things[] = $element; } 
}

$f = new \foo();
$f->count(); // this will pass on PHP < 7.2 and throw warning on PHP 7.2
$f->push(1);
$f->count(); // this will pass
```

So having `count` method implemented does not guarantee that it **works properly internally**.

Checking that `count == 0` is needed because the future phpunit versions will mark tests without assertions as risky, actually I could just call the `(new CookieJar())->count()` without any assertions.

tests/Handler/StreamHandlerTest.php
```diff
@@ -270,6 +270,7 @@ public function testAddsProxy()
     public function testAddsProxyByProtocol()
     {
         $url = str_replace('http', 'tcp', Server::$url);
+        $url = rtrim($url, '/');
```
sagikazarmark: Can you add a quick not why we are doing this? Just to make sure we find it. Thanks

sagikazarmark:
Ok, one small comment, then it's good to go.

tests/Handler/StreamHandlerTest.php
scaytrase: Done, I've referred your #1823 issue

sagikazarmark:
Thanks for your work.

PR #1827, opened 2017-05-03
Use `\GuzzleHttp\Promise\rejection_for` function instead of object init.

Aliance:


sagikazarmark:
Thanks for the PR.

Can you tell me what's the advantage of this?

Aliance:
Uniformity. There is an alias functions, it some cases they were used, in others - not. I just made it uniform.

Sorry for missing the description. I thought it was clear.

sagikazarmark:
Hm, okay. I don't really care so much about it, but it's an improvement after all, so thank you.

PR #1842, opened 2017-05-24
Re-seek body streams after logging

klaude:
If a request or response body is seekable then log its entire contents regardless of their streams' current positions and `seek()` their streams back to their original positions after logging. If a body stream is not seekable then log an empty body to preserve the original stream's state.

This is a modified version of pull request #1262, attempting to follow @mtdowling's example while addressing @sagikazarmark's comments. Hopefully this can help address stream-related side effects from logging. 

I went with doing stream `tell()`ing, `seek()`ing, and `rewind()`ing in the log middleware to avoid having to modify both the `MessageFormatter`'s `{req_body}` and `{res_body}` template tags and `GuzzleHttp\Psr7\str()`, which is used by the formatter's `{request}` and `{response}` tags. It's made the logger middleware a bit bigger, but I'm open to suggestions for simplifying.

I've tested this to work locally in PHP 5.5.38, 5.6.30, 7.0.19, and 7.1.5 on macOS Sierra. New tests achieve 100% code coverage of the new work.

There may be a way to remove more stream-related side effects across the whole library, but the only way I can think of to do it would be some low level modifications to message bodies, probably by writing to a duplicate buffer when the underlying stream is read. That's a sizeable change and outside the scope of a logging fix. I'm open to discussing that outside this PR if y'all are. Thanks for your consideration!


sagikazarmark:
Thanks for your work, I will check it out soon.

klaude:
I saw this PR was bumped yesterday, so I fixed up the test syntax to match changes adopted over the past two years. The BC check is failing, but that looks like it's unrelated to this work. Let me know if this looks alright. Thanks!

ossinkine:
@klaude @sagikazarmark @Nyholm Huys, let's review and merge this PR

GrahamCampbell:
I'm still not sure this is a good idea, actually.

ossinkine:
@GrahamCampbell What is your proposal? I'm sorry if you already posted it anywhere, please just leave a link.

klaude:
I'm down for whatever's the best fix for the "where'd my response body go after I log it" issue. There's been discussion about it for years now with a number of proposed fixes. 

I'm happy to fix conflicts so this can cleanly go into master. If one of the other PRs is a better way to do it then that's fine too. If none of the PRs are good then maybe this kind of issue is best fixed in documentation. I can help there too.

mfn:
`->getContents()` vs. `->__toString()` and the impact on seekable streams is definitely a gotcha people run into.

I implemented by own logging and experienced exactly the same issues, as probably everyone else. Not sure _where_ in the docs it would fit best, but that would be a good addition either way.

Nyholm:
Thank you for the PR. I think it is well written. 

I dont think we should care about where the pointer in the stream is. If you want to read the stream from the beginning, then you move the pointer. Dont assume that it is at the beginning. 

I am not too keen on moving the pointer in situations it is not needed either.

PR #1843, opened 2017-05-25
Fix describe_type()'s test when Xdebug is loaded

klaude:
Xdebug's overloaded `var_dump()` may generate different output from the built-in `var_dump()`, causing this test to fail. Temporarily disable Xdebug's `var_dump()` during the test.

I ran into this while working on #1842. I usually avoid extension specific logic branches, but Xdebug may be common enough in development environments to warrant it here. Thanks!

### DIFF
--- a/tests/functionsTest.php
+++ b/tests/functionsTest.php
@@ -51,10 +51,12 @@ class FunctionsTest extends TestCase
             ini_set('xdebug.overload_var_dump', 0);
         }
 
-        self::assertSame($output, GuzzleHttp\describe_type($input));
-
-        if (extension_loaded('xdebug')) {
-            ini_set('xdebug.overload_var_dump', $originalOverload);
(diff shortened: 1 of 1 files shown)


sagikazarmark:
Thanks for the PR.

Looking at Travis, we have the overload specifically turned on for PHP 5.x versions:

https://github.com/guzzle/guzzle/blob/master/.travis.yml#L18

TBH I have no idea why. @Nyholm any ideas?

Nyholm:
Ahh, lovely. Thank you for this PR. =)

I rebased it and just waiting for the test to run. 


(it works locally)

GrahamCampbell:
This needs a try-finally, otherwise the code to restore it will never run?

Nyholm:
Oh, thank you @GrahamCampbell =)

PR #1851, opened 2017-06-01
Update Quickstart.rst to fix syntax issue

Erikmitk:
I’m just getting started with Guzzle so I might be wrong about that. I just tried to implement by example and I believe the given syntax from the quickstart example doesn’t work that way. My example didn’t respond as I expected and that change made it work.

`$headers` has to be a nested array otherwise it will be ignored by the request. The example doesn’t read as concise as before but that way it should be correct and work out of the box.

(I’m not sure about $body since I haven’t used that yet, but I suspect the same issue.)

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -131,7 +131,7 @@ You can also use the `sendAsync()` and `requestAsync()` methods of a client:
     use GuzzleHttp\Psr7\Request;
 
     // Create a PSR-7 request object to send
-    $headers = ['X-Foo' => 'Bar'];
+    $headers = ['headers' => ['X-Foo' => 'Bar']];
     $body = 'Hello!';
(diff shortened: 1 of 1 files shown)


sagikazarmark:
I am not sure if that's right. The constructor clearly accepts a header array:

https://github.com/guzzle/psr7/blob/master/src/Request.php#L35

/cc @Tobion 

I just gave it a quick shot, I will try to take a closer look later.

Tobion:
The change is wrong. See tests of guzzle/psr7 etc.

Erikmitk:
Ah, I see my mistake.

The example directly uses `new Request(…)` where this works fine. But if I start with `$client->request('GET', $path , $headers);` it wouldn’t, right? `$headers` in that case would be the whole group of options and not only my headers. Sorry for the confusion. :)

PR #1852, opened 2017-06-03
Trigger an E_USER_DEPRATED warning if handler is passed as an option

gmponos:
Developers are able to pass a handler option and override the handler that was passed during construction of the client.

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -4,9 +4,9 @@ namespace GuzzleHttp;
 use GuzzleHttp\Cookie\CookieJar;
 use GuzzleHttp\Promise;
 use GuzzleHttp\Psr7;
-use Psr\Http\Message\UriInterface;
 use Psr\Http\Message\RequestInterface;
 use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\UriInterface;
 
 /**
  * @method ResponseInterface get(string|UriInterface $uri, array $options = [])
@@ -230,6 +230,13 @@ class Client implements ClientInterface
             } elseif (!is_array($options['headers'])) {
                 throw new \InvalidArgumentException('headers must be an array');
             }
+        }
+
+        if (isset($options['handler'])) {
+            @trigger_error(
+                'Sending a request using handler as an option is deprecated since version 6.3 and will be removed in 7.0.',
+                E_USER_DEPRECATED
+            );
         }
 
         // Shallow merge defaults underneath options.


gmponos:
Ping?

sagikazarmark:
Sorry, I missed this one.

Thinking about it, but I am not sure why we would not want to allow that. Of course it's not a good idea, but even if we remove it in a major version, it's an undocumented feature, so I don't feel like we have to trigger a deprecation notice.

WDYT @Nyholm ?

gmponos:
> Thinking about it, but I am not sure why we would not want to allow that.

From the docblock of the Client constructor [here](https://github.com/guzzle/guzzle/blob/master/src/Client.php#L50)

> "handler" is a constructor only option that cannot be overridden in per/request options.

Maybe the docblock is wrong but the way I read it I thought that allowing overriding the handler is a bug. `Cannot` is pretty strictful.

> it's an undocumented feature, so I don't feel like we have to trigger a deprecation notice

In the way you put it you might be right. That since it's undocumented it would be easy to remove to a major version but as I said before I thought it is a bug.

sagikazarmark:
You are right, it's not allowed. But then I would feel it a little bit contradictive to deprecate something which is not supported (according to docs) in the first place. It's probably a bug though, but I think we can just add a check in 7.0.

Thanks for the contribution though.

guilliamxavier:
Please do not remove it, it has proven useful for me. It is even actually "documented" (kind of) at http://docs.guzzlephp.org/en/stable/request-options.html#json (adding `Middleware::tap` for a specific request).

History:

- The statement that "`handler` is a constructor only option that cannot be overridden in per/request options" was added to the phpDoc as part of commit 5eabdb6db0bf96a40754607c88bd7b6c8b7a4b13 "Simplifying some client methods and removing URI template arrays" (and reflected in the docs as part of commit b0a33d8923e0aa7a22eb7f673208e28988dcfda7 "Updating docs") by the end or March 2015, when that was *effectively* true (in 6.0-dev, when 6.0.0 was not released yet).

- Two months later the same person (@mtdowling) made commit 1cb0535933779dbc68f0155ce7a884900b9365f9 "Renaming getDefaultOption to getConfig. No longer removing the handler from the config. *Allowing the handler to be overridden in request options*" (emphasis mine), and a few days later *6.0.0 was released with this feature*.

- A few weeks later commit e812f17245f6b6678e39661fce8102101e392e6f "Fixing `json` docs" added an example using this feature in the documentation (part of release 6.0.2).

So what about rather *removing the obsolete statement from the docs* (which probably should have been done as part of 1cb0535933779dbc68f0155ce7a884900b9365f9 but anyway)?

PR #1862, opened 2017-06-22
Make sure we at least run the test on HHVM

Nyholm:


Nyholm:
Thank you for merging.

Issue #1582, opened 2016-09-08
Empty body when debugging request with `Middleware::log`


pbuyle:
If you do something like

```
...
$stack = \GuzzleHttp\HandlerStack::create();
$stack->push(Middleware::log($logger, new MessageFormatter($message_template)));
...
```

Where `$message_formatter` include `{request}` or `{res_body}`, then the message formatter will consume the response's body. Because the body is a stream, any other piece of code consuming the body will have to rewind the stream (if possible, otherwise the response is lost to them).

As a consumer of the API, I would assume `Middleware::log` would take care to ensure the returned log middleware does not actually alter the logged objects.

This behavior is not documented, neither for `Middleware::log` or `MessageFormatter`. Someone with a good understanding of the behind the scene stream handling may guess it, but otherwise users are left to wonder what's going on and what they did wrong.

pbuyle:
The following is a workaround this issue, but I've not been using Guzzle for long so I'm not sure of its robustness...
```php
$stack = \GuzzleHttp\HandlerStack::create();
$stack->push(
  function (callable $handler) use ($logger, $message_formatter, $logLevel) {
    return function ($request, array $options) use ($handler, $logger, $message_formatter, $logLevel) {
      return $handler($request, $options)->then(
        function (ResponseInterface $response) use ($logger, $request, $message_formatter, $logLevel) {
          if (!$response->getBody()->isSeekable()) {
            $response = $response->withBody(new CachingStream($response->getBody()));
          }
          $position = $response->getBody()->tell();
          $message = $message_formatter->format($request, $response);
          $response->getBody()->seek($position);
          $logger->log($logLevel, $message);
          return $response;
        },
        function ($reason) use ($logger, $request, $message_formatter) {
          $response = $reason instanceof RequestException
            ? $reason->getResponse()
            : null;
          if ($response) {
            if (!$response->getBody()->isSeekable()) {
              $response = $response->withBody(new CachingStream($response->getBody()));
            }
            $position = $response->getBody()->tell();
          }
          $message = $message_formatter->format($request, $response, $reason);
          if ($response) {
            $response->getBody()->seek($position);
          }
          $logger->notice($message);
          return \GuzzleHttp\Promise\rejection_for($reason);
        }
      );
    };
  }
);
``` 

[...]

PR #1872, opened 2017-07-04
Fix the empty response body when logging {res_body}

perico:
This only fixes the issue with {res_body} variable.

When logging with {res_body} the Stream goes to the eof.

This fix will make to return:
- `NULL` when response is null
- `RESPONSE_NOT_LOGGEABLE` when the Stream is not seekable.
- Body to String and rewind if possible

To fix the {response} I would tackle into Psr7\str class

Also found on the following issues #1262 and  #1582

### DIFF
--- a/src/MessageFormatter.php
+++ b/src/MessageFormatter.php
@@ -109,7 +109,7 @@ class MessageFormatter
                         $result = $request->getBody();
                         break;
                     case 'res_body':
-                        if(!isset($response)){
+                        if(!$response instanceof ResponseInterface){
                             $result = 'NULL';
                             break;
                         }
@@ -122,7 +122,6 @@ class MessageFormatter
                         }
 
                         $result = $response->getBody()->__toString();
-                        $response->getBody()->rewind();
                         break;
                     case 'ts':
                     case 'date_iso_8601':
--- a/src/MessageFormatter.php
+++ b/src/MessageFormatter.php
@@ -109,14 +109,14 @@ class MessageFormatter
                         $result = $request->getBody();
                         break;
                     case 'res_body':
-                        if(!$response instanceof ResponseInterface){
+                        if (!$response instanceof ResponseInterface) {
                             $result = 'NULL';
                             break;
                         }
 
                         $body = $response->getBody();
 
-                        if(!$body->isSeekable()){
+                        if (!$body->isSeekable()) {
                             $result = 'RESPONSE_NOT_LOGGEABLE';
                             break;
                         }
--- a/src/MessageFormatter.php
+++ b/src/MessageFormatter.php
@@ -109,7 +109,20 @@ class MessageFormatter
                         $result = $request->getBody();
                         break;
                     case 'res_body':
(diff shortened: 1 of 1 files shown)


julillosamaral:
+1!

AV4TAr:
👍

sagikazarmark:
Thanks for the contribution. While I consider whether this is the proper fix for the problem, I would like to propose a bit of refactor which makes the code cleaner in my opinion:

``` php
                        $result = 'NULL';

                        if ($response !== null) {
                            $body = $response->getBody();

                            if ($body->isSeekable()) {
                                $result = (string) $body;
                                $body->rewind();
                            } else {
                                $result = 'RESPONSE_NOT_LOGGABLE';
                            }
                        }
```

Less breaks makes it more linear.

perico:
Hi @sagikazarmark, 
Sure I am totally fine with the approach. I usually try to [avoid the arrow pattern](https://lostechies.com/derekgreer/2009/10/05/the-arrow-anti-pattern/) But I am totally ok with this.
Thanks

coudenysj:
The problem with this fix (in contrast to #1262 or #1842, where the fix is on a higher level), is that the response is possibly used multiple times in the MessageFormatter:

* https://github.com/perico/guzzle/blob/564d0e155a03f7fef47940ce9f2602adcd1462ba/src/MessageFormatter.php#L88
* https://github.com/perico/guzzle/blob/564d0e155a03f7fef47940ce9f2602adcd1462ba/src/MessageFormatter.php#L110-L123

dimabory:
Let's resolve an issue already.

sonikarc:
Hi all, any updates on this one?

Is someone able to provide a summary of what the actual issue preventing this from being resolved? I'd be happy to tackle this one if I understood a little bit more about it.

The most comprehensive solution seems to be in #1842 which rewinds the stream in the Log middleware itself (if possible) however as @sagikazarmark mentioned in #1582 it seems to only solve half the problem, what is the other half of the issue?

src/MessageFormatter.php
```diff
@@ -107,7 +107,20 @@ function (array $matches) use ($request, $response, $error, &$cache) {
...
+
+                        $body = $response->getBody();
+
+                        if(!$body->isSeekable()){
+                            $result = 'RESPONSE_NOT_LOGGEABLE';
+                            break;
+                        }
+
+                        $result = $response->getBody()->__toString();
+                        $response->getBody()->rewind();
```
Nyholm: No need to rewind it.

GrahamCampbell:
Ping @perico.

Nyholm:

src/MessageFormatter.php
```diff
@@ -107,7 +107,20 @@ function (array $matches) use ($request, $response, $error, &$cache) {
                         $result = $request->getBody();
                         break;
                     case 'res_body':
-                        $result = $response ? $response->getBody() : 'NULL';
+                        if(!isset($response)){
```
Nyholm, line 110:
```suggestion
                        if(!$response instanceof ResponseInterface){
```

PR #1879, opened 2017-07-21
Updating the date on the liscense

AchoArnold:


Nyholm:
The date should be the year we first created the product. The correct way of writing a copyright statement would be: 
```
Copyright (c) 2011 Michael Dowling
```

GrahamCampbell:
> The correct way of writing a copyright 

"A correct way", not "the correct way. ;)

sagikazarmark:
AFAIK it is THE correct way from a legal point of view. Copyright (unlike a trademark) does not have to be renewed. The standard length for copyright is the life of the person how holds them + 50 years.

But I think there is nothing wrong with extending the copyright year like that. For example some advise to do so after a major refactor.

Unfortunately I haven't found a single guide for software developers which helps us with this legal matter.

sagikazarmark:
Thanks, but I think it's just fine how it is right now.

PR #1884, opened 2017-07-26
Prevent concurrent writes

### LINKED ISSUE #1883: Invalid JSON in FileCookieJar
| Q            | A
| ------------ | ---
| Bug?         | yes
| New Feature? | no
| Version      | 5.3.1

#### Actual Behavior

Invalid json is saved in the file used as backend for `FileCookieJar`. I have had this issue in the past with what might be another version, I remember there were 2 extra brackets at the end of the file. I experienced the issue today again, but I arrived too late to see the invalid contents, which had already been replaced (which kind of makes me think that there might have been many concurrent writes). I know it was invalid because I got: Unable to parse JSON data: JSON_ERROR_SYNTAX - Syntax error, malformed JSON and it came from [here](https://github.com/guzzle/guzzle/blob/5.3/src/Cookie/FileCookieJar.php#L77).

#### Expected Behavior

A cookie jar containing valid JSON

#### Steps to Reproduce

I'm not sure how to reproduce this, I think it might happen with many concurrent writes.

#### Possible Solutions

Just prevent concurrent writes:

```diff
-        if (false === file_put_contents($filename, json_encode($json))) {
+        if (false === file_put_contents($filename, json_encode($json), LOCK_EX)) {
```

greg0ire:
Concurrent writes might lead to invalid JSON being saved in the cookie jar.
Hopefully fixes #1883 .

### DIFF
--- a/src/Cookie/FileCookieJar.php
+++ b/src/Cookie/FileCookieJar.php
@@ -50,7 +50,7 @@ class FileCookieJar extends CookieJar
             }
         }
 
-        if (false === file_put_contents($filename, json_encode($json))) {
+        if (false === file_put_contents($filename, json_encode($json), LOCK_EX)) {
             // @codeCoverageIgnoreStart
             throw new \RuntimeException("Unable to save file {$filename}");
             // @codeCoverageIgnoreEnd


j0k3r:
Shouldn't this be backported to the master so it'll be available for the 6.x version too?

sagikazarmark:
Yeah, that's a good idea. Are you up for it?

PR #1888, opened 2017-07-28
Added a test case for version option

gmponos:
This just adds a test case that vesion option is set correctly

### DIFF
--- a/tests/ClientTest.php
+++ b/tests/ClientTest.php
@@ -10,6 +10,7 @@ use GuzzleHttp\Psr7;
 use GuzzleHttp\Psr7\Request;
 use GuzzleHttp\Psr7\Response;
 use GuzzleHttp\Psr7\Uri;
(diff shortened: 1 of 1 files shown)


gmponos:
Basically I could not understand how version was overwriiten when it was passed as an option.
Since I could not understand this I checked the tests to see if there was a test case and since there was not I decided to write some to see if everything works as expected.

Hope they are fine.

gmponos:
Hello,

I was checking the code of client and I saw that although you ovewrite a lot of options when the users passes inside the options array like the body of the request ([check this](https://github.com/guzzle/guzzle/pull/1888/files#diff-668268a1040801a44451100bf9062b17R309)) you are not doing the same with the version option. Check the tests added.

Is this intentional or would you like me to fix this?

Nyholm:
What is the status on this?

PR #1890, opened 2017-08-01
Apply on request the version set in options parameters

gmponos:
This will allow [this MR](https://github.com/guzzle/guzzle/pull/1888) to pass the tests and be merged

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -4,9 +4,9 @@ namespace GuzzleHttp;
 use GuzzleHttp\Cookie\CookieJar;
 use GuzzleHttp\Promise;
 use GuzzleHttp\Psr7;
-use Psr\Http\Message\UriInterface;
 use Psr\Http\Message\RequestInterface;
 use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\UriInterface;
 
 /**
  * @method ResponseInterface get(string|UriInterface $uri, array $options = [])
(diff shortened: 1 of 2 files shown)


gmponos:
ping.

gmponos:
ping again.

Nyholm:
Why do we need this?

Isnt the Request responsible for their properties? Why should that be overridden?

gmponos:
1. Well the precedence on everything else is on the RequestOptions when you are using the `send` function. I think for instance the query is overwritten.

2. This will unlock new possibilities.

> The reason is that you can use the PSR factory to create a request just by using the URL and the method/verb. Afterwards applyOptions of guzzle will handle all the rest. The only thing that applyOptions is not handling at the moment is the version. So imy PR above gets merge then you will be able to use a PSR Factory.

2a. If you could ever consider allowing guzzle to inject a RequestFactory during construct. Then how would you set the version of the request?

2b. I could extract modify to something separate. This will give me the ability to use PSRs and modify request through guzzle.

What I mean:

```
MyHanlder {
   public function __construct(PsrClient $client, PsrFactory $factory) {

    }

   public function send(string $email) {
         $request =  $this->factory->createRequest('get', 'www.myuri.com');
         Guzzlemethod_to_modify($request, [
                 // Guzzle options
         ]);
         $this->client->sendRequest($request);
   }
}
```

Now on the example above I use guzzle but I have a lazy function to modify the request according to my needs.. and I can easily switch from guzzle to anything else... All these blocked by this simple thing :)

Nyholm:
Sorry for the delay. I think this makes sense. 

Could you rebase the PR?

PR #1907, opened 2017-09-04
fix faq php.net link

ceau:
the link of the php.net is an error

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -69,7 +69,7 @@ used with a client.
 How can I add custom stream context options?
 ============================================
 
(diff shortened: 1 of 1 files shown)


sagikazarmark:
I can't see the error. What do you see?

ceau:
It is a coincidence. I use the guzzle first time. I open the link, the browser give me a error. Then I enter [php.net](http://php.net/), It is ok . but when you ask me the question , I open the link again, It is ok . That is very strange.

sagikazarmark:
Looks like there is nothing to do here.

Issue #1909, opened 2017-09-06
Undefined Offset: 1 in /guzzle/src/Handler/EasyHandle.php


Vladnev:
**Bug**
**Version:** 6.3.0

When I try to take headers from one of urls(I think it can be repeat) has error:
>Undefined offset: 1 in file: vendor/guzzlehttp/guzzle/src/Handler/EasyHandle.php line: 77

**Call example with url:**
```php
$client->head('http://orennahum.dyndns.org/');
```
**Debug:**
```php
$this->headers = [
  0 => "HTTP/1.0"
  1 => "Server: Quick 'n Easy Web Server"
  2 => "Connection: Keep-Alive"
  3 => "Content-Length: 0"
  4 => "Set-Cookie: SESSIONID=00000230; path=/;version=1"
  5 => "Date: Wed, 06 Sep 2017 08:06:42 GMT"
]
```

**After explode ofcourse we have only one header with zero offset code:**
```php
$startLine = explode(' ', array_shift($this->headers), 3);
```

**So we need to add checking for this code: **
```php
$this->response = new Response(
            $startLine[1],
            $headers,
            $this->sink,
            substr($startLine[0], 5),
            isset($startLine[2]) ? (string) $startLine[2] : null
        );
```

alexeyshockov:
Seems like your server doesn't follow [HTTP protocol](https://www.w3.org/Protocols/rfc2616/rfc2616-sec6.html), because the status line must include status code.

> 6.1 Status-Line
>
>The first line of a Response message is the Status-Line, consisting of the protocol version followed by a numeric status code and its associated textual phrase, with each element separated by SP characters. No CR or LF is allowed except in the final CRLF sequence.
>
>     Status-Line = HTTP-Version SP Status-Code SP Reason-Phrase CRLF

Because of that it doesn't look like a bug.

Vladnev:
I agree - that server doesn`t follow HTTP protocol.
And first line is invalid.
But this isn`t a occasion for package fall, is it?

alexeyshockov:
It is. There are a lot of variants where a server is able not to follow the spec, so consider all of them in the code is a bad idea, IMO. I mean that you cannot assume that a status line without a code always means `200 OK`.

BTW, the error itself can be improved (throwing a parsing exception instead of PHP's notice).

Vladnev:
About responses not following specification - yes we doesn`t need consider all of troubles - but in api do you check what you recive or you say - it`s ok we agreed.

[...]

PR #1910, opened 2017-09-06
Fix of undifined offset

Vladnev:
Some servers return not valid first header.
Fix according to issue #1909

### DIFF
--- a/src/Handler/EasyHandle.php
+++ b/src/Handler/EasyHandle.php
@@ -74,7 +74,7 @@ final class EasyHandle
 
         // Attach a response to the easy handle with the parsed headers.
         $this->response = new Response(
-            isset($startLine[1]) ? (integer) $startLine[1] : 200,
+            isset($startLine[1]) ? (int) $startLine[1] : 200,
             $headers,
             $this->sink,
             substr($startLine[0], 5),
--- a/src/Handler/EasyHandle.php
+++ b/src/Handler/EasyHandle.php
@@ -74,7 +74,7 @@ final class EasyHandle
 
         // Attach a response to the easy handle with the parsed headers.
         $this->response = new Response(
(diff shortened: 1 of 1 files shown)


src/Handler/EasyHandle.php
```diff
@@ -74,7 +74,7 @@ public function createResponse()
 
         // Attach a response to the easy handle with the parsed headers.
         $this->response = new Response(
-            $startLine[1],
+            isset($startLine[1]) ? (integer) $startLine[1] : 200,
```
GrahamCampbell: Should be `(int)` rather than `(integer)`.

GrahamCampbell: Also, rather than 200, doesn't Guzzle have an http ok constant that could be used?

Vladnev: Fixed to (int) - about const - I did not found any statuses constants.
But I don't sure is better to stay 200 status or maybe we can have some better variants

sagikazarmark: No constants AFAIK, just this list: https://github.com/guzzle/psr7/blob/master/src/Response.php#L15

sagikazarmark:
Sorry for the huge delay.

I don't think it's a good idea to return 200 automatically. That's something one simply cannot assume.

The starting line is a key part of the HTTP message, if it's not valid then I believe we cannot safely return a response. That would produce unreliable behavior, good luck debugging that.

FrontEndCoffee:
I think this should be closed, or a Parsing Exception should be thrown in the case of an invalid first header.

This would be more clear to developers IMHO.

sagikazarmark:
I would be happy to accept a PR implementing the suggested solution. Until then, closing this.

PR #1911, opened 2017-09-08
Fix a docblock of BadResponseException

gmponos:
Since BadResponseException should not be initialized with an empty response it makes sense the get response not to return null.

This helps autocompletion/inspections and tools like PHPStan not to detect it as a possible error.

for Instanche if you had a code like:

``` php
try{
 $client->get('/');
} catch (BadResponseException $e){
   $e->getResponse()->getStatusCode();
}
```

PHPStan was detecting the above as a possible error.

### DIFF
--- a/src/Exception/BadResponseException.php
+++ b/src/Exception/BadResponseException.php
@@ -6,6 +6,8 @@ use Psr\Http\Message\ResponseInterface;
 
 /**
  * Exception when an HTTP error occurs (4xx or 5xx error)
+ * 
+ * @method ResponseInterface getResponse()
  */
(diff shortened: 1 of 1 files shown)


sagikazarmark:
Well, technically it *can* still return null, so is it a good idea to silence this error at this point?

gmponos:
Well this is a weird case. Yes it can still return null but does the guzzle package ever initialize it with having it null?

Furthermore if a developer is initializing this by having it null  and ignoring the deprecated message most probably he is following a very bad practice.

Not sure what can I say.

gmponos:
This would have been a bad addition.

Issue #1886, opened 2017-07-26
Extact applyOptions out of the client class


gmponos:
| Q            | A
| ------------ | ---
| Bug?         | no
| New Feature? | yes
| Version      | Specific version or SHA of a commit

#### Actual Behavior

Hello,

I would like to have a way of converting the RequestOptions of guzzle into a RequestInterface class. At the momemt the function that provides this kind of functionallity is [applyOptions](https://github.com/guzzle/guzzle/blob/master/src/Client.php#L291) which is inside Client class and it's private.

Would you consider extracting this function somewhere else and making it public. Either to a factory class or put it inside functions.php file?

The reason I am making this request is that I need to create a factory that will accept the guzzle options and return a RequestInterface.

So I need

```
Class GuzzleRequestFactory{

    public function createRequest($method, $uri, $options)
    {
         //return RequestInterface here;
    }
}
```

mtdowling:
Yeah, a factory interface would be great. There was a similar issue here: https://github.com/guzzle/guzzle/issues/1455

An interesting consideration would also be if the implementation for this interface could potentially allow for customizable handling of each request option with the ability to add custom options.

PR #1914, opened 2017-09-14
Create a RequestFactory to generate requests using options arrays

duncan3dc:
I recently wrote some code using `Client::createRequest()` before realising that this is not present in Guzzle 6, I then found #1886 and this pr is an attempt to allow `RequestInterface`s to be generated using array options.

The only thing I'm not comfortable with is the reference variable (`$options`) used in `make()`. But I've provided the static method `create()` which provides a version that operates on a copy of `$options`.

Let me know what you think

### DIFF
--- a/src/RequestFactory.php
+++ b/src/RequestFactory.php
@@ -1,0 +1,196 @@
+<?php
+
+namespace GuzzleHttp;
+
+use Psr\Http\Message\RequestInterface;
+
+class RequestFactory implements RequestFactoryInterface
+{
+
+    public static function create($method, $uri = '', array $options = [])
+    {
+        return (new self)->createRequest($method, $uri, $options);
+    }
+
+
+    public function createUri($uri, array $config)
+    {
+        // for BC we accept null which would otherwise fail in uri_for
+        $uri = Psr7\uri_for($uri === null ? '' : $uri);
+
+        if (isset($config['base_uri'])) {
+            $uri = Psr7\UriResolver::resolve(Psr7\uri_for($config['base_uri']), $uri);
+        }
+
+        return $uri->getScheme() === '' && $uri->getHost() !== '' ? $uri->withScheme('http') : $uri;
+    }
+
+
+    public function createRequest($method, $uri = '', array &$options = [])
+    {
+        $headers = isset($options['headers']) ? $options['headers'] : [];
+        $body = isset($options['body']) ? $options['body'] : null;
+        $version = isset($options['version']) ? $options['version'] : '1.1';
+
+        $uri = $this->createUri($uri, $options);
+
+        $request = new Psr7\Request($method, $uri, $headers, $body, $version);
+
+        // Remove the option so that they are not doubly-applied.
+        unset($options['headers'], $options['body'], $options['version']);
+
+        $request = $this->applyOptions($request, $options);
+
+        return $request;
+    }
+
+
+    /**
+     * Applies the array of request options to a request.
+     *
+     * @param RequestInterface $request
+     * @param array            $options
+     *
+     * @return RequestInterface
+     */
+    public function applyOptions(RequestInterface $request, array &$options)
+    {
+        $modify = [
+            'set_headers' => [],
+        ];
+
+        // save_to -> sink
+        if (isset($options['save_to'])) {
+            $options['sink'] = $options['save_to'];
+            unset($options['save_to']);
+        }
+
+        // exceptions -> http_errors
+        if (isset($options['exceptions'])) {
+            $options['http_errors'] = $options['exceptions'];
+            unset($options['exceptions']);
+        }
+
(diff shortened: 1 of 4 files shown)


gmponos:
I have also thought about this but I hadn't submitted any PR. I wanted to do the same tging that you did. The reason that i have not proceed is that this kind of implemantation is incompatible with the PSR that defines the request factories.

duncan3dc:
I wasn't aware that PSR-7 defined a factory interface, I'll take a look at implementing that :+1:

gmponos:
No. It is not PSR7 that defines RequestFactories but another one. I think its on draft now. If that PSR is accepted and guzzle wants to follow that PSR if they accept your PR it will be painfull to move there without a BC. At least that is what I think.

Sorry i am throught my mobile and it is difficult to find and link stuff.

gmponos:
On the other hand if guzzle developers do not want to follow that PSR most probably your PR can be rewiewed.

duncan3dc:
If the PSR hasn't been released yet, and it doesn't look likely it will any time soon, I'd suggest that gets implemented in Guzzle 7, and until then Guzzle uses it's own `RequestFactoryInterface`

mtdowling:
From what I understand, the PSR you are referring to would be very different from Guzzle's feature set. Even if that PSR is ever ratified, Guzzle would need both a Guzzle-specific request factory and a PSR compatible request factory (that would likely not have the same number of features).

src/RequestFactoryInterface.php
```diff
@@ -0,0 +1,53 @@
...
+     * options that are applied to the request are then removed
+     * from the array to avoid them being applied twice.
+     *
+     * @param string $method
+     * @param string $uri
+     * @param array  $options
+     *
+     * @return RequestInterface
+     */
+    public function make($method, $uri = '', array &$options = []);
```
mtdowling: Passing options by reference will probably lead to subtle bugs and unintended mutation.

mtdowling: Can this method name be made more specific? Like createRequest?

src/RequestFactoryInterface.php
```diff
@@ -0,0 +1,53 @@
...
+
+    /**
+     * Applies any URI options to an existing URI to generate a new one.
+     *
+     * @param string $uri
+     * @param array  $config
+     *
+     * @return string
+     */
+    public function buildUri($uri, array $config);
```
mtdowling: ``createUri``?

src/RequestFactory.php
```diff
@@ -0,0 +1,171 @@
+<?php
+
+namespace GuzzleHttp;
+
+use Psr\Http\Message\RequestInterface;
+
+class RequestFactory implements RequestFactoryInterface
+{
+
+    public static function create($method, $uri = '', array $options = [])
```
mtdowling: Given that this factory will likely only be used in the context of a client, this helper method doesn't seem necessary.

gmponos:
I am talking about this one.

https://github.com/php-fig/fig-standards/tree/master/proposed/http-factory

gmponos:
Sorry.. it all comes to mind now.

Personally i would be also happy if guzzle included a RequestFactory and if a PR like this gets merged it will also close an issue of mine (#1886).

When I created the above issue I started investigating the possibility of applying a RequestFactory that it could be compliant to the above PSR.

The way I thought that this could be done was like @duncan3dc did.. passing it as an option.

But if I wanted this to be compliant to the above PSR I needed this PR to me merged first #1890 

The reason is that you can create with the factory a request just by using the URL and the method/verb and then apply options will handle all the rest. The only thing that applyOptions is not handling is the version.

Although you are already applying body and headers in requestAsync I believe this is not needed because applyOptions sets them again. Except if there is a catch why you are doing that.

src/RequestFactory.php
duncan3dc: My use case is creating Request instances to be sent later, so I don't have a client when I'm creating them.
This also provides a "safe" creation method that doesn't involve references

src/RequestFactoryInterface.php
duncan3dc: Sure, I just moved it here from `Client` and kept the name it had there

src/RequestFactoryInterface.php
duncan3dc: I've renamed it now. The tests fail if this isn't a reference, as `Client` relies on the options being mutated

duncan3dc:
I've updated this to work with the latest changes from `master`.

@mtdowling Do you think this is a useful addition to Guzzle? Or would you rather not support this functionality?

Nyholm:
This looks like PSR-17 which is implemented in Guzzle/PSR7. 
(After you made the PR). 

Thank you for the PR, but I will close this as PHP Fig have came around to make a specification for this.

PR #1915, opened 2017-09-15
Empty headers support

### LINKED ISSUE #1882: How to send empty request headers?
Hello. I have a problem with sending empty request headers with guzzle. I think that sending empty http headers is not entirely standard and correct behavior, but for example postman allows this. When I send header with no value in postman, it actually sends it without the value. The same goes for curl in terminal. If you type this:

```
curl -X GET http://example.com -H 'API_KEY;' --head -v
```

it actually sends the empty `API_KEY` header. The request looks like this:

```
GET / HTTP/1.1
Host: example.com
User-Agent: curl/7.47.0
Accept: */*
API_KEY:
```

But of course, you must give it the semicolon, not colon after the header name `API_KEY;`.

The problem is, that I have no idea, if this is possile in guzzle. Is it somehow possible? I have to test my api for empty headers and I have no idea how to test this another way.

It cannot be done this way:

```
$options = [
    'base_uri'    => 'http://example.com',
    'http_errors' => false,
    'debug'        => true,
];
$httpClient = new GuzzleHttp\Client($options);
$httpClient->get('/', ['headers' => ['API_KEY' => '']]);
````

not either this way:

```
$options = [
    'base_uri'    => 'http://example.com',
    'http_errors' => false,
    'debug'        => true,
];
$httpClient = new GuzzleHttp\Client($options);
$httpClient->request('GET', '/', ['headers' => ['API_KEY' => '']]);
```

and not either this way:

```
$options = [
    'base_uri'    => 'http://example.com',
    'http_errors' => false,
    'debug'       => true,
];
$httpClient = new GuzzleHttp\Client($options);
$request = new GuzzleHttp\Psr7\Request('GET', '/', ['API_KEY' => '']);
$httpClient->send($request);
```

All of the above will output this:

```
PHPUnit 6.2.3 by Sebastian Bergmann and contributors.

*   Trying 93.184.216.34...
* Connected to example.com (93.184.216.34) port 80 (#0)
> GET / HTTP/1.1
Host: example.com
User-Agent: GuzzleHttp/6.2.1 curl/7.47.0 PHP/7.1.7-1+ubuntu16.04.1+deb.sury.org+1

< HTTP/1.1 200 OK
< Cache-Control: max-age=604800
< Content-Type: text/html
< Date: Tue, 25 Jul 2017 16:19:26 GMT
< Etag: "359670651+ident"
< Expires: Tue, 01 Aug 2017 16:19:26 GMT
< Last-Modified: Fri, 09 Aug 2013 23:54:35 GMT
< Server: ECS (iad/182A)
< Vary: Accept-Encoding
< X-Cache: HIT
< Content-Length: 1270
< 
* Connection #0 to host example.com left intact
.                                                                   1 / 1 (100%)

Time: 346 ms, Memory: 4.00MB
```

The `API_KEY:` (without the value) is missing in the request.

alexeyshockov:
Fixes #1882

### DIFF
--- a/src/Handler/CurlFactory.php
+++ b/src/Handler/CurlFactory.php
@@ -288,7 +288,14 @@ class CurlFactory implements CurlFactoryInterface
     {
         foreach ($conf['_headers'] as $name => $values) {
             foreach ($values as $value) {
-                $conf[CURLOPT_HTTPHEADER][] = "$name: $value";
+                $value = (string) $value;
+                if ($value === '') {
+                    // cURL requires a special format for empty headers.
+                    // See https://github.com/guzzle/guzzle/issues/1882 for more details.
+                    $conf[CURLOPT_HTTPHEADER][] = "$name;";
+                } else {
+                    $conf[CURLOPT_HTTPHEADER][] = "$name: $value";
+                }
             }
         }
 


sagikazarmark:
Thanks!

For reference: https://stackoverflow.com/questions/32909411/send-empty-http-header-with-libcurl

PR #1916, opened 2017-09-15
Proper service headers modification

### LINKED ISSUE #1913: Multipart `Content-Type` vs. `content-type`
Our code creates a client like this:

```php
$client = new Client([
    'base_uri' => $this->serverUrl,
    'headers' => [
        'content-type' => 'text/plain',
        'x-internal-header' => $value,
    ]
]);
```

Note the lower-case HTTP headers – this shouldn't matter in HTTP/1.1 and is necessary in HTTP/2 (admittedly, we're future-proofing a bit).

It causes this code:

https://github.com/guzzle/guzzle/blob/dfd01d60a38cf7e16b3456d4b1d7c10033b929c0/src/Client.php#L383-L387

to store into `$options['_conditional']` both `content-type` and `Content-Type`:

```
Array
(
    [content-type] => text/plain
    [Content-Type] => multipart/form-data; ...
)
```

This than causes problem here:

https://github.com/guzzle/guzzle/blob/dfd01d60a38cf7e16b3456d4b1d7c10033b929c0/src/Client.php#L393-L397

where only `content-type` (`text/plain`) makes is to the final headers array, breaking the app.

alexeyshockov:
Fixes #1913

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -290,7 +290,14 @@ class Client implements ClientInterface
      */
     private function applyOptions(RequestInterface $request, array &$options)
     {
-        $modify = [];
+        $modify = [
+            'set_headers' => [],
+        ];
+
+        if (isset($options['headers'])) {
+            $modify['set_headers'] = $options['headers'];
+            unset($options['headers']);
+        }
 
         if (isset($options['form_params'])) {
             if (isset($options['multipart'])) {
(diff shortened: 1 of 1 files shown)


sagikazarmark:
Thanks a lot!

CmdrSharp:
This seems to break my unit tests. It seems anything currently sending over JSON formatted information now returns `Undefined index: _conditional`.

PR #1920, opened 2017-09-19
I think before excecute retryMiddleware  rewind request Body will be …

wadee:
I think rewind the request Body before excecute retryMiddleware  in library will be a good choice, otherwise it will take a bunch time for user to realise should rewind request body in the retryMW ...


zuoRambo:
vote

Nyholm:
One should never assume the position of a stream.

Issue #1127, opened 2015-06-16
Asynchronous requests with requestAsync() are not actually sent until you call  ->wait()


drscre:
I expected that using `$client->getAsync()` with Curl will fire request immediately, leaving me some time to perform additional work before it completes:

```php
$req = $client->getAsync('http://api.com/get_a_lot_of_data'); // Shouldn't request be sent here ? ...

// do something while the long request is in progress in order not to waste time
do_work();

$req->then(...);
$req->wait(); // ... so here request can already complete
```

But it turns out, that request will actually be fired only when you call `$req->wait()`, acting like a plain synchronous request.

Briareos:
You can check out example no.2 in @mtdowling's response: https://github.com/aws/aws-sdk-php/issues/621#issuecomment-111232418

The key is to invoke `tick` manually instead of relying on `wait` to do so for you. That `Promise::wait` is basically a proxy to `CurlMultiHandler::wait`.

You can call `CurlMultiHandler::tick` once when you enqueue requests via `xxxAsync()`, it won't block. That's because `curl_multi_select` is called first; before any calls to `curl_multi_exec`, that actually fires requests.

To wait for all requests to complete you can use this from the mentioned example:
```php
while (!Promise\is_settled($requests)) {
    $curlMultiHandle->tick();
}
```

drscre:
@Briareos Thanks for explicit `tick()` hint!

Still, I don't quite get whether it's a bug or is it a normal way the things should work?
Why not make `tick()` to get called automatically from `xxxAsync()`?

kamilsk:
> it's a bug or is it a normal way
I think it's a normal way.
> Why not make tick() to get called automatically from xxxAsync()
How you can use `promises` in this case?

mtdowling:
It works this way because calling tick when starting a promise could have the side effect of resolving other promises and resolving them. This could have pretty bad unforeseen consequences-- most of which revolve around unwanted recursion. Take the following example:

You are building a rate-limited pool of concurrent promises that yields promises from a generator. When a promise completes, the abstraction calls next on the generator, sends the request, and adds a promise to the pool. If initiating a promise in cURL caused other promises to complete, then it would cause the abstraction to attempt to call next on the generator while it's still yielding the previous value. This causes a strange sort of recursion in PHP generators and will trigger a fatal error.

You should either use the tick approach I mentioned in the comment linked by @Briareos or use a different adapter that hooks into an event loop in which your whole application is run asynchronously.

[...]

PR #1924, opened 2017-09-22
Fixed CurlMultiHandler Blocking Issue

shtse8:
I am trying to fix this issue https://github.com/guzzle/guzzle/issues/1127

1. Tick() should not block and should not sleep, it should return immediately without waiting.
2. select_timeout changed to 0. return immediately if no message. wait for next tick.
3. if select returns 0 or -1. do nothing and wait for next tick. no need to wait inside for -1. we should wait for next tick.

### DIFF
--- a/src/Handler/CurlMultiHandler.php
+++ b/src/Handler/CurlMultiHandler.php
@@ -38,7 +38,7 @@ class CurlMultiHandler
         $this->factory = isset($options['handle_factory'])
             ? $options['handle_factory'] : new CurlFactory(50);
         $this->selectTimeout = isset($options['select_timeout'])
-            ? $options['select_timeout'] : 1;
+            ? $options['select_timeout'] : 0;
     }
 
     public function __get($name)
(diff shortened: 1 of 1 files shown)


shtse8:
For those who are using Curl 7.47.0 with proxy and multi.
Please upgrade to Curl 7.55.1 with following commands:
```shell
sudo add-apt-repository ppa:jonathonf/curl
sudo apt-get update
```


https://curl.haxx.se/changes.html
image

src/Handler/CurlMultiHandler.php
```diff
@@ -95,17 +95,11 @@ public function tick()
...
         P\queue()->run();
 
-        if ($this->active &&
-            curl_multi_select($this->_mh, $this->selectTimeout) === -1
-        ) {
-            // Perform a usleep if a select returns -1.
-            // See: https://bugs.php.net/bug.php?id=61141
-            usleep(250);
+        $cms = curl_multi_select($this->_mh, $this->selectTimeout);
+        if ($cms != 0) {
```
gregurco: Wrong comparison. Read docs: http://php.net/manual/en/function.curl-multi-select.php
```
On failure, this function will return -1
```
So, in this case you will do same action when it's failure and success. Also variable name is not clear. I think `$numberOfDescriptors` will be more cleaer. Finally it's better to write `$numberOfDescriptors >= 0`

mtdowling:
This will still block, and I think this change will cause 100% CPU utilization.

shtse8:
@mtdowling I have tested and my project is running without any blocking and CPU issue for 3 months already.

PR #1933, opened 2017-10-05
Do-while mistake in CurlMultiHandler.php ?

iorsa:
Do-while mistake that can cause problems with curl on some systems.
Am I correct or it was fine?

### DIFF
--- a/src/Handler/CurlMultiHandler.php
+++ b/src/Handler/CurlMultiHandler.php
@@ -95,15 +95,15 @@ class CurlMultiHandler
         // Step through the task queue which may add additional requests.
         P\queue()->run();
 
-        if ($this->active &&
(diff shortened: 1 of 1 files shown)


jhdxr:
No, You are wrong.

https://bugs.php.net/bug.php?id=61141 says
> The libcurl documentations says "When max_fd returns 
with -1, you need to wait a while and **then proceed and call curl_multi_perform 
anyway**".

and the [original code](https://github.com/guzzle/guzzle/pull/1933/files#diff-392a8065a86ef320fdf3763730ce850eL106) means `curl_multi_exec` will keep executing when it returns `CURLM_CALL_MULTI_PERFORM`, but you code changes it to call `curl_multi_select` during the loop, which is absolutely wrong.

Issue #1938, opened 2017-10-17
cURL is not used if "allow_url_fopen" is set


arendvw:
| Q            | A
| ------------ | ---
| Bug?         | yes
| New Feature? | no
| Version      |  9e28b54996d260184abc9c36c59e2740be2184a4

#### Actual Behavior

When creating a new guzzle client without defining a custom handler, guzzle will use the stream wrapper by default if allow_url_fopen is defined. 

#### Expected Behavior

In my opinion Guzzle should always choose cURL als it's handler when available.

#### Steps to Reproduce

1. Enable allow_url_fopen in the php config.
2. Enable the curl plugin
3. Test the default handler in this manner:
```
use GuzzleHttp\Client;
$client = new Client();
var_dump(get_class($client->getConfig('handler')));
```

returns:
string(23) "GuzzleHttp\HandlerStack"

Which in my opinion should be 
// string(35) "GuzzleHttp\Handler\CurlMultiHandler"

#### Possible Solutions

See pull request. In the function choose_handler(), src/functions.php, line 113, first cURL is detected. But if allow_url_fopen is defined, the curl handler is then overwritten with the stream wrapper. Changing the "if" on line 114 to else if, should fix this behaviour.

alexeyshockov:
Please, check the `HandlerStack` object, it should contain `CurlMultiHandler` in the end. The stack is there to collect all the middlewares and run them in a chain. The handler is just the last element in this stack.

Guzzle's behaviour is equal to what you described. If cURL is available, it's selected as a handler. Check [`choose_handler()`](https://github.com/guzzle/guzzle/blob/master/src/functions.php#L103) for proofs.

Nokim:
If you look Proxy code more in detail, you'll see that wrapper will be used only if 'stream' option set, otherwise chosen before handler will be used (like CurlMultiHandler) or StreamHandler if none was chosen.

sagikazarmark:
Yup, see the discussion in the linked issue.

sagikazarmark:
@arendvw is your issue resolved?

arendvw:
@sagikazarmark Yes, my apologies for not responding - this issue was indeed resolved with your comments, and this issue and pull request should be closed.

PR #1939, opened 2017-10-17
fix for issue #1938, cURL handler is overwritten if allow_url_fopen is set

arendvw:
See https://github.com/guzzle/guzzle/issues/1938


sagikazarmark:
This should only be a problem, if you send requests with the `stream` option being set to true, which is actually the desired behaviour (curl needs to download everything first):

https://github.com/guzzle/guzzle/blob/master/src/Handler/Proxy.php#L49

PR #1953, opened 2017-11-10
Use PHPUnit\Framework\TestCase instead of PHPUnit_Framework_TestCase

carusogabriel:
I use the `PHPUnit\Framework\TestCase` notation instead of `PHPUnit_Framework_TestCase` while extending our TestCases. This will help us migrate to PHPUnit 6, that [no longer support snake case class names](https://github.com/sebastianbergmann/phpunit/blob/master/ChangeLog-6.0.md#changed-1).

I just need to bump PHPUnit version to [`^4.8.35`](https://github.com/sebastianbergmann/phpunit/blob/master/ChangeLog-4.8.md#4835---2017-02-06) and [`^5.7`](https://github.com/sebastianbergmann/phpunit/blob/master/ChangeLog-5.7.md#570---2016-12-02), that support this namespace.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
     },
     "require-dev": {
         "ext-curl": "*",
-        "phpunit/phpunit": "^4.0 || ^5.0",
+        "phpunit/phpunit": "^4.8.35 || ^5.7",
         "psr/log": "^1.0"
     },
     "autoload": {
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
     },
     "require-dev": {
         "ext-curl": "*",
-        "phpunit/phpunit": "^4.8.35 || ^5.7",
+        "phpunit/phpunit": "^4.8.35 || ^5.7 || ^6.4",
         "psr/log": "^1.0"
     },
     "autoload": {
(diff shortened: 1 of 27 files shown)


sagikazarmark:
Thanks a lot for your PR. I believe `^4.8.35 || ^5.7 || ^6.0` should be OK as well.

Can you check why the tests are broken?

gabriel-caruso:
@sagikazarmark I'll work on that

gabriel-caruso:
@sagikazarmark Tests fixed. You can now merge this PR. I'll work on `PHPUnit 6` in another branch :construction:

TomasVotruba:
Just fyi, there is no reason to support 4.0 version, this is  dev dependency. Thus not installed anywhere but in this project.

sagikazarmark:
@carusogabriel many thanks for your work!

@TomasVotruba we still officially support PHP 5.5 and as of PHPUnit 5.0 the minimum PHP requirement is 5.6. So we still need PHPUnit 4.x for a while...but hopefully, not so long.

TomasVotruba:
:+1:

Issue #1257, opened 2015-09-14
ClientInterface::request throws GuzzleException not really useful


ged15:
Hi!
First of all, thanks for all the amazing work here! Really advanced stuff.
I have a question regarding exception doc-blocking. I.e. the [ClientInterface::request method](https://github.com/guzzle/guzzle/blob/master/src/ClientInterface.php#L52) claims to throw a `GuzzleException`. This is a bit unhelpful as the `GuzzleException` does not have any public methods.
Is it correct to change the docblock to `@throws GuzzleHttp\Exception\TransferException`? If so, I cna send a PR?

mtdowling:
Neither classes/interfaces, TransferException or GuzzleException, have public methods. You may be seeing that GuzzleException is an interface that does not extend `\Exception`. Changing the public API of ClientInterface at this point would be a breaking change, so I don't think that can happen.

Maybe a better pull request would be to add `@method` annotations corresponding to methods of PHP's `\Exception` in order help IDEs autocomplete PHP `\Exception` methods.

nesl247:
Is there any chance of this happening in a future release? PhpStorm recently added a feature to detect unhandled exceptions. Unfortunately it does not work well with this case, and EVERY `request()` call results in PhpStorm showing that there is a (generic - GuzzleException) unhandled exception.

The issue here is that we don't want to catch `GuzzleException`, because we let it bubble up to our exception listeners.

PR #1957, opened 2017-11-20
Add annotations for exception methods

jyggen:
Adds `@method` annotations to `GuzzleException`, as suggested in #1257, in order to help IDEs and static analysis tools.


sagikazarmark:
Thanks a lot @jyggen

PR #1961, opened 2017-11-24
Logger context for the log middleware

scuben:
This PR adds the possibility to add a (static) context to the log message created by the guzzle middleware.

### DIFF
--- a/src/Middleware.php
+++ b/src/Middleware.php
@@ -180,25 +180,26 @@ final class Middleware
      * @param LoggerInterface  $logger Logs messages.
      * @param MessageFormatter $formatter Formatter used to create message strings.
      * @param string           $logLevel Level at which to log requests.
+     * @param array            $context Context to log requests.
      *
      * @return callable Returns a function that accepts the next handler.
      */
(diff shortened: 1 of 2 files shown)


gregurco:
@scuben what about failure case? why you don't use context there? On line 201.

scuben:
@gregurco I added a test case covering the new implementation and added the context to the failure case.

scuben:
Guess this is ready to merge, isn't it?

sagikazarmark:
Thinking about this: I believe a better solution would be a decorator which adds default context to the log events.

Something like this:

```php
<?php

final class DefaultContextLogger implements \Psr\Log\LoggerInterface
{
    use \Psr\Log\LoggerTrait;

    private $logger;
    private $context = [];

    public function __construct(\Psr\Log\LoggerInterface $logger, array $context = [])
    {
        $this->logger = $logger;
        $this->context = $context;
    }

    public function log($level, $message, array $context = array())
    {
        $this->logger->log($level, $message, $this->context+$context)
    }
}
```

I would prefer keeping things simple.

sagikazarmark:
Please refer to the solution above.

PR #1974, opened 2017-12-08
Fix documentation using the getHeader method

### LINKED ISSUE #1972: Wrong doc info
| Q            | A
| ------------ | ---
| Bug?         | yes
| New Feature? | no
| Version      | 6.3.0

On page http://docs.guzzlephp.org/en/stable/ (http://joxi.ru/krDJ9yBU06l8W2)
echo $res->getHeader('content-type'); // error - array to string conversion

tomzx:
Add missing 0 index to documentation making use of the getHeader method.
The getHeader method always returns an array, but it may be empty.
For documentation purpose, we assume that the response object always contains the queried header, thus providing at least one element in the array at index 0.

Fixes #1972.

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -159,8 +159,8 @@ together in a single report:
     $response = $client->request('GET', $initialRequest); // Make your request
 
     // Retrieve both Redirect History headers
-    $redirectUriHistory = $response->getHeader('X-Guzzle-Redirect-History'); // retrieve Redirect URI history
-    $redirectCodeHistory = $response->getHeader('X-Guzzle-Redirect-Status-History'); // retrieve Redirect HTTP Status history
+    $redirectUriHistory = $response->getHeader('X-Guzzle-Redirect-History')[0]; // retrieve Redirect URI history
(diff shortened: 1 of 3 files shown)


gregurco:
@tomzx maybe it's better to use `getHeaderLine` instead of `getHeader()[0]`. What do you think?

tomzx:
@gregurco Not entirely sure, since `getHeaderLine` will merge a multi-value arrays using semi-colons as separators.

gregurco:
@tomzx yep, but if there is only one value, then this value will be returned, if header doesn't exist then empty string will be returned. But in your case you will call inexistent index.

tomzx:
@gregurco True, but given some other parts of the documentation already used the zero indexing I thought it would be better to stay consistent.

Should then the other examples using `getHeader` with the 0 index also be replaced with `getHeaderLine`?

sagikazarmark:
Thanks @tomzx for the fix and @gregurco for the comments. I think it would be better to discuss the getHeaderLine question in a separate issue/PR, so I will just merge this one.

PR #1976, opened 2017-12-13
bug fix

keita5500:
fix esacaping delimiter


sagikazarmark:
Thanks, this will be fixed in #1999

PR #1986, opened 2018-01-05
refactoring regarding psr1/psr2 using php-cs-fixer

bjoernffm:
Light changes regarding PSR-1, PSR-2. Tests successfully executed


sagikazarmark:
Thanks a lot!

PR #1991, opened 2018-01-23
Allow cookies with an empty string as value

### LINKED ISSUE #1975: Empty cookie value is not handled
| Q            | A
| ------------ | ---
| Bug?         | yes
| New Feature? | no
| Version      | 6.3.0

#### Actual Behavior

Many websites (ex: github.com) set cookie without value on sign out. With cookiejar strict mode set to true you have an RuntimeException. With strict mode set to false, the cookie is ignored.

#### Expected Behavior

Cookie value should be an empty string.

#### Steps to Reproduce

```php
$header = 'user_session=; path=/; max-age=0; expires=Thu, 01 Jan 1970 00:00:00 -0000';

$req = new \GuzzleHttp\Psr7\Request('GET', new Uri('https://github.com'));
$res = new \GuzzleHttp\Psr7\Response(200, ['Set-Cookie' => $header]);

$cj = new \GuzzleHttp\Cookie\CookieJar(true); 
$cj->extractCookies($req, $res);  // RuntimeException: Invalid cookie: The cookie value must not be empty

var_dump($cj->getCookieByName('user_session'));
```

HanXHX:
Fixes #1975

### DIFF
--- a/src/Cookie/SetCookie.php
+++ b/src/Cookie/SetCookie.php
@@ -384,9 +384,9 @@ class SetCookie
                 . 'following characters: ()<>@,;:\"/?={}';
         }
 
-        // Value must not be empty, but can be 0
+        // Value must not be empty, but can be 0 or an empty string
         $value = $this->getValue();
-        if (empty($value) && !is_numeric($value)) {
+        if (empty($value) && !is_numeric($value) && $value !== '') {
             return 'The cookie value must not be empty';
         }
 
--- a/tests/Cookie/SetCookieTest.php
+++ b/tests/Cookie/SetCookieTest.php
@@ -152,10 +152,11 @@ class SetCookieTest extends \PHPUnit_Framework_TestCase
     {
         return array(
             array('foo', 'baz', 'bar', true),
+            array('foo', '', 'bar', true),
             array('0', '0', '0', true),
             array('foo[bar]', 'baz', 'bar', true),
             array('', 'baz', 'bar', 'The cookie name must not be empty'),
-            array('foo', '', 'bar', 'The cookie value must not be empty'),
+            array('foo', null, 'bar', 'The cookie value must not be empty'),
             array('foo', 'baz', '', 'The cookie domain must not be empty'),
             array("foo\r", 'baz', '0', 'Cookie name must not contain invalid characters: ASCII Control characters (0-31;127), space, tab and the following characters: ()<>@,;:\"/?={}'),
         );


HanXHX:
ping @sagikazarmark

dbollaer:
@HanXHX  thank you for this fix. I was running against the same problem.

KrlWil:
+1

peckrob:
Any change this could be merged? I am running into this today.

sagikazarmark:
It needs to be rebased first.

src/Cookie/SetCookie.php
```diff
@@ -384,9 +384,9 @@ public function validate()
                 . 'following characters: ()<>@,;:\"/?={}';
         }
 
-        // Value must not be empty, but can be 0
+        // Value must not be empty, but can be 0 or an empty string
         $value = $this->getValue();
-        if (empty($value) && !is_numeric($value)) {
+        if (empty($value) && !is_numeric($value) && $value !== '') {
```
Nyholm: So "0" will still be invalid. 

I suggest modifying this if-statement to: `if ($value === null)`

Nyholm:
Friendly ping

Nyholm:

HanXHX:
I did this PR more than 2 years ago for 4 lines!

I ping team for 1 year in #1975... I had no news.

You have the base code and the issue. Feel free to improve yourself this PR.

GrahamCampbell:
Empty cookie values are not allowed by RFC 6265 anyway?

Nyholm:
Closing this since nobody wants to work with this PR and because we think that empty values are not allowed by RFC 6265.

If empty values are supported and someone wants to work on this, then feel free to open a new PR.

peckrob:
@Nyholm @GrahamCampbell  There are a lot of things out there that are technically against RFCs but nonetheless happen. In my case I was dealing with a SAP application that was sending empty strings. Moreover all the browsers I tested handed it fine, so I would expect guzzle to as well.

I was able to get around it with a TREMENDOUS amount of additional code. I can’t remember all of that was needed because this was the pre Corona and feels like a lifetime ago, a but it involved subclassing a bunch of parts of guzzle so I could get there 4 lines in.

In the only issue is that nobody wants to work it, I will take it over later today and resubmit with the requested fix. But if it’s just going to be declined because of the RFC, let me know so that I don’t waste my time.

Nyholm:
@peckrob That would be great. 

We will schedule this feature for 7.1.

peckrob:
Opened #2691.

GrahamCampbell:
Someone should report this problem to GitHub, since this is a bug in their server implementation.

GrahamCampbell:
> here are a lot of things out there that are technically against RFCs but nonetheless happen.

I mean, I would treat this as a bug. If its not allowed by the spec, then formally Guzzle doesn't know how to handle it.

GrahamCampbell:
There is a similar PR with empty location header for 30x requests. Someone is proposing Guzzle be strict. We can't have it both ways. Either we want this PR and not the other, or vice versa. We can't have them both.

GrahamCampbell:
Otherwise we have contradictory statements about Guzzle's approach to RFCs. The other PR makes Guzzle stricter, and this one (and its replacement) makes it weaker.

peckrob:
Guzzle's internal matters or other pull requests are things I will allow the team members to decide. But I will again reiterate that **browsers handle this correctly**, the way a user would expect them to, even if _correct_ is technically against the RFC.

We have here at least two examples of "broken" non-RFC-compliant implementations: one on one of the largest sites on the Internet (Github), another in a software package (SAP) from a multi-billion dollar behemoth used in who knows how many places (in my specific case, my local utility board.) Who knows how many others are out there? And good luck getting either of those fixed; Github maybe, but SAP? Unless you have a contract with them, they're not going to do a thing to help you and, even if they did, it might be years before that filters down to all of the thousands of places that use their software.

So it seems to me that the options are either fix guzzle to handle this edge case with this ~4 lines of code, or fix every broken implementation on the Internet that might be encountered. Seems pretty clear to me, but I'm just a user.

If this is not going to be merged, it should **at the very minimum** be documented that this is a known issue that can occur, and provide a workaround.

GrahamCampbell:
> browsers handle this correctly, the way a user would expect them to, even if correct is technically against the RFC.

The RFC is the definition of being correct, so browsers implement behaviour outside of the spec. Perhaps there is a replacement RFC out there, or perhaps we need need to delete all the comments that refer to the spec, if we are not interested in following it.

gmponos:
Since we talk about specs I though that it might worth dropping this here: https://github.com/symfony/symfony/pull/36823

> This also relaxes the accepted status codes, with https://www.linkedin.com/company/linkedin/ as an example that returns a non-conformant one (999)

GrahamCampbell:
Right, but symfony still treats it as an error, rather than success.

gmponos:
I was pointing at what linkedin does.. not what sf did..

PR #1992, opened 2018-01-23
Use assertEmpty

carusogabriel:


sagikazarmark:
Thanks a lot @carusogabriel, but your change doesn't work. The proxy key can actually be missing, which is properly handled by empty, but not by the assertEmpty call.

carusogabriel:
@sagikazarmark Thanks for your feedback. Maybe I’ll propose some improvements to PHPUnit after that.

sagikazarmark:
@carusogabriel I'm afraid there is not much that can be done here. The direct access of a non-existent key won't work anywhere outside an empty or an isset call.

PR #1998, opened 2018-01-31
A better fix for #1894 - "Undefined offset: 0"

curtisdf:
There's a bug in `SetCookie::fromString()` which throws an "Undefined offset: 0" error if the inputted `$cookie` string is missing its first key-value pair, but has at least one other pair. For example, it would be triggered with the following (malformed) cookie:

```
; foo
```

Another author submitted a pull request for this (https://github.com/guzzle/guzzle/pull/1894), but the request was rejected due to improper test writing.  However, I don't believe the fix itself is appropriate, as I documented there.

So I am submitting my own pull request.  The change is trivial.  It makes the function return an empty cookie in response to this input, which is the same behavior as this function already has when the first key-value pair is missing the cookie name.

### DIFF
--- a/src/Cookie/SetCookie.php
+++ b/src/Cookie/SetCookie.php
@@ -35,8 +35,8 @@ class SetCookie
         $data = self::$defaults;
         // Explode the cookie string using a series of semicolons
         $pieces = array_filter(array_map('trim', explode(';', $cookie)));
-        // The name of the cookie (first kvp) must include an equal sign.
-        if (empty($pieces) || !strpos($pieces[0], '=')) {
+        // The name of the cookie (first kvp) must exist and include an equal sign.
+        if (empty($pieces[0]) || !strpos($pieces[0], '=')) {
             return new self($data);
         }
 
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
I think https://github.com/guzzle/guzzle/pull/1999 superseeds this, right?

curtisdf:
@GrahamCampbell It's two different fixes in the code, and I put them in different branches of my own tree.  But I'm confused by how GitHub handles this situation.  If the code for both fixes is included in the pull request for #1999 then yeah, I guess it would supersede it.  Otherwise both pull requests might need to be accepted separately.

Thanks!

sagikazarmark:
Thanks a lot for this fix and apologies for the delay.

As for the PR confusion: looks like you opened this PR from your master branch and the other branch (malformed-domain) obviously originates from it. Although I usually don't like pulling from master, merging this then the other PR should work.

curtisdf:
Thanks so much @sagikazarmark.  In the future I guess I'll use a custom branch for any PRs rather than the master branch then. Thanks for the clarification.

Issue #1720, opened 2017-01-09
 Warning: preg_match(): Unknown modifier '/' in SetCookie.php on line 352


turkelt0n:
Hi! I've got some Error using a simple Request 

```
$values["region"]` = "12345_DE";
$cookieJar = CookieJar::fromArray(
    $values,
    "https://www.allyouneedfresh.de"
);

$client = new \GuzzleHttp\Client(['cookies' => $cookieJar]);
$response = $client->get(
    "https://www.allyouneedfresh.de"
);
```
This results in

```
Warning: preg_match(): Unknown modifier '/'
() at /home/alex/foodle/vendor/guzzlehttp/guzzle/src/Cookie/SetCookie.php:352
 GuzzleHttp\Cookie\SetCookie->matchesDomain() at /home/foodle/vendor/guzzlehttp/guzzle/src/Cookie/CookieJar.php:234
...
```
The Error is thrown for every URL i tried.  Any Suggestions? Did I miss something?

Thank you!

sagikazarmark:
The cookie jar domain must be without the schema part, so:

```
$cookieJar = CookieJar::fromArray(
    $values,
    "www.allyouneedfresh.de"
);
```

turkelt0n:
Thank you for that!

curtisdf:
I know this is from a year ago, but the discussion above does not address the underlying bug in the code.  There are servers on the Interwebs that are sending invalid `domain` strings, which wreaks havoc with Guzzle when trying to parse those  cookies.

I have submitted a pull request which fixes the bug in `SetCookie::matchesDomain()`. See https://github.com/guzzle/guzzle/pull/1999.

Issue #526, opened 2014-01-10
Warning: preg_match(): Unknown modifier '/' in Cookie.php on line 440


Thinkscape:
Getting those recently:

```
Warning: preg_match(): Unknown modifier '/' in /app/guzzle/src/Guzzle/Plugin/Cookie/Cookie.php on line 440
```

Probably because the remote server is sending something stupid as the cookie domain. There is some normalization going on in the plugin, though the error means that something is getting through regardless. We might `filter_var()` the `$cookieDomain` against domain name, or something like that...

curtisdf:
I know it's been 4+ years, but this bug has reappeared.  I just submitted a pull request which reapplies the same fix as referenced above. See https://github.com/guzzle/guzzle/pull/1999.

PR #1999, opened 2018-01-31
Error with malformed domain that contains a "/" character

curtisdf:
Cookies where the domain contains a "/" character are triggering an exception with `SetCookie::matchesDomain()`.  The issue is that the call to `preg_quote()` inside this function does not specify the regex delimiter being used, and so this character isn't getting escaped, which leads to a malformed pattern going into `preg_match()`. This can be triggered for example if the cookie improperly contains a URL, or if it appends a path after the hostname, such as in the following examples:

```
foo=bar; domain=example.com/
```
```
foo=bar; domain=http://www.example.com
```

### DIFF
--- a/src/Cookie/SetCookie.php
+++ b/src/Cookie/SetCookie.php
@@ -348,7 +348,7 @@ class SetCookie
             return false;
         }
 
-        return (bool) preg_match('/\.' . preg_quote($cookieDomain) . '$/', $domain);
+        return (bool) preg_match('/\.' . preg_quote($cookieDomain, '/') . '$/', $domain);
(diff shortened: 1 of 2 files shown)


curtisdf:
This is interesting.  Evidently this bug was fixed 4 years ago, but somehow the fix was reverted.  See https://github.com/guzzle/guzzle/issues/526.

Also, https://github.com/guzzle/guzzle/issues/1720 is in reference to the same bug.

GrahamCampbell:
👍

sagikazarmark:
Thanks!

PR #2012, opened 2018-02-14
Add `failed to open stream` to ConnectException

sm2017:
When `connect_timeout` is reached I see the following error but `isConnectionError` is `false`

>Error creating resource: [message] fopen(url): failed to open stream: A connection attempt failed because the connected party did not properly respond after a period of time, or established connection failed because connected host has failed to respond.

### DIFF
--- a/src/Handler/StreamHandler.php
+++ b/src/Handler/StreamHandler.php
@@ -60,7 +60,7 @@ class StreamHandler
             if (strpos($message, 'getaddrinfo') // DNS lookup failed
                 || strpos($message, 'Connection refused')
                 || strpos($message, "couldn't connect to host") // error on HHVM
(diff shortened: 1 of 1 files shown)


kstich:
`failed to open stream` can happen for a multitude of reasons. It seems pretty explicitly safe to use `connection attempt failed` as the determining factor here.

sagikazarmark:
@kstich sounds better

sm2017:
@kstich thank you

sm2017:
@guzzle Can you accept the PR?

sagikazarmark:
Will be reviewed when someone gets there, but I wouldn't expect it to be released for another few weeks. There are some other changes in the pipe too.

sagikazarmark:
Thanks

Issue #1952, opened 2017-11-09
Cookies with a 0 epoch expiry aren't treated as expired


neerolyte:
| Q            | A
| ------------ | ---
| Bug?         | yes
| New Feature? | no
| Version      | 6.3

#### Actual Behavior

What is the actual behavior?

Cookies with an expiry of `Thu, 01 Jan 1970 00:00:00 GMT` are reporting `isExpired()` as false.

#### Expected Behavior

What is the behavior you expect?

They should report as expired.

#### Steps to Reproduce

Run this code:

```php
<?php
require_once('vendor/autoload.php');
use GuzzleHttp\Cookie\SetCookie;
echo "0: ".var_export(SetCookie::fromString('SESSION=destroy; expires=Thu, 01 Jan 1970 00:00:00 GMT;')->isExpired(), true)."\n";
echo "1: ".var_export(SetCookie::fromString('SESSION=destroy; expires=Thu, 01 Jan 1970 00:00:01 GMT;')->isExpired(), true)."\n";
```

You'll get:

```
$ php foo.php
0: false <--- should be true
1: true
```

#### Possible Solutions

The current check in `SetCookie::isExpired()` is:

```php
return $this->getExpires() && time() > $this->getExpires();
```

I believe it should check `getExpires()` for null rather than just implicitly casting it to bool, e.g:

```php
return $this->getExpires() !== null && time() > $this->getExpires();
```

neerolyte:
I've worked around this by avoiding the `use GuzzleHttp\Cookie\CookieJar` and instead overloading it with a check that adjusts 0 values to 1 (which will still always be enough in the past to be immediately expired):

```php
class CookieJar extends GuzzleHttp\Cookie\CookieJar {
	public function setCookie(GuzzleHttp\Cookie\SetCookie $cookie) {
		if ($cookie->getExpires() === 0) {
			$cookie->setExpires(1);
		}
		return parent::setCookie($cookie);
	}
}
```

neerolyte:
Is this project still active?

sagikazarmark:
If you have a proper solution PRs are always welcome.

neerolyte:
Ok, I've added a PR to master with the solution I suggested in the description.

The PR is against master, but this is a problem in the 5.3 release too - do you want a PR against that branch as well?

sagikazarmark:
Yes both the PR and the bug report are great, sorry for the delay. Patch release will be out in a few days.

[...]

PR #2014, opened 2018-02-16
Fix bug parsing 0 epoch expiry times

neerolyte:
As per https://github.com/guzzle/guzzle/issues/1952 expiry times
specified as `Thu, 01 Jan 1970 00:00:00 GMT` should be treated as
expired, but they weren't because 0 was implicitly casting to a bool in
`SetCookie::isExpired()`.

### DIFF
--- a/src/Cookie/SetCookie.php
+++ b/src/Cookie/SetCookie.php
@@ -358,7 +358,7 @@ class SetCookie
      */
     public function isExpired()
     {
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
👍

sagikazarmark:
Great patch, thanks a lot and sorry for the delay.

PR #2035, opened 2018-03-29
Update Middleware::history() docblock to note $container type

webignition:
`Middleware::history()` accepts as its single argument an array or an object implementing `\ArrayAccess`.

The docblock for this method currently states the type for the argument as: `@param array $container`. The current docblock lacks the `\ArrayAccess` type.

When passing in an object that implements `\ArrayAccess`, PHPStorm highlights to me that, due to the docblock, the wrong type is being passed. This is due to the docblock being incorrect.

This change updates the docblock for `\Middleware::history()` to correctly reflect the type for the argument.

### DIFF
--- a/src/Middleware.php
+++ b/src/Middleware.php
@@ -73,7 +73,7 @@ final class Middleware
     /**
      * Middleware that pushes history data to an ArrayAccess container.
      *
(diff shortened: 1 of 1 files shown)


sagikazarmark:
Thanks!

PR #2039, opened 2018-04-03
Fix bug parsing 0 epoch expiry times

neerolyte:
As per https://github.com/guzzle/guzzle/issues/1952 expiry times
specified as "Thu, 01 Jan 1970 00:00:00 GMT" should be treated as
expired, but they weren't because 0 was implicitly casting to a bool in
SetCookie::isExpired().

5.3 specific PR as per https://github.com/guzzle/guzzle/issues/1952#issuecomment-376133950

### DIFF
--- a/src/Cookie/SetCookie.php
+++ b/src/Cookie/SetCookie.php
@@ -333,7 +333,7 @@ class SetCookie implements ToArrayInterface
      */
     public function isExpired()
     {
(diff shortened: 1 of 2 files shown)


sagikazarmark:
Cool, thanks

sagikazarmark:
Thanks a lot!

PR #2040, opened 2018-04-05
Use safer JSON.parse() instead of eval()

### LINKED ISSUE #2004: Why eval in test server
I did a little search in my repo to find security issue, then I found this backdoor-like code:

https://github.com/guzzle/guzzle/blob/master/tests/server.js#L178

```
that.responses = eval('(' + request.body + ')');
```

Why eval here? To parse JSON format in node.js server side, just use `JSON.parse` instead.

Since it's in test code and will not be shipped into production, I directly report it as a bug here.

brbrowning21:
Closes #2004

### DIFF
--- a/tests/server.js
+++ b/tests/server.js
@@ -175,7 +175,7 @@ var GuzzleServer = function(port, log) {
         }
         res.writeHead(400, 'NO RESPONSES IN REQUEST', { 'Content-Length': 0 });
       } else {
-        that.responses = eval('(' + request.body + ')');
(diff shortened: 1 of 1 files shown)


wirwolf:
Hello @brbrowning21, i create a fork of this repo and release this changes into the new version(6.4.0).

Namespace compatibility is 100%, and you can use my changes
* add `"guzzlehttp/guzzle": "6.3.*"` into replace block in your composer.json

PR #2045, opened 2018-04-08
DX: small CS fixes

keradus:


keradus:
thanks for merging !
would you be fine to run PHP CS Fixer on CI level ? I could propose a PR for that

sagikazarmark:
It's on my todo list, but there are things that should be done before.

keradus:
can I help ?

Issue #2046, opened 2018-04-10
HEAD request with 301 response results in GET request when following redirect


webignition:
| Q            | A
| ------------ | ---
| Bug?         | yes
| New Feature? | no
| Version      | 6.3.2

#### Actual Behavior

- send HEAD request
- receive 301 redirect response
- follow redirect with GET request

#### Expected Behavior

- send HEAD request
- receive 301 redirect response
- follow redirect with HEAD request

cURL will follow a redirect response to a HEAD request with another HEAD request. I think it is fair to assume that if cURL behaves in this manner then so should guzzle.

#### Steps to Reproduce

```php
<?php
use GuzzleHttp\Client;
use GuzzleHttp\HandlerStack;
use GuzzleHttp\Middleware;
use GuzzleHttp\Psr7\Request;

$historyContainer = [];

$handlerStack = HandlerStack::create();
$handlerStack->push(Middleware::history($historyContainer));

$client = new Client(['handler' => $handlerStack]);

// Will 301 redirect to either https://www.google.com 
// or https://www.google.[tld for your country]
$request = new Request('HEAD', 'https://google.com/');

$client->send($request);

foreach ($historyContainer as $httpTransaction) {
    $request = $httpTransaction['request'];
    var_dump($request->getMethod() . ' ' . (string)$request->getUri());
}

// current dumped output:
// string(24) "HEAD https://google.com/"
// string(71) "GET https://www.google.co.uk/"

// ideal dumped output:
// string(24) "HEAD https://google.com/"
// string(71) "HEAD https://www.google.co.uk/"
```
#### Possible Solutions

**Overview**
[`RedirectMiddleware::modifyRequest()`](https://github.com/guzzle/guzzle/blob/master/src/RedirectMiddleware.php#L161) changes the request method to GET in cases where the response status code is 301 and where the request has a body (assuming non-strict redirects as is the default). The current code checks this as `$statusCode <= 302 && $request->getBody()`

I assume the checking for a body relates to the need to change from a non-safe method (such as POST) to a safe method (such as GET) and in this respect the method change makes sense.

A request can easily be created without a body, however `GuzzleHttp\Psr7\Request::getBody()` will always return a `Psr\Http\Message\StreamInterface` instance and therefore `$request->getBody()` will always evaluate to `true`.

**Possible solution 1**
[...]

PR #2047, opened 2018-04-10
Retain safe request method when following redirects

webignition:
**TLDR**
A solution to #2046, ensures HEAD request method is retained when following 301 redirects.

**Overview**
`RetryMiddleware::modifyRequest()` modifies the method when following 301 redirects. In most cases, the method is changed to GET.

Safe methods (GET, HEAD, OPTIONS) can be retained safely when following redirects. 

This change ensures that requests issued when following 301 redirects will retain the original request method if that method is safe.

### DIFF
--- a/src/RedirectMiddleware.php
+++ b/src/RedirectMiddleware.php
@@ -169,7 +169,10 @@ class RedirectMiddleware
         if ($statusCode == 303 ||
             ($statusCode <= 302 && !$options['allow_redirects']['strict'])
         ) {
-            $modify['method'] = 'GET';
+            $safeMethods = ['GET', 'HEAD', 'OPTIONS'];
+            $requestMethod = $request->getMethod();
+
+            $modify['method'] = in_array($requestMethod, $safeMethods) ? $requestMethod : 'GET';
             $modify['body'] = '';
         }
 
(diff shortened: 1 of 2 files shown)


Chuvisco88:
Ran into the same issue (HEAD Request gets a 302 redirect and follows up with a GET Request).
This would fix it!
Why is it still open?! :(

webignition:
@Chuvisco88 I'm not a maintainer and so can't merge. Maybe follow this up with @sagikazarmark?

Nyholm:
This looks alright. 

Thank you @Chuvisco88 for the review. 

Thank you @webignition for the PR. Can you rebase this on master?

webignition:
Yep, can rebase, no problem. I'll do so in about a week when I'm back from vacation.

Nyholm:

webignition:
@Nyholm Apologies also for not getting round to doing so years ago!

Rebased now, and as a bonus all the checks are passing :smile:

albig:
@Nyholm, we would appreciate if this fix goes into a 6.5 bugfix release one day. 

We use TYPO3 which has as composer requirement even in current master "guzzlehttp/guzzle: ^6.3.0"...

PR #2050, opened 2018-04-12
Document use of a body with mock responses

aredridel:
This was a thing I had to gather from third parties while learning to test Guzzle clients. An example like this would have made it instantly obvious what the API is.

### DIFF
--- a/docs/testing.rst
+++ b/docs/testing.rst
@@ -32,7 +32,7 @@ a response or exception by shifting return values off of a queue.
 
     // Create a mock and queue two responses.
     $mock = new MockHandler([
-        new Response(200, ['X-Foo' => 'Bar']),
+        new Response(200, ['X-Foo' => 'Bar'], 'Hello, World'),
         new Response(202, ['Content-Length' => 0]),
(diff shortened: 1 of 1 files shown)


docs/testing.rst
```diff
@@ -41,8 +41,11 @@ a response or exception by shifting return values off of a queue.
     $client = new Client(['handler' => $handler]);
 
     // The first request is intercepted with the first response.
-    echo $client->request('GET', '/')->getStatusCode();
+    $response = $client->request('GET', '/');
+    echo $response->getStatusCode();
     //> 200
+    echo $response->getBody()->getContents();
```
sagikazarmark: Using `getContents` is generally not a good idea, so I would like to see it gone from the documentation entirely.

Instead casting the body to string should be the recommended way.

sagikazarmark:
Ping @aredridel :)

wirwolf:
Hello @aredridel, i create a fork of this repo and release this changes into the new version(6.4.0).

Namespace compatibility is 100%, and you can use my changes
* add `"guzzlehttp/guzzle": "6.3.*"` into replace block in your composer.json

PR #2052, opened 2018-04-17
Fix the wrong usage of _caseless_remove() when decode_content is specified

fiboknacky:
The _caseless_remove function was used incorrectly and it will clear all the values inside `$options['_conditional']` when `decode_content` is specified as an option.


sagikazarmark:
Thanks for submitting this fix.

PR #2058, opened 2018-04-20
ClientInterface/Client signatures do not match [request()/requestAsync()]

kdebisschop:
This mismatch creates problems for PhpUnit


Nyholm:
It is okey that an implementation may be accepting a wider range of inputs.

PR #2038, opened 2018-04-02
fixed php 7.2  Warning: count(): Parameter must be an array or an obj…

firma:
…ect that implements Countable

fixed php 7.2 ,Warning: count(): Parameter must be an array or an object that implements Countable

PR #2065, opened 2018-04-23
CurlFactory.php line 66 - handles array is null - throwing an error -…

gmize:
… testing for this fixed issue ( Laravel 5.6 / php 7.2.3-1 / Ubuntu / MailGun )

"guzzlehttp/guzzle": "~5.3|6.0",

Will this work for all or will is mask a deeper issue?


sagikazarmark:
Something really weird is going on here, that property should not be null IMO.

sagikazarmark:
BTW this is a duplicate of #2038

PR #2075, opened 2018-05-20
add ext-curl as dependency

PeterTheOne:


gmponos:
I don't think that it should be required to have cURL in order to install guzzle since it chooses the handler according to [this](https://github.com/guzzle/guzzle/blob/master/src/functions.php#L103) function and it can be a different handler from CurlHandler.

Furthermore you can create your own custom handler that will not use cURL.

GrahamCampbell:
Curl shouldn't be a dependency, as it's not required to use Guzzle. It's only requried to use the cURL adaper.

sagikazarmark:
I don't have anything to add.

PR #2084, opened 2018-06-10
Update Python2 URL to Python3

jdufresne:
Python 3 docs are more actively maintained and are the future of the Python project.


GrahamCampbell:
But guzzle was inspired by the Python 2 version, not 3.

sagikazarmark:
I agree, hopefully Python 2 documentation is not going anywhere. Thanks though.

Issue #1971, opened 2017-12-06
phpstan analysis error when using GuzzleException


babaorum:
| Q            | A
| ------------ | ---
| Bug?         | no
| New Feature? | no
| Version      | actual

#### Actual Behavior

Experiencing error with phpstan like those with level analysis >= 5

```
$exception of class My\Namespace\Exception\MyCustomException constructor expects Exception, GuzzleHttp\Exception\GuzzleException given.
```

#### Expected Behavior

No error found.

#### Steps to Reproduce

I am catching `GuzzleException` like this.
```php
try {
    // use guzzle for an http call
} catch (GuzzleException $exception) {
    throw new MyCustomException($exception);
}
```

with a custom exception like this : 
```php
class MyCustomException extends \Exception
{
    public function __construct(\Exception $exception)
    {
    }
}
```

#### Possible Solutions

The problem seems to be that `GuzzleException` does not extends \Exception directly. And there is no guarantee that every class that will implement `GuzzleException` will extend `\Exception`.

For me, the problem could be solved by changing GuzzleException into an abstract class which will extends `\RuntimeException` since it is the actual extended exception.

ondrejmirtes:
Hi, author of PHPStan here. You should:

1) Require \Throwable and not \Exception in your constructor.
2) Use the latest version of PHPStan (0.9) that understands that even interfaces caught in a catch block can be accepted as Throwable.

ondrejmirtes:
Maintainers of Guzzle can feel free to close this issue 😊

babaorum:
Thank you @ondrejmirtes ! I saw there was some stuff done for Guzzle, but did not saw this problematic.
Sorry everyone for the time lost. Love your projects :)

I am closing the issue.

rvanlaak:
@ondrejmirtes just as a follow-up on this. Since `0.10` was released (🎉) the phpstan error now thrown is:

> PHPDoc tag @throws with type GuzzleHttp\Exception\GuzzleException is not subtype of Throwable

Should the `GuzzleException` interface extend `\Throwable`? It does solve above error.

babaorum:
@rvanlaak Throwable interface can not be extended by userland interfaces. neither can it be implemented by userland classes.

ondrejmirtes:
@babaorum This definitely isn’t true, you can define interface X extends \Throwable.

babaorum:
[this](https://3v4l.org/rZMo8) won't work. 

[...]

PR #2095, opened 2018-06-26
Allow PHPStan to match with `@throws` phpdoc

rvanlaak:
[PHPStan `0.10` was released last Weekend](https://github.com/phpstan/phpstan/releases/tag/0.10) 🎉 , and will throw the following exception starting from level 2:

> PHPDoc tag @throws with type GuzzleHttp\Exception\GuzzleException is not subtype of Throwable

As all exceptions eventually extend `\RuntimeException`, and as we in userland want to rely on the interface, letting  `GuzzleException` interface extend `\Throwable` would be preferable.

also see #1971

### DIFF
--- a/src/Exception/GuzzleException.php
+++ b/src/Exception/GuzzleException.php
@@ -1,13 +1,4 @@
 <?php
 namespace GuzzleHttp\Exception;
 
-/**
- * @method string getMessage()
- * @method \Throwable|null getPrevious()
- * @method mixed getCode()
- * @method string getFile()
- * @method int getLine()
- * @method array getTrace()
- * @method string getTraceAsString()
- */
-interface GuzzleException {}
+interface GuzzleException extends \Throwable {}


rvanlaak:
@Nyholm found out that as well, are there plans for Guzzle 7 on `php: ^7.1` ? 

Any idea on how to cover Guzzle under PHPStan `^0.10`? I can imagine that others will start having this problem as well. 

> @throws \GuzzleHttp\Exception\GuzzleException|\RuntimeException

Using the above as phpdoc won't solve the phpstan error either, so the only thing I can think of at the moment is to let PHPStan exclude it 👎 :

```
# phpstan.neon
parameters:
    ignoreErrors:
        - '#PHPDoc tag @throws with type GuzzleHttp\\Exception\\GuzzleException is not subtype of Throwable#'
```

Majkl578:
Note that PHP 5.5 is **dead** for 2 years already. PHP 5.6 and PHP 7.0 are security-only and will die at the end of this year.
Supporting anything below 7.1 is basically non-sense from maintainer's PoV nowadays. You should migrate to fully typehinted PHP 7.1 code. :blush: (Let me know if I can help with this, did quite a lot of similar work on (not only) Doctrine projects.)

rvanlaak:
> The type provided with this tag MUST represent an object of the class Exception or any subclass thereof.

That actually is an interesting one, as all Guzzle's `GuzzleException` implementations actually do. The only thing is that the `GuzzleException` interface itself doesn't 😉

Majkl578:
Yes, that's correct, unfortunately using interfaces that don't extend Throwable was _never_ supported and was merely a _convention_ of some projects.
Anyone can implement such interface in non-Exception/non-Error class, making it basically pointless marker interface.

This was also brought up in the PHPStan PR: https://github.com/phpstan/phpstan/pull/1001#issuecomment-388532164
I personally believe that now, in 2018, it's almost an edge case (and you can simply add one `ignoreErrors` for legacy code) and PHP 5.x code can safely be ignored as it's not a primary target group.
OTOH it already found a bug elsewhere: https://github.com/doctrine/migrations/pull/739/files#diff-f2788aa62fbd5343d8695bd6cff10804L33

src/Exception/GuzzleException.php
```diff
@@ -1,13 +1,4 @@
...
- * @method string getMessage()
- * @method \Throwable|null getPrevious()
- * @method mixed getCode()
- * @method string getFile()
- * @method int getLine()
- * @method array getTrace()
- * @method string getTraceAsString()
- */
-interface GuzzleException {}
+interface GuzzleException extends \Throwable {}
```
babaorum: This is not valid in php cf my comment in the issue.

Majkl578:
Ad legacy PHP, see: https://github.com/guzzle/guzzle/pull/2098

janedbal:
The user base won't be left alone. They will still have the 6.x version, don't they? You can still maintain that branch if you want...

Majkl578:
> No, it's non-sense from a consumer point of view. From a maintainer point of view you have to consider the user base (you can't leave them alone), other elements of the roadmap, etc.

Not really true. As a maintainer, you have to maintain compatibility with a huge range of PHP versions. This itself makes things much more complex and greatly increases possibilities of bugs in _some_ PHP versions.

As said by @janedbal, releasing 7.0 with PHP 7.1 requirement does not mean _leaving anyone alone_, you still have 6.x with 5.5 support and bugfixes coming in.

OTOH people stuck on dead PHP versions block whole ecosystem in past.

rvanlaak:
If there would be a date when PSR 18 would get released, then that might be Guzzle 7's release goal? At least, based on https://github.com/guzzle/guzzle/pull/2098#issuecomment-400637989 I assume that would be a solid argument to define the future roadmap? As PSR 18 is still in draft, I don't see this happening within now and 2 years though. @Nyholm any plans on the Readiness vote?

This all doesn't mean work can't get started on a `develop` branch yet. It's most likely that Guzzle wil support PSR 18 asap after it's released, meaning that PHP7 will be the minimum. The remaining now would be to decide whether supporting abandoned versions of PHP is enough reason to formulate a roadmap / date when Guzzle `5.x` would become EOL.

Majkl578:
Sorry, I just don't see any correlation between PSR release and Guzzle 7. Those are two totally different milestones...
Also I absolutely don't understand why some unfinished PSR (with one single-method interface) should block all the work on PHP 7.1 migration?

mtdowling:
Does Drupal still rely on Guzzle? Are they still locked at PHP 5.5 or have they raised their minimum version requirement to 7+?

Majkl578:
How is that relevant?
Can Drupal use Guzzle 6? Yes,
Why should Drupal block evolution of whole ecosystem? Isn't that selfish?

wirwolf:
Psalm send me this error
```
ERROR: InvalidThrow - src/Infrastructure/Query/Http/ProgramQuery.php:67:16 - Class supplied for @throws GuzzleHttp\Exception\GuzzleException does not implement Throwable
     * @throws GuzzleException

```

GrahamCampbell:
Doesn't symfony have a polyfil that will allow this code on PHP 5?

rvanlaak:
> Doesn't symfony have a polyfil that will allow this code on PHP 5?

see https://github.com/symfony/polyfill/issues/101

GrahamCampbell:
Oh, yeh, they only have `Error` polyfilled... https://github.com/symfony/polyfill/blob/master/src/Php70/Resources/stubs/Error.php.

rvanlaak:
@Nyholm fixed this in #2375 so this PR can be closed 👍

PR #2098, opened 2018-06-26
Drop support for legacy PHP versions

Majkl578:
Hi,

it's end of Q2 2018 and current PHP version is 7.2, with 7.3 knocking on the door.

Guzzle still supports legacy PHP versions as far as 5.5.

5.5 was released 5 years ago and is dead for 2 years already.
5.6 was released almost 4 years ago and is in security-only mode for 18 months already, to be EOL'd in 5 months.
7.0 was released 2 and half years ago and is in security-only mode for 6 months already, to be EOL'd in 6 months.
7.1 was released 18 months ago and is still maintained, scheduled to enter security.only mode in 5 months.

Reference: https://secure.php.net/supported-versions.php

It's time to move on and leave relics behind.

_Targets v7.0._

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
         }
     ],
     "require": {
-        "php": ">=5.5",
+        "php": "^7.2",
         "guzzlehttp/promises": "^1.0",
         "guzzlehttp/psr7": "^1.4"
     },
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,10 +3,6 @@ language: php
 sudo: false
 
 php:
-  - 5.5
-  - 5.6
-  - 7.0
-  - 7.1
   - 7.2
   - 7.3
   - nightly
@@ -16,7 +12,7 @@ before_script:
   - composer install --no-interaction --prefer-source --dev
   - ~/.nvm/nvm.sh install v0.6.14
   - ~/.nvm/nvm.sh run v0.6.14
-  - if [[ "$TRAVIS_PHP_VERSION" != "7.0" || "$TRAVIS_PHP_VERSION" != "7.1" ]]; then echo "xdebug.overload_var_dump = 1" >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini || true; fi
+  - echo "xdebug.overload_var_dump = 1" >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini || true
 
 script: make test
 


Majkl578:
It is, see the diff. :)

simPod:
Me is blind 👍

Nyholm:
This is great. I do agree with dropping support for old PHP version. However, dropping support for PHP versions just because they are old is not reason enough. 

So, what non BC breaking feature does this PR allow? Even though I really like #2095, that one is a BC break and cannot be merged until version 7.0.0. So #2095 is not a reason to merge this PR.

Majkl578:
> So, what non BC breaking feature does this PR allow? 

Some major benefits:
* strict mode
* parameter/return types
* Throwable
* iterable
* advanced generators (`yield from`)
* private constants

Doctrine, Symfony and many other projects already require PHP 7.1.
Also even more consumers is writing typed code.

sagikazarmark:
Thanks for your feedback (all of you), it's a great thing to see feedback from the community.

I have to tell you bad news though: this won't land in Guzzle for a while. It would require a new major version, and the main blocker for that is the upcoming HTTP Client PSR.

Although I would really like to see Guzzle moving to PHP 7, we are gonna have to wait a little longer.

Majkl578:
What's wrong with shipping type-safe 7.0 and then, once the PSR is stable, shipping 8.0?

Majkl578:
Note that the PSR you mention is 3 years old and there seems to be barely any progress: https://github.com/php-http/httplug/issues/51 + https://github.com/php-fig/fig-standards/commits/master/proposed/http-client/http-client.md

rvanlaak:
If you look at the commits that make it to Guzzle's `master`, there are not that many features that are added in the last few months. In other words, for consumers Guzzle is a really stable and reliable library. This won't change by deciding on a new blocking release focused on language features.

`5.x` according to the [Version Guidance](https://github.com/guzzle/guzzle#version-guidance) still is maintained, but had it's latest patch release on July 15th 2016.

So, I think no customer will feel left out with deciding on a release date and develop branch for `7.x`, and making `5.x` EOL on that exact same date.

Probably even more important, [`http-client` will be `php:^7.0`](https://github.com/php-fig/http-client/blob/master/composer.json).

composer.json
```diff
@@ -38,7 +38,7 @@
     },
     "extra": {
         "branch-alias": {
-            "dev-master": "6.3-dev"
+            "dev-master": "7.0-dev"
```
carusogabriel: Shouldn't be `7.1`?

GrahamCampbell: No, this does not refer to the PHP version.

sagikazarmark:
> What's wrong with shipping type-safe 7.0 and then, once the PSR is stable, shipping 8.0?

Both are BC breaks. Regardless major versions can be used to communicate breaking changes, it's still not a great thing to do (too frequently). Guzzle did that in the past and it was admittedly a mistake.

> Note that the PSR you mention is 3 years old and there seems to be barely any progress

I would like to point out that Guzzle and that PSR (and HTTPlug which the PSR is based on) are maintained by the same people, so if it is blocked, probably it's for a reason.

I understand the reasons, I agree with most of them, but I believe in other priorities at this point.

I'm quite okay with planning and thinking about the 7.x migration as it will definitely happen, I just don't see why we have to do it ASAP, especially because the whole conversation is triggered by a PHPStan issue.

Majkl578:
In around 3 months PHP 5.6 and 7.0 will be dropped completely, PHP 7.1 will be EOL'd.

@Nyholm @mtdowling @jeremeamia @jeskew Can we maybe make a _deal_ here that the version bump will happen _around xmas_? 👍

Majkl578:
@sagikazarmark

> Both are BC breaks. Regardless major versions can be used to communicate breaking changes, it's still not a great thing to do (too frequently). Guzzle did that in the past and it was admittedly a mistake.

This isn't really a great argument. Lots of other projects do that too. Also wIth PHP 7.2+ only return types are a BC break, you can write code fully compatible with both pre-typed and post-typed code.
These are not behavioral breakages, those are much worse to handle ahd have greater impact.

> I would like to point out that Guzzle and that PSR (and HTTPlug which the PSR is based on) are maintained by the same people, so if it is blocked, probably it's for a reason.

I'd love to hear this reason, otherwise this is just an empty assumption. :)

> whole conversation is triggered by a PHPStan issue

That's only one of the consequences.

sagikazarmark:
I initiated a discussion with the maintainers to agree on a roadmap for Guzzle. This includes PSR-18 implementation, dropping PHP 5.x support and implementing type safety features. Once the roadmap is set, I'm going to publish it.

If I didn't make it clear before: I hear you, you are not being ignored, I even agree with most of the arguments you mentioned. But: things are not always as simple, so I'm asking for your patience so we can find the best migration path for everyone.

Majkl578:
PHP 7.1 reached EOL 2 days ago, with PHP 5.6 reaching EOL by the end of the year.

This PR has been rebased and now requires PHP 7.2 as it is the lowest supported PHP version.

Majkl578:
PHP 5.6 and thus all 5.x is dead as of today.

Majkl578:
_bump_

Nyholm:
No point in bumping this issue. We will drop legacy php verstions with the next major release.

84m:
> No point in bumping this issue. We will drop legacy php verstions with the next major release.
Any updates on the timetable here?

Nyholm:
This is fixed in master.

Majkl578:
> This is fixed in master.

Cool!
But not cool that you totally bypassed this PR waiting for merge for one and half year. OSS at its best.

PR #2099, opened 2018-06-27
Add support for Restful APIs with params in uri

lovelock:
Support Restful APIs with parameters in uri, like `/user/{user_id}`, with this middleware, you can call this API like this:

```php
$client = new Client(['base_uri' => 'https://example.com']);
$stack = HandlerStack::create();
$stack->push(Middleware::restful());
$client->request('GET', '/user/{user_id}', [
    'attributes' => [
        'user_id' => 100000,
    ],
]);

### DIFF
--- a/src/ClientInterface.php
+++ b/src/ClientInterface.php
@@ -12,7 +12,7 @@ use Psr\Http\Message\UriInterface;
  */
 interface ClientInterface
 {
-    const VERSION = '6.2.1';
+    const VERSION = '6.2.3';
 
     /**
      * Send an HTTP request.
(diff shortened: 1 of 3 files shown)


mtdowling:
I think this is cool and all, but I'd much rather this sort of thing be resolved *before* the URI is provided to Guzzle. We actually had a UriTemplate thing built into Guzzle in v3, but the complexity it added and coupling to the core library outweighed the utility it provided.

lovelock:
Thanks for your reply.

In fact, the reason why I wrote this middleware we'd better have one unified configuration of outer services to simplify the complexity of a project. Actually the middleware is extracted from a framework I created which has been used for years
and has a pretty good reputation.

Anyway developers can always write a similar middleware if they need as it is not complicated.

sagikazarmark:
Thanks for submitting this PR, it certainly looks useful, but I agree with @mtdowling about the complexity.

It's so trivial for a specific use case, that writing it should be a no brainer, but can't generalize it enough to cover every use case.

PR #2108, opened 2018-07-17
curl error msg for low version

wwulfric:
lower version curl wouldn't tell uri in error message.

### DIFF
--- a/src/Handler/CurlFactory.php
+++ b/src/Handler/CurlFactory.php
@@ -14,8 +14,9 @@ use Psr\Http\Message\RequestInterface;
  */
 class CurlFactory implements CurlFactoryInterface
 {
-    private const CURL_VERSION_STR = 'curl_version';
-    private const LOW_CURL_VERSION_NUMBER = '7.21.2';
+    const CURL_VERSION_STR = 'curl_version';
+    const LOW_CURL_VERSION_NUMBER = '7.21.2';
+
     /** @var array */
     private $handles = [];
 
--- a/src/Handler/CurlFactory.php
+++ b/src/Handler/CurlFactory.php
@@ -14,6 +14,8 @@ use Psr\Http\Message\RequestInterface;
  */
 class CurlFactory implements CurlFactoryInterface
 {
+    private const CURL_VERSION_STR = 'curl_version';
+    private const LOW_CURL_VERSION_NUMBER = '7.21.2';
     /** @var array */
     private $handles = [];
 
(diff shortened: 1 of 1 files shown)


sagikazarmark:
Newer versions does, so this solution sort of duplicates the uri in the error message? Maybe format the message according to the curl version?

wwulfric:
@sagikazarmark Solved

src/Handler/CurlFactory.php
```diff
@@ -14,6 +14,8 @@
  */
 class CurlFactory implements CurlFactoryInterface
 {
+    private const CURL_VERSION_STR = 'curl_version';
```
sagikazarmark: Constant visibility is not supported on PHP 5.x

sagikazarmark:
Thanks @wwulfric

wirwolf:
Hello @wwulfric, i create a fork of this repo and release this marge request into the new version(6.4.0).

src/Handler/CurlFactory.php
```diff
@@ -172,13 +176,22 @@ private static function createRejection(EasyHandle $easy, array $ctx)
...
             );
         }
-
-        $message = sprintf(
-            'cURL error %s: %s (%s)',
-            $ctx['errno'],
-            $ctx['error'],
-            'see https://curl.haxx.se/libcurl/c/libcurl-errors.html'
-        );
+        if (version_compare($ctx[self::CURL_VERSION_STR], self::LOW_CURL_VERSION_NUMBER)) {
```
guilliamxavier, line 179: Sorry to be late but this seems wrong, it's missing a comparison operator (either against 0 or as a third string argument, see [examples](https://www.php.net/manual/en/function.version-compare.php#refsect1-function.version-compare-examples)). Currently it's basically testing `curl_version()['version'] != '7.21.2'`. I guess you intended `>=` here. But actually I couldn't find any evidence that curl 7.21.2+ added the URI in the error message... Could you please explain your scenario?

guilliamxavier: ping @sagikazarmark: the code needs to be fixed!
Either you have evidence that curl 7.21.2+ added the URI in the error message, and then you should add `>= 0` (or `> 0`) between `))`;
or you don't, and then this should be reverted, and maybe re-submit only https://github.com/guzzle/guzzle/pull/2108/commits/bd368857d54081a9eeaea3c0ecfcb4e10455e56b

guilliamxavier: proposed #2648

PR #2110, opened 2018-07-23
Throw a BadResponseException when the Location header is empty on 3xx responses

coudenysj:
This avoids a TooManyRedirectsException, because Guzzle hits the same base uri over and over again

### DIFF
--- a/src/RedirectMiddleware.php
+++ b/src/RedirectMiddleware.php
@@ -209,9 +209,23 @@ class RedirectMiddleware
         ResponseInterface $response,
         array $protocols
     ): UriInterface {
+        $locationHeaderLine = $response->getHeaderLine('Location');
+
+        // Avoid TooManyRedirectsException when following empty location headers
+        if (empty($locationHeaderLine)) {
+            throw new BadResponseException(
+                sprintf(
+                    'The server responded with a %d status code, but supplied an empty Location header',
+                    $response->getStatusCode()
+                ),
+                $request,
+                $response
+            );
+        }
+
         $location = Psr7\UriResolver::resolve(
             $request->getUri(),
-            new Psr7\Uri($response->getHeaderLine('Location'))
+            new Psr7\Uri($locationHeaderLine)
         );
 
         // Ensure that the redirect URI is allowed based on the protocols.
--- a/src/Exception/BadResponseException.php
+++ b/src/Exception/BadResponseException.php
@@ -6,7 +6,7 @@ use Psr\Http\Message\RequestInterface;
 use Psr\Http\Message\ResponseInterface;
 
 /**
- * Exception when an HTTP error occurs (4xx or 5xx error)
+ * Exception when an HTTP error occurs (ie 4xx or 5xx error and 3xx redirects without Location header)
  */
 class BadResponseException extends RequestException
 {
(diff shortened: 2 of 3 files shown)


sagikazarmark:
Currently if there is no Location header the response is just returned to the user. I'm inclined to think that an empty header should trigger the same behavior to remain consistent.

WDYT?

coudenysj:
The RFC doesn't say anything about this, but I think the BadResponseException is more appropriate (more debuggable) than the TooManyRedirectsException.

coudenysj:
Do we make this configurable?

src/RedirectMiddleware.php
```diff
@@ -214,9 +214,20 @@ private function redirectUri(
         ResponseInterface $response,
         array $protocols
     ) {
+        $locationHeaderLine = $response->getHeaderLine('Location');
+
+        // Avoid TooManyRedirectsException when following empty location headers
+        if (empty($locationHeaderLine)) {
+            throw new BadResponseException(
+                'The server responded with a 302 status code, but supplied an empty Location header',
```
Tobion: This seems broken:
1. it does not need to be a 302, it could also be any other 3xx status code
2. An empty location header would be valid for a PUT request that returns a 303 with an empty location header, It means the resource was created and you can GET it from the same URI.
3. A 3xx response usually can also contain a body which a client might want to consume. Throwing an exception seems very unfriendly when the empty Location is just a minor error that could be ignored. If we throw an exception it needs to take the https://github.com/guzzle/guzzle/blob/master/src/RequestOptions.php#L133 option into account.

I think it makes sense to handle this case but needs some more work. Maybe just excluding 303 to solve 2) is enough.

coudenysj: Ok, thanks! Let me have a look tot fix this.

coudenysj: @Tobion I changed the exception message, but the exception should actually be removed and solved in another way.

The way you describe PUT requests with empty locations, means that Guzzle should auto fill the response Location header with the request URI.

This could still end up with a `TooManyRedirectsException` if the same uri keeps returning the same response.

GrahamCampbell:
What's the status of this PR?

Nyholm:
Can you rebase your PR?

coudenysj:
I'll do it tomorrow!

coudenysj:
@Nyholm rebased!

GrahamCampbell:
It would be good to get this fix merged into 6.x.

coudenysj:
How do I fix the `Roave / BC Check (pull_request)`?

Nyholm:
You dont need to fix that. 
Since master is now targeting 7.0 there will be BC breaks.

Nyholm:

gmponos:
Throwing a bad response here will become confusing...

https://github.com/guzzle/guzzle/blob/master/src/Exception/BadResponseException.php#L9

Nyholm:
Hm. But redirect to the same url does not make sense. Could you find a reference to that?

gmponos:
hmm.. https://github.com/guzzle/guzzle/pull/2110#discussion_r291442513

this is why I am 👎 on resolving discussions

Nyholm:
This is a pretty good StackOverlfow answer: https://stackoverflow.com/questions/16194988/for-which-3xx-http-codes-is-the-location-header-mandatory

That builds on to @Tobion's comments. 

So, I agree with @coudenysj, this makes sense and would be easier for everyone. However, it is technically incorrect.

gmponos:
Good resources.. for a future reference this is from the links above..

> 301, 302, 303, and 307 provide a Location only if the next URL is known. **Otherwise, the client/user has to decide what to do next**

PR #2122, opened 2018-08-13
Added support for PSR18

Nyholm:
I know this PR maybe premature for you and we may not be ready to drop PHP5 support. I'm soon about to open the official vote for PSR-18. And I wanted to show how it will effect Guzzle.

See the specification: https://github.com/php-fig/fig-standards/blob/master/proposed/http-client/http-client.md
Read the blog post: https://medium.com/php-fig/the-http-client-psr-9c2535132980

### DIFF
--- a/src/Exception/InvalidRequestException.php
+++ b/src/Exception/InvalidRequestException.php
@@ -1,0 +1,23 @@
+<?php
+
+declare(strict_types=1);
+
+namespace GuzzleHttp\Exception;
+
+use Psr\Http\Message\RequestInterface;
+
+class InvalidRequestException extends \InvalidArgumentException implements \Psr\Http\Client\Exception\RequestException,  GuzzleException
+{
+    private $request;
+    public function __construct(RequestInterface $request, $message)
+    {
+        $this->request = $request;
+        parent::__construct($message);
+    }
+
+
+    public function getRequest(): RequestInterface
+    {
+        return $this->request;
+    }
+}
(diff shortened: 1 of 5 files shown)


jeremeamia:
Seems pretty straightforward.

Nyholm:
The voting for the working group is open. @jeremeamia, add your vote here: https://groups.google.com/forum/#!topic/php-fig/BsroPEZcr_M

src/Exception/InvalidRequestException.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+declare(strict_types=1);
+
+namespace GuzzleHttp\Exception;
+
+use Psr\Http\Message\RequestInterface;
+
+class InvalidRequestException extends \InvalidArgumentException implements \Psr\Http\Client\Exception\RequestException,  GuzzleException
```
gmponos: Maybe you could have a look on this https://github.com/guzzle/guzzle/pull/2163. If we could merge this PR you can have this exception extend that one.

boesing:
How about a sattelite project for guzzle to support PSR-18 instead of directly integrating it here?
Something like `guzzlehttp/guzzle-psr18` with just a decorator inside.

Could work like the decorators for zend-cache which provide either PSR-6 or PSR-16 support.
https://github.com/zendframework/zend-cache/tree/master/src/Psr/SimpleCache
https://github.com/zendframework/zend-cache/blob/master/src/Psr/SimpleCache/SimpleCacheDecorator.php

Nyholm:
This is merged in master.

PR #2128, opened 2018-08-24
support set default handler

Yurunsoft:
We want to be able to manually set the default handler to let the guzzle-based project use the handler by default.

For example, we want to let the guzzle support the swoole coroutine by setting the default handler.

### DIFF
--- a/src/HandlerStack.php
+++ b/src/HandlerStack.php
@@ -18,6 +18,9 @@ class HandlerStack
     /** @var callable|null */
     private $cached;
 
+    /** @var string|callable|null */
+    private static $defaultHandler;
+
     /**
      * Creates a default handler stack that can be used by clients.
      *
@@ -270,4 +273,28 @@ class HandlerStack
 
         return 'callable(' . spl_object_hash($fn) . ')';
     }
+
+    /**
+     * Set a default handler
+     *
+     * @param string|callable|null $handler class name or callable. If value is null, that has no default handler
+     * @return void
+     */
+    public static function setDefaultHandler($handler)
+    {
+        static::$defaultHandler = $handler;
+    }
+
+    /**
+     * Get default handler
+     * 
+     * If return null, that has no default handler
+     *
+     * @return string|callable|null
+     */
+    public static function getDefaultHandler()
+    {
+        return static::$defaultHandler ?: null;
+    }
+
 }
--- a/src/functions.php
+++ b/src/functions.php
@@ -103,7 +103,12 @@ function debug_resource($value = null)
 function choose_handler()
 {
     $handler = null;
-    if (function_exists('curl_multi_exec') && function_exists('curl_exec')) {
+    $defaultHandler = get_default_handler();
+    if(is_string($defaultHandler)) {
+        $handler = new $defaultHandler;
+    } elseif (is_callable($defaultHandler)) {
+        $handler = $defaultHandler();
+    } elseif (function_exists('curl_multi_exec') && function_exists('curl_exec')) {
         $handler = Proxy::wrapSync(new CurlMultiHandler(), new CurlHandler());
     } elseif (function_exists('curl_exec')) {
         $handler = new CurlHandler();
@@ -121,6 +126,33 @@ function choose_handler()
     }
 
     return $handler;
+}
+
+/**
+ * Set a default handler
+ *
+ * @param string|callable|null $handler class name or callable. If value is null, that has no default handler
+ * @return void
+ */
+function set_default_handler($handler)
+{
+    $GLOBALS['guzzle_default_handler'] = $handler;
+}
+
+/**
+ * Get default handler
+ * 
+ * If return null, that has no default handler
+ *
+ * @return string|callable|null
+ */
+function get_default_handler()
+{
+    if(isset($GLOBALS['guzzle_default_handler'])) {
+        return $GLOBALS['guzzle_default_handler'];
+    } else {
+        return null;
+    }
 }
 
 /**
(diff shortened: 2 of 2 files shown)


Yurunsoft:
Hello, I have modified the code.

sagikazarmark:
Thanks for contributing to the project, but honestly I'm not really in favor of this change. I don't really see the need for it, rather see it as an unnecessary complication added to the already complicated enough configuration.

If you need a specific handler, why don't you instantiate Guzzle that way? Explicit configuration is almost always better. Or you could simply create a custom factory with your defaults and use it everywhere where you need to create an HTTP client.

Yurunsoft:
@sagikazarmark Because we want to adapt to other handlers without modifying some sdk code.

Nyholm:
Ah okey. You mean that third party libraries are creating a guzzle client and you have no way to modify their behavior. That is a good use case and I understand your situation. 

The problem is that those third party libraries depend on a concrete implementation of an HTTP client (guzzle) AND they do not allow you to access that client. 

I do not think this PR brings us closer to what is considered good architecture. I think you should be making PRs to your dependencies that put you in this situation in the first place.

Yurunsoft:
My PR does not affect the original running logic when the default handler is not set.

I think this is very useful, that the guzzle handler is not limited to built-in support.

When using a custom handler, you can change the original code as little as possible.

Nyholm:
> My PR does not affect the original running logic when the default handler is not set.

I know

> I think this is very useful, that the guzzle handler is not limited to built-in support.

It’s not limited, you just have to configure it the “good” way. 

> When using a custom handler, you can change the original code as little as possible.

You are talking about “the original code” as in “other third party libraries” (not guzzle). See my last comment about the other libraries and the use of guzzle.

sagikazarmark:
@Yurunsoft is this SDK open source? Actually, the proper way of solving the problem would be modifying that SDK to optionally accept a guzzle instance, not modifying guzzle.

Yurunsoft:
If you allow the global default handler to be set, there is absolutely no need to make any changes when using a third-party sdk, because guzzle follows the psr-7 specification.

And allows to set the global default handler, and will not affect the default handler acquisition logic when the handler is not set.

I think guzzle supports handlers, is that let developers develop their own handlers to handle requests.

Now more and more developers are using swoole, if you can pass this PR, swoole users will benefit endlessly.

Allows previous code to seamlessly support swoole coroutine by setting the global default handler.

I hope that you can carefully consider.

Thank you very much!

Nyholm:
It has been super carefully considered. I’ve been working 3 years with this kinda of questions and Mark got even more knowledge and experience then I do. 

The reason agaist this PR, as well as the solution to your problem is discussed above. You seam to ignore them for some reason...

Thank you for your PR and for the discussion. Please open a issue on Swoole, explain your problem and refer to this issue.

huangzhhui:
@Nyholm @sagikazarmark 
Swoole extension has an ability, that could make most of I/O operations becomes to an coroutine async non-blocking I/O operations automatically, but not inclue `curl` extension, there are a lots of framework has been adapte the swoole coroutine, like zend-expressive, swoft, imi, etc., and blocking I/O is a nightmare for a coroutine application, this is the premise.

Why the users need this PR ?
When the users wants to use some thirdparty sdks, and the sdk require guzzle, that means the sdk is a Blocking I/O driver, it will cause a drastic decline QPS (in coroutine application), the best pratice is allow users to set a global default handler, then the all thirdparty sdks would use this handler if the sdk has not set the handler, and now, all thirdparty sdks would becomes a coroutine async Non-Blocking driver, this PR make this become possible.

Nyholm:
I understand your scenario, I understand the code in the PR, I understand the implications of it. I also understand that your are frustrated (or starting to be frustrated) with me not accepting this PR.  

Do you understand my comments and responses?

Nyholm:
Great. 

If you have an issue with Swoole: 

> Please open a issue on Swoole, explain your problem and refer to this issue.


—————-

>  I have no ability and time to submit a lots of PRs to the thirdparty libraries,

I don’t understand this? But you will naturally assume that I got that time to review, comment and fix the issue for you? You also assume that I got the time to maintain this code for ever? And you assume that I should answer support calls on this code when people have weird side-effects because of this PR? If so, that is VERY selfish of you. I hope I misunderstood you. 

> the best way is Guzzle adapte this.

I understand that you see it that way. But it is not the best decision for Guzzle or the community as explained above.

sagikazarmark:
> the best way is Guzzle adapte this.

I'm sorry, but this is definitely not true. This is the easy way, not the proper solution of the problem. Messing with global options is one of the most common source of confusion.

Properly written SDKs should allow injecting a customized instance, so the problem here is badly written SDKs, not Guzzle. Making Guzzle solve this problem is simply not the solution.

I'm really sorry, but we are not going to accept this PR.

huangzhhui:
The project that I working on is part of swoole, so the issue with Swoole is not necessary, we communicate everyday, and that is why this PR exists.
And I want a solution, not a controversy, I am so sorry to let you misunderstand the comments, If you reject this PR, then maybe you should provide another solution, as an owner of Guzzle.
Maybe just provide a way to let `GuzzleHttp\Handler\StreamHandler` become the default handler, it will be great, Swoole can make SteamHandler become a coroutine async Non-Blocking driver.

sagikazarmark:
Well, an obvious solution would be to disable curl on your systems, that will make Guzzle choose the stream handler.

Again, I don't think Guzzle should be held responsible or forced to fix problems that is caused by poorly written software, so the solution shouldn't come from Guzzle either. But that's only one of the reasons why this PR is rejected, the other is the global scope messing.

PR #2132, opened 2018-08-28
Fix RetryMiddleware default exponential delay

dluces:
`RetryMiddleware::exponentialDelay` was previously returning an integer (good) representing SECONDS of delay. The `RetryMiddleware` was putting that value directly into the `delay` option for the retried request. However, the [docs say](http://docs.guzzlephp.org/en/latest/request-options.html#delay) that the `delay` option should be in milliseconds.

If the `RetryMiddleware` was used without a custom delay function, the `RetryMiddleware::exponentialDelay` is used, meaning that the delay was going to be minimal by default.

### DIFF
--- a/src/RetryMiddleware.php
+++ b/src/RetryMiddleware.php
@@ -47,7 +47,7 @@ class RetryMiddleware
      *
      * @param int $retries
      *
-     * @return int
+     * @return int milliseconds.
      */
     public static function exponentialDelay($retries)
     {
--- a/src/RetryMiddleware.php
+++ b/src/RetryMiddleware.php
@@ -51,7 +51,7 @@ class RetryMiddleware
      */
     public static function exponentialDelay($retries)
     {
-        return (int) pow(2, $retries - 1);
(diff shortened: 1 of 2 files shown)


Fuco1:
Can we get more attention from maintainers to this issue?  The middleware by default does "not" back off at all since it returns 2^1/2/3/4/5 which is <100 in milliseconds now.  This bombards the servers with tens of requests per second.

dluces:
I've just updated this PR's branch with latest master. I still think this should not be testing the actual delay mechanism as nothing related to that has been changed. Instead, it only changes how the delay time is calculated, @Nyholm

src/RetryMiddleware.php
```diff
@@ -51,7 +51,7 @@ public function __construct(
      */
     public static function exponentialDelay($retries)
```
Nyholm: Could you update the comment here saying what unit we are returning. 

We are returning `ms`

Nyholm:
Friendly update

PR #2139, opened 2018-09-07
prevent exception when handles is null

thenetimp:
I ran into a situation where $this->handles was null, but because null is not countable an exception was thrown.  This prevents that I added a check to see if handles is null.


Nyholm:
Thank you for this PR. 
It seams fair but I need you to write a short test to prove that your PR fixes the issue. So a test that fails without your `$this->handles != NULL ` check. 

Also, it is a great idea to use the PR template when creating the PR. It will help the maintainers.

Nyholm:
This was fixed in https://github.com/guzzle/guzzle/pull/1686

PR #2143, opened 2018-09-13
MockHandler - Add support to reset internal queue

ingluisjimenez:
| Q            | A
| ------------ | ---
| Bug?         | no
| New Feature? | yes
| Version      | 6.3.3

#### Actual Behavior

MockHandler doesn't offer a way of resetting the internal queue. There are testing scenarios where the handler has been initialized in the setUp() method of a test suite, to avoid duplication, and some specific responses are queued up as the default behavior but then you have tests which need to queue up something different (i.e a request error).

In this cases, it is not possible to reuse the handler defined in setUp(), and instad, it is required to re-instantiate and re-wire a handler and the dependencies all over again.

#### Expected Behavior

To avoid having to wire and set up the handler all over again, it would be great to expose a way to reset the internal queue in the MockHandler instance, and then just queue the new response(s).

This PR adds a method to MockHandler to reset the internal queue. Calling `$handler->reset(); ` would reset the internal array and allow the tester to queue up new responses.

### DIFF
--- a/docs/testing.rst
+++ b/docs/testing.rst
@@ -50,9 +50,18 @@ a response or exception by shifting return values off of a queue.
     echo $client->request('GET', '/')->getStatusCode();
     //> 202
 
+    // Reset the queue and queue up a new response
+    $mock->reset();
+    $mock->append(new Response(201));
+
+    // As the mock was reset, the new response is the 201 CREATED,
+    // instead of the previously queued RequestException
+    echo $client->request('GET', '/')->getStatusCode();
+    //> 201
+
+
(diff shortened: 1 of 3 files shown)


ingluisjimenez:
@sagikazarmark @Nyholm Please let me know what do you think about this little feature, and whether this is the correct way of doing this or if there is a better way that maybe I'm not aware of.

ingluisjimenez:
rebased with latest from master

ingluisjimenez:
Thank you for reviewing/merging! 🎉

Issue #2145, opened 2018-09-18
Feature suggestion: methods constants on RequestOptions


joubertredrat:
| Q            | A
| ------------ | ---
| Bug?         | no
| New Feature? | yes
| Version      | 

#### Actual Behavior

Today we use hardcoded methods on client for request.

#### Expected Behavior

We can consider to use constants instead actual behavior

#### Steps to Reproduce

Isn't a bug.

#### Possible Solutions

Constants for methos like examples below:
```
GuzzleHttp\RequestOptions::METHOD_GET
GuzzleHttp\RequestOptions::METHOD_POST
```
With this, will be more easy to maintain a default, like this.
```php
        $client
            ->request(
                \GuzzleHttp\RequestOptions::METHOD_GET,
                '/get',
                [
                    \GuzzleHttp\RequestOptions::DEBUG => true,
                ]
            )
        ;
```

PR #2147, opened 2018-09-19
Add method constants to request options

joubertredrat:
This solves feature suggestion in #2145


Nyholm:
There are no benefits added by merging this. See #2145.

PR #2150, opened 2018-09-24
Set response status code to 0 if status-code is missing in status line

ShakirzyanovArsen:
This avoids undefined offset if status code is missing ([issue](https://github.com/guzzle/guzzle/issues/1909))

### DIFF
--- a/src/Handler/EasyHandle.php
+++ b/src/Handler/EasyHandle.php
@@ -74,7 +74,7 @@ final class EasyHandle
 
         // Attach a response to the easy handle with the parsed headers.
         $this->response = new Response(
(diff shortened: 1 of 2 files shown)


FrontEndCoffee:
This would allow the code to fail silently. 

I would suggest throwing a parsing error if the statuscode is missing. (This was also suggested by @alexeyshockov [here](https://github.com/guzzle/guzzle/issues/1909#issuecomment-327425123))

sagikazarmark:
As mentioned in #1910, I would be happy to accept a PR implementing the suggested solution. Until then, closing this.

PR #2162, opened 2018-10-04
Remove unused exception class. Closes #2102

### LINKED ISSUE #2102: SeekException not used in Guzzle or any dependency
| Q            | A
| ------------ | ---
| Bug?         | no
| New Feature? | no
| Version      | 6.0.0

#### Actual Behavior
The class `GuzzleHttp\Exception\SeekException` is not used in any stream implemenations since 6.0.0.
It's also missing in the documentation.

#### Expected Behavior
Could it be removed.

#### Steps to Reproduce
```
composer require guzzlehttp/guzzle
fgrep -r SeekException .
```

#### Possible Solutions
Remove the class and tests of the class.

casconed:


casconed:
@gmponos not sure I see the need for a deprecation notice for a heretofore-unused exception.

gmponos:
Anyway it does not matter anymore.. I think 6.5 will be the last version of v6 which was tagged yesterday. So this can be easily merged for v7.

Nyholm:
Deprecation notices are the way to go, even if the class is never used. 

But master is not targeting 7.0 so no need for that (as @gmponos says). 

I've rebased your PR and waiting for the tests to pass. 
Thank you!

PR #2163, opened 2018-10-06
Add invalid argument exception

gmponos:
Hello there,

Let's say I have the following class:

``` php
class MyClientWrapper {
    private $httpClient;

    public function callSomething(){
          $response = $this->httpClient->get('/');
          return GuzzleHttp\json_decode($response->getBody());
    }
}
```

when I call the function above I have to do something like this:

``` php
try{
   $mywrapper->callSomething();
} catch (\InvalidArgumentException | GuzzleException $e){
    ....
}
```

This has the defect that I if another part of my code throws an `InvalidArgumentException` inside this class I can not separate it from the the GuzzleHttp `InvalidArgumentException`.

### DIFF
--- a/src/Exception/InvalidArgumentException.php
+++ b/src/Exception/InvalidArgumentException.php
@@ -2,4 +2,4 @@
 
 namespace GuzzleHttp\Exception;
 
-class InvalidArgumentException extends \InvalidArgumentException implements GuzzleException {}
+final class InvalidArgumentException extends \InvalidArgumentException implements GuzzleException {}
--- a/src/Exception/InvalidArgumentException.php
+++ b/src/Exception/InvalidArgumentException.php
@@ -1,0 +1,5 @@
+<?php
+
+namespace GuzzleHttp\Exception;
+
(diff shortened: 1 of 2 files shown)


src/Exception/InvalidArgumentException.php
```diff
@@ -0,0 +1,5 @@
+<?php
+
+namespace GuzzleHttp\Exception;
+
+class InvalidArgumentException extends \InvalidArgumentException implements GuzzleException {}
```
sagikazarmark: I think this class should be final for now, we can always open it up later.

Also, please add a trailing new line to the end of the file.

gmponos: Done.. I added the new line but I am not sure if it is visible through the PR changes tab

FrontEndCoffee:
Looks good 👍

sagikazarmark:
Looks good as always @gmponos ! Thanks!

wirwolf:
Hello @gmponos, i create a fork of this repo and release this marge request into the new version(6.4.0).

PR #2164, opened 2018-10-06
Update PHPDoc

FrontEndCoffee:
I went ahead and updated some of the php-doc-blocks. I also made sure that all the possible typehints are applied (keeping the php 5.5 version constraint in mind).

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -75,7 +75,13 @@ class Client implements ClientInterface
         $this->configureDefaults($config);
     }
 
-    public function __call($method, $args)
+    /**
+     * PHP Magic method
+     * @param  string $method
+     * @param  array $args
+     * @return Promise\PromiseInterface
+     */
+    public function __call($method, array $args)
     {
         if (count($args) < 1) {
             throw new \InvalidArgumentException('Magic request methods require a URI and optional options array');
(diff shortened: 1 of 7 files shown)


src/RetryMiddleware.php
```diff
@@ -90,6 +96,12 @@ private function onFulfilled(RequestInterface $req, array $options)
         };
     }
 
+    /**
+     * Execute rejected closure
+     * @param  RequestInterface $req
+     * @param  array            $options
+     * @return mixed
```
BackEndTea: Doesn't this return a closure?

src/functions.php
```diff
@@ -60,7 +60,7 @@ function describe_type($input)
  *                     format: "Name: Value"
  * @return array
  */
-function headers_from_lines($lines)
```
BackEndTea: Currently this can take any iterable like a `Generator` etc. Adding the array type will be a BC break

FrontEndCoffee: I fixed this

src/RetryMiddleware.php
FrontEndCoffee: Changed it to callable

gmponos:
Some of the functions have a docblock on the interface. So they inherit the one from the interface by default. What is the value of adding one on the functions as well?

FrontEndCoffee:
@gmponos in my opinion it's much nicer to work with code that has docblocks on every method, even though it might be a duplicate of its interface.

It has the added benefit that most of the IDE's can interpret those docblocks.

src/Client.php
```diff
@@ -89,6 +95,15 @@ public function __call($method, $args)
             : $this->request($method, $uri, $opts);
     }
 
+    /**
+     * Asynchronously send an HTTP request.
+     *
+     * @param RequestInterface $request Request to send
```
Nyholm: "Request to send" is not needed.

Nyholm:
This PR was merged in #2384

PR #2167, opened 2018-10-08
Apply PSR2

omarkdev:


localheinz:
Rather than manually applying the changes, what do you think about leveraging one of 

* [`friendsofphp/php-cs-fixer`](https://github.com/FriendsOfPHP/PHP-CS-Fixer)
* [`squizlabs/php_codesniffer`](https://github.com/squizlabs/PHP_CodeSniffer)

?

sagikazarmark:
It's on the todolist, but that would make lots of PRs conflicting.

gmponos:
> It's on the todolist, but that would make lots of PRs conflicting.

Then there will never be a good time to do so.

sagikazarmark:
Right, but the PHP 7 and PSR-18 upgrade will also be a huge upgrade, so I planned to time applying the PHP CS fixer to that date.

gmponos:
Having on CI (travis) the checks about codestyle wont this save you some review time from these huge upgrades? And as far as I know according to the public development that I see from PRs on github the progress for applying PSR-18 and PHP 7 is a not that big... maybe I am wrong.

Sorry if I push stuff... that's not my intention but I believe it would be better to have a set of rules and enforce them through CI, rather than having these huge upgrades and having more changes later on. Anyway I stop here.. your decision at the end of the day.

localheinz:
@sagikazarmark 

Then let's get these PRs merged! Please let us know if we can help move forward! 🤓

sagikazarmark:
Thanks @omarkdev @localheinz . This will be fixed when I get to #2239

PR #2172, opened 2018-10-21
Order Curl options set with request for curl_opts as last

### LINKED ISSUE #2171: Set CURLOPT_RETURNTRANSFER in curl options afte CURLOPT_FILE
I need to set `CURLOPT_RETURNTRANSFER` after setting `CURLOPT_FILE` with guzzle. When set not after but before, the `CURLOPT_RETURNTRANSFER` is neutralized by `CURLOPT_FILE` and vise versa. The behaviour is mentioned eg. here http://php.net/manual/en/function.curl-setopt.php#99082.

The problem is that curl options configuration array is built during several steps when creating the request and that `CURLOPT_RETURNTRANSFER` is among default options and is set first here https://github.com/guzzle/guzzle/blob/003757426cdd50fbad1c34cdbcc8a55f70a4fe8b/src/Handler/CurlFactory.php#L197

Then there's `CURLOPT_FILE` set https://github.com/guzzle/guzzle/blob/003757426cdd50fbad1c34cdbcc8a55f70a4fe8b/src/Handler/CurlFactory.php#L384 (or here https://github.com/guzzle/guzzle/blob/003757426cdd50fbad1c34cdbcc8a55f70a4fe8b/src/Handler/CurlFactory.php#L379 if sink is set but that neutralizes previously set `CURLOPT_RETURNTRANSFER` as well)

Finally, when passing curl config with request under `'curl'` key, they're applied here https://github.com/guzzle/guzzle/blob/003757426cdd50fbad1c34cdbcc8a55f70a4fe8b/src/Handler/CurlFactory.php#L49 with `array_replace`. That preserves options order (maybe desired behaviour?). 

Therefore my `CURLOPT_RETURNTRANSFER` settings passed with request is in options array before `CURLOPT_FILE` (because was set as default option) and is not applied to the request. How can I set `CURLOPT_RETURNTRANSFER` then? With current state of library I suppose it's not possible.

simPod:
This would fix #2171

### DIFF
--- a/src/Handler/CurlFactory.php
+++ b/src/Handler/CurlFactory.php
@@ -46,7 +46,11 @@ class CurlFactory implements CurlFactoryInterface
 
         // Add handler options from the request configuration options
         if (isset($options['curl'])) {
-            $conf = array_replace($conf, $options['curl']);
+            foreach ($options['curl'] as $key => $_) {
+                unset($conf[$key]);
+            }
+
+            $conf += $options['curl'];
         }
 
         $conf[CURLOPT_HEADERFUNCTION] = $this->createHeaderFn($easy);


simPod:
This is still malfunctioning, investigating

PR #2173, opened 2018-10-21
Pass handle to 'progress' callable

simPod:
The progress callable accepts 5 parameters where first is the handle as mentioned here http://php.net/manual/en/function.curl-setopt.php

image

Current state of the library passes only 4 arguments and removes the first argument containing the handle because it was not available in php versions <5.5. However this library requires [5.5+](https://github.com/guzzle/guzzle/blob/master/composer.json#L16) so it can be added https://github.com/guzzle/guzzle/blob/master/composer.json#L16

I guess it's BC break.

### DIFF
--- a/src/Handler/CurlFactory.php
+++ b/src/Handler/CurlFactory.php
@@ -460,14 +460,7 @@ class CurlFactory implements CurlFactoryInterface
                 );
             }
             $conf[CURLOPT_NOPROGRESS] = false;
-            $conf[CURLOPT_PROGRESSFUNCTION] = function () use ($progress) {
-                $args = func_get_args();
-                // PHP 5.5 pushed the handle onto the start of the args
-                if (is_resource($args[0])) {
-                    array_shift($args);
-                }
-                call_user_func_array($progress, $args);
-            };
+            $conf[CURLOPT_PROGRESSFUNCTION] = $progress;
         }
 
         if (!empty($options['debug'])) {
--- a/docs/request-options.rst
+++ b/docs/request-options.rst
@@ -750,6 +750,7 @@ The function accepts the following positional arguments:
         '/',
         [
             'progress' => function(
+                $handle,
                 $downloadTotal,
                 $downloadedBytes,
                 $uploadTotal,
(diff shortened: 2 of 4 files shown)


src/Handler/CurlFactory.php
```diff
@@ -460,14 +460,7 @@ private function applyHandlerOptions(EasyHandle $easy, array &$conf)
...
             }
             $conf[CURLOPT_NOPROGRESS] = false;
-            $conf[CURLOPT_PROGRESSFUNCTION] = function () use ($progress) {
-                $args = func_get_args();
-                // PHP 5.5 pushed the handle onto the start of the args
-                if (is_resource($args[0])) {
-                    array_shift($args);
-                }
-                call_user_func_array($progress, $args);
-            };
```
gmponos: If I am correct what guzzle library was doing here is to wrap one callable function inside another.

So this `call_user_func_array($progress, $args);` calls the developers callback function that he added from his options.

According to the docs the handler was not used and there fore that's why this part of the code exists:
```
                if (is_resource($args[0])) {
                    array_shift($args);
                }
```

If you ask why the handler does not exists. Most probably it's because guzzle provides an abstraction layer. Developers should not care about the handler that the guzzle internally uses.. So my guess is that they guzzle maintainers didn't want to have developers when setting this option having to think about the handler that the client internally uses.

simPod: Yes, but it's for compatibility purpose as it doesn't exists in old php versions. It's also documented there.

The handle is eg. accessible here https://github.com/guzzle/guzzle/blob/003757426cdd50fbad1c34cdbcc8a55f70a4fe8b/src/Handler/EasyHandle.php#L17 public and not internal

simPod:
Maybe we can reorder args and put handle as the last one. Then it won't even be BC

src/Handler/CurlFactory.php
gmponos: Can you leave this as unresolved? I believe that guzzle maintainers should share their toughs on this.

gmponos:
Closing.. as this PR makes the consumers of `Client` class to be aware of the `handler` and loses the purpose of a client abstraction.

If another maintainer feels differently we can re-discuss.

PR #2176, opened 2018-10-25
Update docblock

kevdotbadger:
Just updated the description of the `getRequest()` method.


FrontEndCoffee:
I think it's a little redundant

PR #2188, opened 2018-10-31
Remove assertEquals in favor of assertSame

gmponos:
Although in most of these changes using `assertSame` instead of `assertEquals` does not make much of a difference I believe that well known libraries like guzzle should promote using `assertSame`.  It feels like I am walking through a code that has `==` vs `===`.

### DIFF
--- a/tests/ClientTest.php
+++ b/tests/ClientTest.php
@@ -114,7 +114,7 @@ class ClientTest extends TestCase
             'handler'  => $mock,
             'base_uri' => 'http://bar.com'
         ]);
-        $this->assertEquals('http://bar.com', (string) $client->getConfig('base_uri'));
+        $this->assertSame('http://bar.com', (string) $client->getConfig('base_uri'));
         $request = new Request('GET', '/baz');
         $client->send($request);
         $this->assertSame(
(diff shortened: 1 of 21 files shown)


sagikazarmark:
Hm, I'm not sure I completely agree. Checking equality in case of *values* should be enough IMHO, since they are....values, not entities. What's the reasoning behind using assertSame? Is there a recommendation from PHPUnit?

gmponos:
This PR does not add much value to guzzle repository. Mostly I believe that big libraries should set a good programming example.

How would you feel if you went to a library, browsed through the code and you saw that it was full with `$var == 'my_string'`. Same way I feel also about `assertSame` vs `assertEqual`

In my programming life forcing using `assertSame` or `===` has created let's say "muscle" memory that I now try to use it everywhere even for trivial stuff and it saved me from a lot of possible bugs.

Anyway, I didn't thought that referencing an expertise would have been needed but here are some links:

https://github.com/sebastianbergmann/phpunit-documentation-english/issues/3
https://twitter.com/search?q=phpunit%20assertSame&src=typd

sagikazarmark:
Okay, I'm convinced!

Thanks!

PR #2192, opened 2018-11-12
Fix: Add test case for Url::combine()

localheinz:
This PR

* [x] adds a test case for `Url::combine()` to document the current behaviour

💁‍♂️ Might be interesting for people who have run into issues when configuring a client with a `base_url` that has a path component, but does not end with a trailing `/`, and then issuing requests using a relative path.


localheinz:
Thank you, @sagikazarmark!

PR #2202, opened 2018-11-15
Use dedicated assertEmpty assertion

carusogabriel:


sagikazarmark:
Do you mind opening another PR with key checking?

PR #2204, opened 2018-11-16
Use more accurate assertion while checking for non-existent array's key

carusogabriel:
As discussed in #2202

### DIFF
--- a/tests/Handler/StreamHandlerTest.php
+++ b/tests/Handler/StreamHandlerTest.php
@@ -286,7 +286,7 @@ class StreamHandlerTest extends TestCase
             'no'   => ['*']
         ]]);
         $opts = stream_context_get_options($res->getBody()->detach());
-        $this->assertTrue(empty($opts['http']['proxy']));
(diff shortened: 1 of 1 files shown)


tests/Handler/StreamHandlerTest.php
```diff
@@ -286,7 +286,7 @@ public function testAddsProxyButHonorsNoProxy()
             'no'   => ['*']
         ]]);
         $opts = stream_context_get_options($res->getBody()->detach());
-        $this->assertTrue(empty($opts['http']['proxy']));
+        $this->assertArrayNotHasKey($opts['http']['proxy']);
```
carusogabriel: Looks like this assertion does not handle undefined indexes as well. So, let's keep as it is? :)

sagikazarmark:
LGTM, thanks!

wirwolf:
Hello @carusogabriel, i create a fork of this repo and release this marge request into the new version(6.4.0).

PR #2212, opened 2018-11-20
Prevent Exception

giginos:
For PHP 7.x this may cause a "count(): Parameter must be an array or an object that implements Countable" Exception. See https://www.drupal.org/project/drupal/issues/2885610


sagikazarmark:
Duplicate of #2139 and  #2038. Can you please provide a test that fails without your fix?

giginos:
Hi @sagikazarmark ! It seems to be fixed on master since private $handles; is changed into private $handles = [];. In this PR ist now only a small unit-test for the solved issue - you can merge it if you like.

sagikazarmark:
Thanks @giginos !

PR #2219, opened 2018-11-25
RequestException should accept throwable as prev

gmponos:
Not sure if there should be also a wrapThrowable function or it is an overkill


gmponos:
Stupid of me... 🤦‍♂️ sorry for the noise.

PR #2220, opened 2018-11-25
Remove an always false check from RequestException

gmponos:
Is it possible the Response to be a PromiseInterface? 🤔


gmponos:
ping.. seems a simple fix here...

PR #2221, opened 2018-11-25
Create the important method "createRequest"

nadavvin:
GuzzleHttp\Psr7\Request allow to create Request instance but it not handle guzzle options

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -108,6 +108,15 @@ class Client implements ClientInterface
 
     public function requestAsync($method, $uri = '', array $options = [])
     {
+        $request = $this->createRequest($method, $uri, $options);
+        // Remove the option so that they are not doubly-applied.
+        unset($options['headers'], $options['body'], $options['version']);
+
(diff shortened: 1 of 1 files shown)


nadavvin:
Not really needed

nadavvin:
I reopen since it might still needed because for method like ->post() can get json array in body, but it is not happen in PSR Request.

So if this is possible solution ir you change Request to recognize 'json' key in body params

sagikazarmark:
I don't really think I want to expose request creation that way. It contains a lot of internal logic.

You can create your own request as easy as just encoding the body to json manually.

nadavvin:
It is not only json, what about query option. I need to escape the params myself while it not necessary to do in direct request like ->get()

Nyholm:
Im also 👎 

There is no need to have a method for creating `Request`s

nadavvin:
What about file upload? at least need helper methods to create the body...

PR #2217, opened 2018-11-24
[Minor] Fixed typehint $transferTime of constructor

gmponos:

PR #2223, opened 2018-11-25
Missing docblocks affecting code

gmponos:
Fixing some Docblocks and typehints that are "affecting" the code in a way.

I was inspired to have a look to the repo by [this PR](https://github.com/guzzle/guzzle/pull/2217)

I should have added these changes on that one but I didn't thought about it... 

I have found more "minor" stuff affecting the code but I am trying not to mix unrelated things together in the PRs.

### DIFF
--- a/src/Cookie/CookieJar.php
+++ b/src/Cookie/CookieJar.php
@@ -22,7 +22,7 @@ class CookieJar implements CookieJarInterface
      *                           arrays that can be used with the SetCookie
      *                           constructor
      */
-    public function __construct($strictMode = false, $cookieArray = [])
+    public function __construct($strictMode = false, array $cookieArray = [])
     {
         $this->strictMode = $strictMode;
 
@@ -94,8 +94,8 @@ class CookieJar implements CookieJarInterface
      */
     public function getCookieByName($name)
     {
-        // don't allow a null name
-        if ($name === null) {
+        // don't allow a non string name
+        if ($name === null || !is_scalar($name)) {
             return null;
         }
         foreach ($this->cookies as $cookie) {
@@ -103,6 +103,8 @@ class CookieJar implements CookieJarInterface
                 return $cookie;
             }
         }
+
+        return null;
     }
 
     public function toArray()
--- a/src/Cookie/CookieJar.php
+++ b/src/Cookie/CookieJar.php
@@ -22,7 +22,7 @@ class CookieJar implements CookieJarInterface
      *                           arrays that can be used with the SetCookie
      *                           constructor
      */
-    public function __construct($strictMode = false, array $cookieArray = [])
+    public function __construct($strictMode = false, $cookieArray = [])
     {
         $this->strictMode = $strictMode;
 
--- a/src/Handler/CurlFactory.php
+++ b/src/Handler/CurlFactory.php
@@ -161,7 +161,7 @@ class CurlFactory implements CurlFactoryInterface
 
         // If an exception was encountered during the onHeaders event, then
         // return a rejected promise that wraps that exception.
-        if ($easy->onHeadersException instanceof \Exception) {
+        if ($easy->onHeadersException) {
             return \GuzzleHttp\Promise\rejection_for(
                 new RequestException(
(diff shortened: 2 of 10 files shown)


src/Cookie/CookieJar.php
```diff
@@ -22,7 +22,7 @@ class CookieJar implements CookieJarInterface
      *                           arrays that can be used with the SetCookie
      *                           constructor
      */
-    public function __construct($strictMode = false, $cookieArray = [])
+    public function __construct($strictMode = false, array $cookieArray = [])
```
Nyholm: This is technically a BC break.. Or i mean. It has always been the intention but people will have problem if we merge this.

src/RedirectMiddleware.php
```diff
@@ -172,7 +172,7 @@ public function modifyRequest(
         // would do.
         $statusCode = $response->getStatusCode();
         if ($statusCode == 303 ||
-            ($statusCode <= 302 && $request->getBody() && !$options['allow_redirects']['strict'])
```
Nyholm: Why was this removed?

src/TransferStats.php
```diff
@@ -18,18 +18,18 @@
...
+     * @param mixed                  $handlerErrorData Handler error data.
+     * @param array                  $handlerStats     Handler specific stats.
      */
     public function __construct(
         RequestInterface $request,
         ResponseInterface $response = null,
         $transferTime = null,
         $handlerErrorData = null,
-        $handlerStats = []
+        array $handlerStats = []
```
Nyholm: This is fine because the class if final

src/Cookie/CookieJar.php
gmponos: What I had in mind:

1. Constructor is allowed to be overridden in PHP and it is not required to follow parent.
2. If some has passed a different value rather than array it would still have failed on foreach...

but now that I think of it.. in cases where someone was passing an `iterable` this will fail...  So yea I will revert. 

😢 👍

src/RedirectMiddleware.php
gmponos: Am I wrong? :/

src/Cookie/CookieJar.php
gmponos: Reverted... resolving.

src/TransferStats.php
gmponos: Reverted... resolving.

src/RedirectMiddleware.php
Nyholm: Nope. You are correct https://github.com/guzzle/psr7/blob/master/src/MessageTrait.php#L126

src/Handler/CurlFactory.php
```diff
@@ -161,7 +161,7 @@ private static function createRejection(EasyHandle $easy, array $ctx)
 
         // If an exception was encountered during the onHeaders event, then
         // return a rejected promise that wraps that exception.
-        if ($easy->onHeadersException) {
+        if ($easy->onHeadersException instanceof \Exception) {
```
gmponos: In a way this is also a BC...

If someone passed on `onHeadersException` a `StreamInterface` it would fail on the next line.. but now it wont.. and you will silently not see the error.

gmponos: Maybe I must revert this as well? WDYT?

Nyholm: I hesitated on this line. Let’s be safe rather then sorry. 

Revert it.

src/TransferStats.php
gmponos: @Nyholm ok.. l am unresolving this let's continue the discussion here

1. As I said even if it was not final it does not actually matter since I can do this:
```
public function __constructor(){
    parent::__constructor('myval');
}
```

One constructor does not have to follow the same signature as the parent constructor.

2. The strong argument here is not about final or not... The main point here is this:

```
$stats = new TransferStats($request, null, null, null, null, null)
$stats->getHandlerStats(); // someone is expecting null here... although the docblock says otherwise...
```

As you said.. better be safe than sorry.

PR #2224, opened 2018-11-25
Use fully qualified function calls

gmponos:
source: https://veewee.github.io/blog/optimizing-php-performance-by-fq-function-calls/

For a library such as guzzle performance is a must.


sagikazarmark:
Let's put this on hold until we sort out how to introduce automatic code fixes. I would rather trust PHPCSFixer to do this job.

Thanks @gmponos anyway!

PR #2229, opened 2018-12-04
Add composer.json to .gitattributes

reedy:


reedy:
It works fine :)

See for example https://github.com/wikimedia/css-sanitizer https://github.com/wikimedia/css-sanitizer/blob/master/.gitattributes#L8 and then it in a committed vendor repo https://github.com/wikimedia/mediawiki-vendor/tree/master/wikimedia/css-sanitizer and it in composer.lock https://github.com/wikimedia/mediawiki-vendor/blob/master/composer.lock#L2973-L2987

Dependancies can still be viewed in the composer.lock if they really need to

sagikazarmark:
I'm not 100% sure why this is necessary either. composer.lock is not version controlled in this project as it shouldn't be in a library.

Why is it bad to have composer.json in the generated archive?

Nyholm:
I’m :-1:

It could be helpful to have composer.json and it is not any benefit to remove it.

reedy:
Composer doesn't use the library's composer.json when installing stuff, it just uses the indexes at packagist

So at that point, this file is redundant in resultant vendor directories

sagikazarmark:
Still, I don't see any real benefit from removing this file, so closing. Thanks anyway.

legoktm:
The main benefit for removing the file is that...it's just extra. We (and I hope others do as well!) do thorough reviews of all upstream dependencies that we pull in, and the less files that need looking at/are updated in each new upstream release just makes it easier to maintain and review.

All of the information in composer.json is available in composer.lock and vendor/composer/installed.json as well for people who do find that information useful.

GrahamCampbell:
👎 on merging

sagikazarmark:
@legoktm we considered the proposal and decided not to merge it. Thank you!

PR #2242, opened 2019-01-03
Replace microtime() usages with hrtime()

### LINKED ISSUE #2214: PHP 7.3: use hrtime() instead of microtime()

**Description**  
Consider switching from microtime() to hrtime(), available in PHP 7.3. It provides a monotonic timestamp rather than a realtime timestamp.

**How to reproduce**  
Using microtime() is unsafe e.g. on a DST crossover or whenever the system time is adjusted, since the timestamps aren't relative to each other.

It is used in the CurlMultiHandler and the `on_stats` option.

Apart from needing to keep support for < 7.3, the change here should be without a BC break.

Dzhuneyt:
Fixes #2214

There are 5 test failures, but they are not caused by this PR. They are failing on master as well.

### DIFF
--- a/src/functions.php
+++ b/src/functions.php
@@ -340,5 +340,5 @@ function json_encode($value, $options = 0, $depth = 512)
  */
 function current_time()
 {
-    return function_exists('hrtime') ? hrtime(true) / 1e7 : microtime(true);
+    return function_exists('hrtime') ? hrtime(true) / 1e9 : microtime(true);
 }
--- a/src/functions.php
+++ b/src/functions.php
@@ -337,6 +337,7 @@ function json_encode($value, $options = 0, $depth = 512)
  * (depending on the PHP version, one of the two is used)
  *
  * @return float|mixed UNIX timestamp
+ * @internal
  */
 function current_time()
 {
--- a/src/functions.php
+++ b/src/functions.php
@@ -331,3 +331,14 @@ function json_encode($value, $options = 0, $depth = 512)
 
     return $json;
 }
+
+/**
+ * Wrapper for the hrtime() or microtime() functions
+ * (depending on the PHP version, one of the two is used)
+ *
+ * @return float|mixed UNIX timestamp
+ */
+function current_time()
+{
+    return function_exists('hrtime') ? hrtime(true) / 1e9 : microtime(true);
+}
--- a/src/functions.php
+++ b/src/functions.php
@@ -340,5 +340,5 @@ function json_encode($value, $options = 0, $depth = 512)
  */
 function current_time()
 {
-    return function_exists('hrtime') ? hrtime(true) / 1e9 : microtime(true);
+    return function_exists('hrtime') ? hrtime(true) / 1e7 : microtime(true);
 }
--- a/src/functions.php
+++ b/src/functions.php
@@ -339,7 +339,7 @@ function json_encode($value, $options = 0, $depth = 512)
  * @return float|mixed UNIX timestamp
  * @internal
  */
-function current_time()
+function _current_time()
 {
     return function_exists('hrtime') ? hrtime(true) / 1e9 : microtime(true);
 }
--- a/src/Handler/CurlMultiHandler.php
+++ b/src/Handler/CurlMultiHandler.php
@@ -88,7 +88,7 @@ class CurlMultiHandler
     {
         // Add any delayed handles if needed.
         if ($this->delays) {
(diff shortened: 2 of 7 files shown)


GrahamCampbell:
Would be better to declare a guzzle specific time function in https://github.com/guzzle/guzzle/blob/master/src/functions.php, instead of copy-pasting this code all over the place.

Dzhuneyt:
Introduced a new wrapper function `\GuzzleHttp\current_time()`.

src/functions.php
```diff
@@ -331,3 +331,14 @@ function json_encode($value, $options = 0, $depth = 512)
 
     return $json;
 }
+
+/**
+ * Wrapper for the hrtime() or microtime() functions
+ * (depending on the PHP version, one of the two is used)
+ *
+ * @return float|mixed UNIX timestamp
+ */
```
gmponos: I would suggest adding the internal annotation and also follow the convention of adding an underscore at the start of the function name... as it is done on other functions IIRC

Dzhuneyt: I don't see any of the other functions having the internal annotation nor having an underscore prefix. They are all under a namespace, so no name collisions should happen.

src/functions.php
```diff
@@ -331,3 +331,14 @@ function json_encode($value, $options = 0, $depth = 512)
...
+function current_time()
+{
+    return function_exists('hrtime') ? hrtime(true) / 1e9 : microtime(true);
```
Nyholm: Should be `1e7`, right?

https://3v4l.org/bZ1uo

src/functions.php
GrahamCampbell: > I would suggest adding the internal annotation and also follow the convention of adding an underscore at the start of the function name

No it's not?

GrahamCampbell: > so no name collisions should happen.

Yes, exactly.

gmponos: anyway not sure if this is a direction that must be followed... someone from guzzle could validate

GrahamCampbell: Only others of the functions in that file is like that, others are all not.

src/functions.php
Dzhuneyt: Right, my mistake. Fixed.

gmponos:
I am not sure why the discussion is resolved.. but unless if guzzle maintainers say otherwise I still insist on my proposal to mark the function `current_time` as internal and have it with underscore as `_current_time`... In the same way that it is done [here](https://github.com/guzzle/psr7/blob/master/src/functions.php#L762), [here](https://github.com/guzzle/psr7/blob/master/src/functions.php#L826) and [here](https://github.com/guzzle/psr7/blob/master/src/functions.php#L884)...

gmponos:
The reason suggesting the underscore is only for consistency with the rest of them.. in general I am not in favor of it.. but none the less I believe that it should be marked as internal in order not to have it under BC promise.

Llbe:
It must be `1e9` (nano) and not `1e7`, please see https://3v4l.org/OidWu.

Nyholm:
Thank you @Llbe. You are correct. One must never mix hrtime and microtime. 

1e9 is correct. My bad.

Nyholm:
Thank you. 

I see that you have not resolved @gmponos concern yet. If we mark the function as internal we can easily change it later if we want to. Say that we want to introduce `superSpecificTime()` that is available in PHP7.4. If we dont mark it as internal there is no BC way to introduce that. Also, it will be impossible to drop `microtime()`.

Dzhuneyt:
I see. Marked is as `@internal` now.

Dzhuneyt:
@Nyholm is there something else that stops these changes from being "mergeable"?

src/functions.php
```diff
@@ -331,3 +331,15 @@ function json_encode($value, $options = 0, $depth = 512)
...
 }
+
+/**
+ * Wrapper for the hrtime() or microtime() functions
+ * (depending on the PHP version, one of the two is used)
+ *
+ * @return float|mixed UNIX timestamp
+ * @internal
+ */
+function current_time()
```
Nyholm: One minor thing. All our other internal methods starts with an underscore. Please rename to `_current_time()`

GrahamCampbell:
Symfony has a polyfill for the `hrtime` function, actually. Maybe it would be better to use that.

wirwolf:
Hello @Dzhuneyt, i create a fork of this repo and release this marge request into the new version(6.4.0).

PR #2246, opened 2019-01-07
Update Client.php

dujuwen:
update \GuzzleHttp\Client class add json_encode_options for json_encode in line 324:

before:
$options['body'] = \GuzzleHttp\json_encode($options['json']);

after:
$options['body'] = \GuzzleHttp\json_encode($options['json'], $options['json_encode_options'] ?? 0);

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -321,7 +321,7 @@ class Client implements ClientInterface
 
         if (isset($options['json'])) {
             //add json_encode options
-            $options['body'] = \GuzzleHttp\json_encode($options['json'], $options['json_encode_options'] ?? 0);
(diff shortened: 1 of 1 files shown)


dujuwen:
What do I do when I need to pass in an option?

sagikazarmark:
Simple: do not pass in a json option, but a body option with your above code. I'm a bit hesitant about this code and there are tons of other priorities at the moment, so I'm asking for your patience.

Tobion:
👎 The json option is just a helper. If you need customization, then pass in the body directly with your json. Otherwise we will need to add alot of those secondary options.

PR #2247, opened 2019-01-07
Move NodeJS Test Server to src

danizord:
Currently, when installing Guzzle through Composer, the mocked NodeJS server does not get downloaded. This is because the [Git export-ignore](https://github.com/guzzle/guzzle/blob/6.3.3/.gitattributes#L10) is removing the /tests directory from exported zip.

I've moved the Server to the /src directory using the same namespace, so that it does not break BC.


tests/bootstrap.php
```diff
@@ -1,7 +1,6 @@
 <?php
 namespace GuzzleHttp\Test {
     require __DIR__ . '/../vendor/autoload.php';
-    require __DIR__ . '/Server.php';
```
danizord: This one was removed because now the class can be autoloaded by Composer.

danizord:
@sagikazarmark I'd like to use it to mock external APIs for my e2e tests, where my test runner runs in a separate container from the application, so I can't mock it at PHP level.

I guess an `@internal` comment on class docblock would be enough to tell people to not rely on it. What do you think?

danizord:
@sagikazarmark 

>I'm pretty sure there are much better API mocking tools out there.

Any recommendations? The Guzzle Server seems like the best option for me since the test runner is in PHP (Behat) but does not have access to the application runtime (separate container).

>What I think we should probably do is update the relevant parts of the documentation to make it clear that these tools are not generally available for use.

So I think it would be better to simply drop the http://docs.guzzlephp.org/en/latest/testing.html#test-web-server part from documentation :)

sagikazarmark:
You are probably right, I will update the documentation at some point.

Any test http server should work fine, you only need to make sure it's started prior to running tests in a separate process.

Issue #2248, opened 2019-01-08
429 - to many requests


Gemorroj:
**Guzzle version(s) affected**: 6.3.3

**Description**  
It creates a lot of connections per host.

**How to reproduce**  
https://gist.github.com/Gemorroj/97df5b1d3a50534dfa6fc21ae7f41813

debug mode says (As far as I understand, a new connection is created for each request):
```
* Found bundle for host api.tempo.io: 0x1340edbe300 [serially]
* Found bundle for host api.tempo.io: 0x1340edbe300 [serially]
* Found bundle for host api.tempo.io: 0x1340edbe300 [serially]
* Found bundle for host api.tempo.io: 0x1340edbe300 [serially]
* Found bundle for host api.tempo.io: 0x1340edbe300 [serially]
* Found bundle for host api.tempo.io: 0x1340edbe300 [serially]
* Found bundle for host api.tempo.io: 0x1340edbe300 [serially]
* Found bundle for host api.tempo.io: 0x1340edbe300 [serially]
* Found bundle for host api.tempo.io: 0x1340edbe300 [serially]
* Found bundle for host api.tempo.io: 0x1340edbe300 [serially]
* Found bundle for host api.tempo.io: 0x1340edbe300 [serially]
* Found bundle for host api.tempo.io: 0x1340edbe300 [serially]
* Found bundle for host api.tempo.io: 0x1340edbe300 [serially]
* Found bundle for host api.tempo.io: 0x1340edbe300 [serially]
* Found bundle for host api.tempo.io: 0x1340edbe300 [serially]
* Found bundle for host api.tempo.io: 0x1340edbe300 [serially]
* Found bundle for host api.tempo.io: 0x1340edbe300 [serially]
* Found bundle for host api.tempo.io: 0x1340edbe300 [serially]
* Found bundle for host api.tempo.io: 0x1340edbe300 [serially]
* Found bundle for host api.tempo.io: 0x1340edbe300 [serially]
* Found bundle for host api.tempo.io: 0x1340edbe300 [serially]
* Found bundle for host api.tempo.io: 0x1340edbe300 [serially]
* Found bundle for host api.tempo.io: 0x1340edbe300 [serially]
* Found bundle for host api.tempo.io: 0x1340edbe300 [serially]
*   Trying 52.72.73.251...
* TCP_NODELAY set
* Hostname 'api.tempo.io' was found in DNS cache
*   Trying 52.72.73.251...
* TCP_NODELAY set
[...]

PR #2249, opened 2019-01-08
limit connections to host

Gemorroj:
Not a very nice fix for #2248.
But at least something.

### DIFF
--- a/src/Handler/CurlMultiHandler.php
+++ b/src/Handler/CurlMultiHandler.php
@@ -52,7 +52,7 @@ class CurlMultiHandler
         if ($name === '_mh') {
             $this->_mh = curl_multi_init();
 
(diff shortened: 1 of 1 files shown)


sagikazarmark:
Wouldn't using `'curl'` [request option](https://github.com/guzzle/guzzle/blob/master/src/Handler/CurlFactory.php#L51) be a solution?

Gemorroj:
@sagikazarmark this options for `curl_init`, not for `curl_multi_init` https://github.com/guzzle/guzzle/blob/master/src/Handler/CurlMultiHandler.php#L50

alexeyshockov:
IMO we need a more elegant solution for that. A new option `curl_multi` (like `curl`, but for `curl_multi_init()`/`curl_multi_setopt()`) is more flexible, since there are many `CURLMOPT_*` configuration constants available.

alexeyshockov:
@sagikazarmark, @Gemorroj, put some code in #2287. Take a look.

Tobion:
#2287 looks better to me indeed

PR #2253, opened 2019-01-24
Update Client.php

jeskiZhao:
I think it's better not to overwrite  the query string values supplied in the URI of a request  when the query option is an empty array.  For the request like this,
$client->request('GET', '/get?abc=123', ['query' => []]); 
After making test I think it should throw an exception.

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -365,13 +365,13 @@ class Client implements ClientInterface
             }
         }
 
-        if (!empty($options['query'])) {
+        if (isset($options['query'])) {
             $value = $options['query'];
-            if (is_array($value)) {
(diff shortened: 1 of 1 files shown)


phansys:
IMO, inferring that an empty array must behave differently than a array with values is counter intuitive. Even, the same case could occur when `$options['query']` is a string (regardless it is empty or not).
For your use case, I think you should simply use `null` with this option, since with this value the replace must be not performed:
https://github.com/guzzle/guzzle/blob/90696662779a42ed9b6c973c8afa4a23f88eee98/src/Client.php#L368

Tobion:
👎 I agree with @phansys

PR #2255, opened 2019-01-28
Upgrade guzzlehttp/ringphp composer reference to 1.1.1

### LINKED ISSUE #2254: PHP 7.3 warning from dependency for ringphp library on Guzzle 5.3.3
**Guzzle version(s) affected**: 5.3.3

**Description**  
We should upgrade the Composer reference to the `guzzlehttp/ringphp` library to `1.1.1` where the PHP 7.3 switch continue warning is fixed.

rebeccahum:
Fixes #2254.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,7 @@
     ],
     "require": {
         "php": ">=5.4.0",
-        "guzzlehttp/ringphp": "^1.1",
+        "guzzlehttp/ringphp": "^1.1.1",
         "react/promise": "^2.2"
     },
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Why is this necessary? Composer will already resolve the newest version, and 1.1.0 does work OK, just not on 7.3 as you outlined.

Also, this change actually doesn't stop composer resolving 1.1.0 of ringphp, because it can just resolve an older version of guzzle in order to resolve 1.1.0 if someone forced it.

PR #2259, opened 2019-02-03
Update Client.php to handle null headers

okdewit:
Any request with options `['json' => $data, 'headers' => null]` gives an exception on line 326 of Client.php:

```
Undefined index: _conditional in the Psr7\_caseless_remove() call on line 326 
```

This seems like a bug to me, because the code itself says "Allows default headers to be unset", with a specific `if ($options['headers'] === null)` check:

https://github.com/guzzle/guzzle/blob/master/src/Client.php#L226-L229

When that if statement is true, it will currently always throw an exception in one of the `applyOptions` branches, because `$defaults['_conditional']` must be an array.

When the headers key exists but is null, it seems better to set `$defaults['_conditional']` to an empty array instead, this way `headers => null` is handled gracefully.

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -225,7 +225,7 @@ class Client implements ClientInterface
         if (array_key_exists('headers', $options)) {
             // Allows default headers to be unset.
             if ($options['headers'] === null) {
-                $defaults['_conditional'] = null;
+                $defaults['_conditional'] = [];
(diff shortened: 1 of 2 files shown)


okdewit:
@Nyholm Any update on this? If anything is wrong with the change or test, I'd gladly fix it 😃

Issue #2185, opened 2018-10-31
Exception Message truncation isn't so friendly


deleugpn:
Hi everyone,

I know this has been extensively discussed here https://github.com/guzzle/guzzle/issues/1722 but unfortunately the exception truncation is a constant pain-point for me to the point that I'm almost maintaining a fork of Guzzle just to change that number. I understand a few of the points made in the thread such as the uncertainty of the size of the content in the response. A counter-argument that I would like to raise is the fact that we are forced into treating Guzzle Exceptions differently in our systems because Guzzle won't let me see enough of the response content. And then I have to turn this

```
$client->get($resource);
```

into this

```
try {
    $client->get($resource);
} catch(BadResponseException $e) {
    $e->getResponse()->getBody()->getContents();
}
```

throughout my entire platform. Sometimes we include packages that make usage of Guzzle and the package won't let me see the full error message as well.

This seem such a small issue that brings a lot of headache. an extremely simple static `Client::MESSAGE_SIZE = 6400;` would suffice for me and probably for most of the people suffering from this. The pain-point for me is located on `5xx` requests and `4xx` requests between internal microservices. Those are not production-ready code that I'm working on and trying to stabilize them in a way that I'll never have a `5xx` or `4xx`, but I don't know what to fix because I can't see the error message.

Something else described on the linked thread was this: 

> Obviously you don't want to log that 10k JSON, so in this case you should extract the necessary/useful information from the body, and properly log it instead of the exception message.

Why **Obviously** I don't want to log that 10k JSON? What if I do? Even if I don't need the 10k, 120 characters is still not enough for me to see which table name does not exist from a standard MySQL error.

Anyway, I would really hope that this could be revisited. I would like to finish by saying that what Guzzle is doing right now is forcing me to catch an exception not because I should catch it, but because I cannot see the exception's message if I don't. I would like to be able to only catch exceptions that I really need to catch and let all other exceptions be reported.

Thanks for the this amazing tool.

gmponos:
1. Although I am not a member of this repository I can't help to notice that I believe that a new issue was not required since you found the old one. You could just comment there.
2. Personally I dislike the truncate that happens there but from a different point of view. I believe that a title on the exception should have been enough. 

https://github.com/guzzle/guzzle/blob/master/src/Exception/RequestException.php#L105

```
        $message = sprintf(
            '%s: `%s %s` resulted in a `%s %s` response',
            $label,
            $request->getMethod(),
            $uri,
[...]

PR #2272, opened 2019-03-03
Allow developer to configure size of response body on exception

deleugpn:
Following up on https://twitter.com/mtdowling/status/1102266416005144577 I'm proposing a solution to https://github.com/guzzle/guzzle/issues/2185. 
Currently, Guzzle will truncate the response body in case of exception to protect the developer from spamming their own logging system. However, 120 characters is incredibly small and doesn't cover enough characters to assess what caused the operation to fail.
Furthermore, it is a pretty valid use-case where the developer is in control of both the client and the server. Knowing the full extend of the server application gives the developer privileged knowledge on what is the best amount of characters to log. For my use-case, personally, I don't need a limit as I know all my endpoints will return a valid JSON response that can easily fit in my logging system.

This pull request allows Guzzle to provide an opt-in alternative that complies with the Exception `getMessage()` method. A logging system usually have a code that resembles the following snippet:

```
public function report(Throwable | Exception $e): void {
    $this->logger->error($e->getMessage());
}
```

With the current state of Guzzle, that code has to cater for a special case just for Guzzle:

```
public function report(Throwable | Exception $e): void {
    if ($e instanceof GuzzleHttp\Exception\RequestException) {
        $this->logger->error(json_encode($e->getResponse->getBody()->getContents()));
    }

    $this->logger->error($e->getMessage());
}
```

If this proposal gets merged, we can go back to relying on polymorphism on the Exception Handler if we're willing to configure Guzzle with `RequestException::$responseSize = null` and the 1st snippet would be enough.

### DIFF
--- a/src/Exception/RequestException.php
+++ b/src/Exception/RequestException.php
@@ -11,6 +11,9 @@ use Psr\Http\Message\UriInterface;
  */
 class RequestException extends TransferException
 {
+    /** @var int */
+    private static $responseSize = 120;
+
     /** @var RequestInterface */
     private $request;
 
@@ -136,11 +139,14 @@ class RequestException extends TransferException
             return null;
         }
 
-        $summary = $body->read(120);
-        $body->rewind();
+        if (is_null(self::$responseSize)) {
+            $summary = $body->getContents();
+        } else {
+            $summary = $body->read(self::$responseSize);
 
-        if ($size > 120) {
-            $summary .= ' (truncated...)';
+            if ($size > self::$responseSize) {
+                $summary .= ' (truncated...)';
+            }
         }
 
         // Matches any printable character, including unicode characters:
@@ -150,6 +156,36 @@ class RequestException extends TransferException
         }
 
         return $summary;
+    }
+
+    /**
+     * Define the amount of bytes to use when summarizing the response exception.
+     *
+     * @param int|null $size
+     *
+     * @return void
+     */
+    public static function setResponseSize($size)
+    {
+        if (is_null($size) || is_int($size)) {
+            self::$responseSize = $size;
+
+            return;
+        }
+
+        $message = '$size must be an integer with the amount of bytes before truncating or null to avoid truncation altogether';
+
+        throw new InvalidArgumentException($message);
+    }
+
+    /**
+     * Restore the amount of bytes to use when summarizing the response exception to it's default value.
+     *
+     * @return void
+     */
+    public static function restoreDefaultResponseSize()
+    {
+        self::$responseSize = 120;
     }
 
     /**
--- a/src/Exception/RequestException.php
+++ b/src/Exception/RequestException.php
@@ -179,16 +179,6 @@ class RequestException extends TransferException
     }
 
     /**
-     * Restore the amount of bytes to use when summarizing the response exception to it's default value.
-     *
-     * @return void
-     */
-    public static function restoreDefaultResponseSize()
-    {
-        self::$responseSize = 120;
-    }
-
-    /**
      * Obfuscates URI if there is an username and a password present
      *
      * @param UriInterface $uri
--- a/tests/Exception/RequestExceptionTest.php
+++ b/tests/Exception/RequestExceptionTest.php
@@ -118,7 +118,7 @@ class RequestExceptionTest extends TestCase
         $response = new Response(500, [], $content);
         $e = RequestException::create(new Request('GET', '/'), $response);
         $this->assertContains($content, $e->getMessage());
-        RequestException::restoreDefaultResponseSize();
+        RequestException::setResponseSize(120);
     }
 
     public function testExceptionMessageIgnoresEmptyBody()
--- a/tests/Exception/RequestExceptionTest.php
+++ b/tests/Exception/RequestExceptionTest.php
@@ -111,6 +111,16 @@ class RequestExceptionTest extends TestCase
         $this->assertContains($expected, $e->getMessage());
     }
 
+    public function testCreatesExceptionWithoutTruncatingResponseBody()
+    {
+        RequestException::setResponseSize(null);
+        $content = '{"message":"The given data was invalid.","errors":{"email":["The email field is required."],"password":["The password field is required."],"first_name":["The first name field is required."],"last_name":["The last name field is required."],"address":["The address field is required."],"zipcode":["The zipcode field is required."]}}';
+        $response = new Response(500, [], $content);
+        $e = RequestException::create(new Request('GET', '/'), $response);
+        $this->assertContains($content, $e->getMessage());
+        RequestException::restoreDefaultResponseSize();
+    }
+
     public function testExceptionMessageIgnoresEmptyBody()
     {
         $e = RequestException::create(new Request('GET', '/'), new Response(500));


src/Exception/RequestException.php
```diff
@@ -11,6 +11,9 @@
  */
 class RequestException extends TransferException
 {
+    /** @var int */
+    public static $responseSize = 120;
```
Tobion: missing phpdoc explaining what this does

specialtactics:
Hi, what is the status of this?

@deleugpn can you please add the phpdoc, if that's all that's required?

I agree, this is a really big hassle, and it needs to be improved.

deleugpn:
Sorry, I dont know how I missed the messages here. I will take a look at this on Monday.

src/Exception/RequestException.php
Nyholm: Maybe use a setter to enforce this being an integer.

src/Exception/RequestException.php
```diff
@@ -136,13 +139,18 @@ public static function getResponseBodySummary(ResponseInterface $response)
             return null;
         }
 
-        $summary = $body->read(120);
-        $body->rewind();
+        if (self::$responseSize) {
```
Nyholm: This check is not needed if it is an integer. 

(Positive numbers will truncate, negative number will allow everything)?

src/Exception/RequestException.php
```diff
@@ -136,13 +139,18 @@ public static function getResponseBodySummary(ResponseInterface $response)
...
+        } else {
+            $summary = $body->getContents();
+        }
 
-        if ($size > 120) {
+        if (self::$responseSize && $size > self::$responseSize) {
             $summary .= ' (truncated...)';
         }
 
+        $body->rewind();
```
Nyholm: Why should we rewind it? Maybe it was not rewinded form the start..

deleugpn: It's been some time, I don't fully remember why I did the `rewind`. Removed.

src/Exception/RequestException.php
deleugpn: I changed the check for `null` only so that we get the full contents if the developer chooses to `setResponseSize(null);`

deleugpn:
- Property made private;
- Added setter without type-hint or return type due to PHP 5.5 constraints;
- Added code check for typed property;
- Added exception for invalid;

When Guzzle is ready for 7.1 requirement, the `setResponseSize` method can be simplified.

src/Exception/RequestException.php
```diff
@@ -152,6 +158,36 @@ public static function getResponseBodySummary(ResponseInterface $response)
...
+
+        throw new InvalidArgumentException($message);
+    }
+
+    /**
+     * Restore the amount of bytes to use when summarizing the response exception to it's default value.
+     *
+     * @return void
+     */
+    public static function restoreDefaultResponseSize()
```
Nyholm: Is this needed?

deleugpn: I think this could add a great DX, but I'm happy to take it out if you prefer.

Nyholm: I dont see the scenario and it can easily be replaced with `$foo->setResponseSize(120)`.

specialtactics:
Fantastic news! Thanks @deleugpn & @Nyholm

sagikazarmark:
Personally, I'm not a huge fan of this change, for the same reason this limit was introduced in the first place.

Consider the following example: A library decides to set this global limit to `PHP_INT_MAX`. Regardless if an application sets it to `120` or not, a library might still set it on the fly.

Then **every** HTTP request you send through Guzzle will result in potentially large exception messages which the application will probably log or send to another service.

This limit was introduced to make sure that Guzzle don't flood your system with logs. This is a direct and unsafe workaround for that and one can't effectively prevent it in an application.

deleugpn:
As I already argued in the ticket and on Twitter: that's not Guzzle's problem and I prefer that problem over not knowing what is happening. 
It's like the argument of banning the sales of knives because somebody might cut themselves. 
If a library decides to do that then it's not a library I want to use as it's not that library's job to take away my ability to configure this, which makes another WTF argument: "As a library,  Guzzle will not let you configure this because another library can take that control away from you. So let's be that library ourselves and not give control in the first place".

sagikazarmark:
@deleugpn You **have** control. You can catch an exception and get the body from it. You can even automate it by writing a filter in your error handling solution, to attach the body to the exception message.

You can even write your own `http_error` middleware that throws an exception with the body in the message.

So you have full control, but you disagree with the default behavior. And the thing is, the default behavior **MUST** be as defensive as possible. A globally changeable variable goes directly against that.

And I haven't even started talking about how bad the `http_error` middleware is in the first place. An HTTP error response is actually not exceptional from the HTTP client point of view. It's a perfectly valid answer that tells the consumer of the HTTP client that something went wrong. You can then throw an exception at that point that's specific to your domain.

Given this is only a problem in the http_error middleware, I would prefer this to be an option in that middleware, probably with a `RequestException::createWithResponseSize` named constructor. It would require you to instantiate the middleware for yourself, so there is now way other than the *owner* of the client can change that size limit.

deleugpn:
@sagikazarmark We've both been discussing this for months already and always back-and-forth on the same arguments. As I stated on the Ticket and on Twitter, the behavior you describe breaks the basic concept of `Exception::getMessage()`. You want me to treat Guzzle as a special snowflake by not adhering to the purpose of `Exception::getMessage()`, which should give me the message for the exception. So, yes, I can write a bunch of boilerplate everytime I'm using Guzzle because of it, but that's what I'm trying to address. My exception handler is based off of a well established method that is `Exception::getMessage()`. If I, or any other developer, wants to cut off possibly large texts to avoid logging a giant text, I could use the same argument you made and say that it's the developers job to truncate **any** result of `Exception::getMessage()` before sending it off to a logger. It's no Guzzle's job to tell me something went wrong and refuse to tell me what went wrong because of some imaginary problem with spamming logger systems.

sagikazarmark:
> As I stated on the linked issue and on Twitter, the behavior you describe breaks the basic concept of `Exception::getMessage()`

I disagree. The message tells you what went wrong:

```
Client Error: `GET /` resulted in a `404 Not Found` response
```

*Why* did it went wrong? The response body might explain, it might not. But semantically, the HTTP error middleware is supposed to turn **HTTP errors** into exceptions. It does that with 100% accuracy.

> So, yes, I can write a bunch of boilerplate everytime I'm using Guzzle because of it, but that's what I'm trying to address.

No, you are trying to push something to the core which poses serious risks.

I've given you a fair amount of solutions. You could write a wrapper library of your own, or just  a module in your own application. I've even given you an example what I would accept as a solution in the Guzzle core.

> I could use the same argument you made and say that it's the developers job to truncate any result of Exception::getMessage() before sending it off to a logger

That's much more error prone and poses serious risks. The core and the default behavior must aim for security by default.

> It's not Guzzle's job to tell me something went wrong and refuse to tell me what went wrong because of some imaginary problem with spamming logger systems.

This is simply incorrect and ignorant. Read my earlier comments why.

I'm sorry this issue makes you upset. I think this is a wrong solution. I've given you some examples that I can imagine as better solutions. I'm not sure what more I can do.

deleugpn:
I already disagreed with all that over and over again. This is precisely my point of making the same arguments over and over again. It's certainly not my desire to be keep debating this forever while getting insulted.
As stated in the original link, I only proceeded with opening this PR because of what Michael Dowling said on Twitter. I will no longer be engaging in this pointless conversation. 
If anyone has anything else to talk about, I'm open to reworking the PR to any feedback provided and I may rebase and resolve the git conflicts if necessary.

sagikazarmark:
> Given this is only a problem in the http_error middleware, I would prefer this to be an option in that middleware, probably with a `RequestException::createWithResponseSize` named constructor. It would require you to instantiate the middleware for yourself, so there is now way other than the owner of the client can change that size limit.

Is this something you would be willing to contribute?

GrahamCampbell:
This need to be re-implemented now, if it is still wanted.

GrahamCampbell:
I have prepared a replacement which doesn't need any global (static) hacks. Instead, creating the http error middleware supports setting a custom body summary function, and the handler stack's create method now has a simple option for setting max length. See https://github.com/guzzle/guzzle/pull/2795.

PR #2273, opened 2019-03-07
Make GuzzleException extend Throwable whereever it's available

ErikBooijCB:
Our static analyzer is giving us some issues when catching Guzzles exceptions, because they don't extend throwable and thus aren't technically catchable.

I get that we don't want to extend it by default because it would break BC with PHP5.x, but I think the proposed solution would give us the best of both worlds.

### DIFF
--- a/src/Exception/GuzzleException.php
+++ b/src/Exception/GuzzleException.php
@@ -1,13 +1,20 @@
 <?php
 namespace GuzzleHttp\Exception;
 
-/**
- * @method string getMessage()
- * @method \Throwable|null getPrevious()
- * @method mixed getCode()
- * @method string getFile()
- * @method int getLine()
- * @method array getTrace()
- * @method string getTraceAsString()
- */
-interface GuzzleException {}
+use Throwable;
+
+if (interface_exists(Throwable::class)) {
+    interface GuzzleException extends Throwable {}
+} else {
+    /**
+     * @method string getMessage()
+     * @method \Throwable|null getPrevious()
+     * @method mixed getCode()
+     * @method string getFile()
+     * @method int getLine()
+     * @method array getTrace()
+     * @method string getTraceAsString()
+     */
+    interface GuzzleException {}
+}
+


src/Exception/GuzzleException.php
```diff
@@ -1,13 +1,29 @@
...
+
+if (interface_exists(Throwable::class)) {
+    /**
+     * @method string getMessage()
+     * @method Throwable|null getPrevious()
+     * @method mixed getCode()
+     * @method string getFile()
+     * @method int getLine()
+     * @method array getTrace()
+     * @method string getTraceAsString()
```
dbu: these annotations should not be needed anymore because the method signatures are provided by [throwable](http://php.net/manual/en/class.throwable.php)

dbu:
just stumbled into this because of phpstan complaining about `@throws GuzzleException` when that is no throwable.

ErikBooijCB:
@dbu @Tobion Thanks for the review and the feedback! I've updated the PR, but a unit test is now failing because of a hardcoded date in the `CookieJarTest` that has passed between the creation of this PR and now.

I have created a separate PR to fix that and as soon as that's merged or the issue is otherwise fixed, I'll rebase this PR to make it pass inspection.

https://github.com/guzzle/guzzle/pull/2278

sagikazarmark:
@ErikBooijCB thanks a lot for the PR. Can you please rebase it?

ErikBooijCB:
@sagikazarmark Absolutely, done 👍

sagikazarmark:
Thanks!

wirwolf:
Hello @ErikBooijCB, i create a fork of this repo and release this marge request into the new version(6.4.0).

src/Exception/GuzzleException.php
```diff
@@ -1,13 +1,20 @@
...
- * @method mixed getCode()
- * @method string getFile()
- * @method int getLine()
- * @method array getTrace()
- * @method string getTraceAsString()
- */
-interface GuzzleException {}
+use Throwable;
+
+if (interface_exists(Throwable::class)) {
```
kronthto: `interface_exists` has a second parameter `autoload` that defaults to true: https://www.php.net/manual/en/function.interface-exists.php

Can we explicitely pass this as `false`? It makes no sense imo to try autoloading a SPL class/interface, and causes issues with bad autoloader implementations that `include`/`require` everything passed to them even when such a file does not exist.

I can make a PR if needed.

Tobion: pr welcome

PR #2274, opened 2019-03-09
 Fix invalid `on_headers` option handling in MockHandler

Horat1us:
Before this fix when use `on_headers` (as in [example](http://docs.guzzlephp.org/en/stable/request-options.html#on-headers)) with any exception in mock queue you will receive TypeError.

Test case without fix:
```bash
There was 1 error:

1) GuzzleHttp\Test\Handler\MockHandlerTest::testEnsuresResponseTypeBeforeOnHeaders
TypeError: Argument 3 passed to GuzzleHttp\Exception\RequestException::__construct() must implement interface Psr\Http\Message\ResponseInterface or be null, instance of Exception given, called in /Projects/guzzle/src/Handler/MockHandler.php on line 85

/Projects/guzzle/src/Exception/RequestException.php:23
/Projects/guzzle/src/Handler/MockHandler.php:85
/Projects/guzzle/tests/Handler/MockHandlerTest.php:164
```

### DIFF
--- a/src/Handler/MockHandler.php
+++ b/src/Handler/MockHandler.php
@@ -97,7 +97,7 @@ class MockHandler implements \Countable
         $this->lastOptions = $options;
         $response = \array_shift($this->queue);
 
-        if (isset($options['on_headers'])) {
+        if (isset($options['on_headers']) && $response instanceof ResponseInterface) {
             if (!\is_callable($options['on_headers'])) {
                 throw new \InvalidArgumentException('on_headers must be callable');
             }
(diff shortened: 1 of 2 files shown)


Horat1us:
Bug introduced in #1580

GrahamCampbell:
Ping @Horat1us. This needs rebasing, and the tests are failing.

Horat1us:
@GrahamCampbell, done.

GrahamCampbell:
Isn't this still not working? It doesn't crash, but the on_headers stuff is not dealt with for async stuff.

Horat1us:
`$this->queue` in MockHandler can contain exception.
My purpose is to check what will be passed to `on_headers`.
Because example in the documentation uses strict argument type (ResponseInterface). And without my fix unexpected TypeError may be thrown.
image

Horat1us:
Same to my type checking is performed on line 116:
image

Horat1us:
Can it be reviewed and merged? Actual version still throws exception with valid configuration.

PR #2278, opened 2019-03-20
Fix unit test that is breaking because of hardcoded cookie expiration date

ErikBooijCB:

PR #2286, opened 2019-04-14
Internationalized domain name (IDN) support

### LINKED ISSUE #2285: Request to cyrillic domain fails

**Description**  

Request to cyrillic domain fails.

**How to reproduce**  

```
$httpClient = new \GuzzleHttp\Client();
$response = $httpClient->get('https://русские-домены.рф');
echo (string)$response->getBody();
```
**Possible Solution**  

It fails due to *Host* header in `GuzzleHttp\Psr7\Request::updateHostFromUri()`
Header must be converted using `idn_to_ascii()` function.

**Additional context**  

But it works using `curl`

```
$ch = curl_init('https://русские-домены.рф');
curl_setopt($ch, CURLOPT_RETURNTRANSFER, true);
$html = curl_exec($ch);
curl_close($ch);
echo $html;
```

alexeyshockov:
This PR introduces IDN support on client's level, so all the handlers will work with the ASCII version of the IDN with no need for additional conversion.

Also fixes #2285.

### DIFF
--- a/src/RequestOptions.php
+++ b/src/RequestOptions.php
@@ -133,10 +133,10 @@ final class RequestOptions
     const HTTP_ERRORS = 'http_errors';
 
     /**
-     * idn: (bool|int, default=true) Set to false to disable exceptions
-     * when a non- successful HTTP response is received. By default,
-     * exceptions will be thrown for 4xx and 5xx responses. This option only
-     * works if your handler has the `httpErrors` middleware.
+     * idn: (bool|int, default=true) A combination of IDNA_* constants for
+     * idn_to_ascii() PHP's function (see "options" parameter). Set to false to
+     * disable IDN support completely, or to true to use the default
+     * configuration (IDNA_DEFAULT constant).
      */
     const IDN_CONVERSION = 'idn_conversion';
 
--- a/src/RequestOptions.php
+++ b/src/RequestOptions.php
@@ -133,6 +133,14 @@ final class RequestOptions
     const HTTP_ERRORS = 'http_errors';
 
     /**
+     * idn: (bool|int, default=true) Set to false to disable exceptions
+     * when a non- successful HTTP response is received. By default,
+     * exceptions will be thrown for 4xx and 5xx responses. This option only
+     * works if your handler has the `httpErrors` middleware.
+     */
+    const IDN_CONVERSION = 'idn_conversion';
+
+    /**
      * json: (mixed) Adds JSON data to a request. The provided value is JSON
      * encoded and a Content-Type header of application/json will be added to
      * the request if no Content-Type header is already present.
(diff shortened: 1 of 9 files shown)


alexeyshockov:
The build will be green when #2278 is merged. Other tests are OK.

alexeyshockov:
Although this PR is self-contained, guzzle/psr7#269 makes IDN support overall better.

sagikazarmark:
@alexeyshockov Thanks! Can you please rebase your branch?

alexeyshockov:
@sagikazarmark, done

src/RequestOptions.php
```diff
@@ -132,6 +132,14 @@ final class RequestOptions
      */
     const HTTP_ERRORS = 'http_errors';
 
+    /**
+     * idn: (bool|int, default=true) Set to false to disable exceptions
```
hsgce: Looks like this docblock needs updating.

tests/bootstrap.php
```diff
@@ -1,4 +1,9 @@
 <?php
+
+namespace {
+    setlocale(LC_ALL, 'C');
```
hsgce: Do we want all tests running in this locale?

src/RequestOptions.php
alexeyshockov: Thanks, fixed

alexeyshockov:
@sagikazarmark, could you take a look also at this for the next release?

tests/bootstrap.php
alexeyshockov: Currently Guzzle fails on other locales. See guzzle/psr7#269 for details, it also solves this.

sagikazarmark:
Yeah, it's on my list, thanks. I recall a related PR in the psr7 package. Is it a dependency of this PR? Do we have to raise the minimum version?

alexeyshockov:
> Is it a dependency of this PR?

No, the PR to PSR7 is not a dependency, but an addition (very convenient, though, because without it Guzzle will struggle with locales, that's why hardcoded it to `'C'` in the tests).

> Do we have to raise the minimum version?

Minimum version of PHP stays the same (IDN support is enabled only when `intl` extension is available).

sagikazarmark:
> Minimum version of PHP stays the same (IDN support is enabled only when intl extension is available).

I meant PSR7 package version, but it's not necessary then.

alexeyshockov:
Yeah, I mixed things with PSR7 package a bit, will prepare a separate PR to `1.x` branch

Nyholm:
I think this looks good. Could you add a test where you use base_uri?

Nyholm:
If you rebase this on master and then add a test where you use an internationalized domain in `base_uri`, then I would be happy to merge.

alexeyshockov:
@Nyholm, done

PR #2291, opened 2019-04-25
Added 'base_uri' constant to RequestOptions

drull755:
It would be cool to use constant instead of string when creating a client


Nyholm:
Thank you for this PR. 

Duplicate of #2315.

PR #2293, opened 2019-04-27
Update `EasyHandle` in order to avoid bad writes of properties related to the handle

phansys:
|Q            |A     |
|---          |---   |
|Branch       |master|
|Bug fix?     |no    |
|New feature? |no    |
|BC breaks?   |no    |
|Deprecations?|no    |
|Tests pass?  |yes   |
|Fixed tickets|n/a   |
|License      |MIT   |
|Doc PR       |n/a   |

With these changes, setting fake values [like this](https://github.com/guzzle/guzzle/pull/2292/files#diff-e61f3c0116b6df977a5a8e4c3df2ccb3R788) can be prevented.

### DIFF
--- a/src/Handler/CurlFactory.php
+++ b/src/Handler/CurlFactory.php
@@ -147,7 +147,7 @@ class CurlFactory implements CurlFactoryInterface
             self::invokeStats($easy);
         }
 
-        if (!$easy->response || $easy->errno) {
+        if (!$easy->response || \CURLE_OK !== $easy->errno) {
             return self::finishError($handler, $easy, $factory);
         }
 
@@ -192,7 +192,7 @@ class CurlFactory implements CurlFactoryInterface
         $factory->release($easy);
 
         // Retry when nothing is present or when curl failed to rewind.
-        if (empty($easy->options['_err_message']) && (!$easy->errno || $easy->errno == 65)) {
+        if (empty($easy->options['_err_message']) && (!$easy->errno || $easy->errno == /* \CURLE_SEND_FAIL_REWIND */ 65)) {
             return self::retryFailedRewind($handler, $easy, $ctx);
         }
 
--- a/src/Handler/EasyHandle.php
+++ b/src/Handler/EasyHandle.php
@@ -11,6 +11,16 @@ use Psr\Http\Message\StreamInterface;
 /**
  * Represents a cURL easy handle and the data it populates.
  *
+ * @property resource|\CurlHandle   $handle                  resource cURL resource
+ * @property StreamInterface        $sink                    Where data is being written
+ * @property array                  $headers                 Received HTTP headers so far
+ * @property ResponseInterface|null $response                Received response (if any)
+ * @property RequestInterface       $request                 Request being sent
+ * @property array                  $options                 Request options
+ * @property int                    $errno                   int cURL error number
+ * @property \Throwable|null        $onHeadersException      Exception during on_headers (if any)
+ * @property \Throwable|null        $createResponseException Exception during createResponse (if any)
+ *
  * @internal
  */
 final class EasyHandle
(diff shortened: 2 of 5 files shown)


Nyholm:
Thank you for the PR. 
I would like you to clarify one thing for me: From whom are you preventing to set fake values?

phansys:
The prevention is done in order to avoid the object to accept a set of inconsistent values, regardless who is writing them. In my specific use case, the situation was caused from a 3rd party library using Guzzle, but I don't have enough details since I think the problem was related to #2292.

src/Handler/CurlFactory.php
```diff
@@ -147,7 +147,7 @@ private static function finishError(
 
         // Retry when nothing is present or when curl failed to rewind.
         if (empty($easy->options['_err_message'])
-            && (!$easy->errno || $easy->errno == 65)
+            && (!$easy->errno || $easy->errno == /* CURLE_SEND_FAIL_REWIND */ 65)
```
GrahamCampbell: When is this constant not set? Was added at least 10 years ago, wasn't it?

phansys: I don't think so. See php/php-src#4079.

GrahamCampbell: Oh, I see. They never made it into PHP-land. ;)

src/Handler/EasyHandle.php
```diff
@@ -100,13 +114,76 @@ public function createResponse(): void
...
+
+    /**
+     * @param string $name
+     *
+     * @throws \LogicException
+     */
+    public function __unset($name)
+    {
+        if ('handle' !== $name) {
+            // BC: Change to `\Error` when bumping PHP version to ^7.0
```
GrahamCampbell, line 183: Guzzle 7 already requires PHP 7.2. :)

GrahamCampbell (changes requested):

YasinAshorian (approved):

YasinAshorian (approved):

GrahamCampbell:
@phansys are you able to resolve the merge conflicts please?

phansys:
>@phansys are you able to resolve the merge conflicts please?

Since this PR is almost 5 years old, I guess I must also update the behavior on some checks, since the added support for PHP >= 8 in the target branch also changed some return values in CURL functions (like `resource` => `CurlHandle`).

src/Handler/EasyHandle.php
```diff
@@ -98,15 +113,84 @@ public function createResponse(): void
...
+                    }
+                }
+            }
+        }
+
+        $this->{$name} = $value;
+    }
+
+    /**
+     * @throws \LogicException
```
mfn: Where does it throw LogicException in `__unset`?

phansys: This exception isn't thrown in the current implementation. The docblock was updated, thank you.

GrahamCampbell:
Looking back at this, I think this PR is obsolete now.

The original motivation seems to have come from the investigation in #2292, where cURL multi failures could produce incomplete error messages. That was traced to PHP bug #77946: `curl_multi_info_read()` returned the failure result, but the returned easy handle did not expose the same state through `curl_errno()` / `curl_error()`. PHP fixed that in php/php-src@c8c183e via php/php-src#4431. The fix shipped in PHP 7.2.22 and 7.3.9, and is present in PHP 7.4+ and all supported PHP 8 versions.

Since Guzzle 7 users can work around this by upgrading to a patch version of their PHP install to get the fix and Guzzle 8 supports `^7.4 || ^8.0` where every supported runtime already has the PHP fix, there's no need to make any further code changes in Guzzle. Given `EasyHandle` is internal, I don’t think adding this magic-property validation is worth the extra complexity.

PR #2309, opened 2019-06-06
stringify Content-Length header in PrepareBodyMiddleware

hutnikau:
Please find detailed description of issue here:
https://github.com/guzzle/psr7/issues/273

Bug may be reproduced only on `master` branch of `guzzlehttp/psr7` package.
So update `composer.json` and set with the following line:
`"guzzlehttp/psr7": "dev-master"`


Tobion:
This is not required anymore with https://github.com/guzzle/psr7/pull/272

PR #2314, opened 2019-06-18
Use PHPUnit static assertions

simPod:
There are static (eg. `self::assertEquals()`) and dynamic (eg. `$this->assertEquals()`  methods in PHPUnit. There's no reason not to call them properly. Also it speeds up a tests a bit.

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -1,7 +1,11 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<phpunit bootstrap="./tests/bootstrap.php"
-         backupGlobals="true"
-         colors="true">
+  <phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+           xsi:noNamespaceSchemaLocation="vendor/phpunit/phpunit/phpunit.xsd"
+           bootstrap="./tests/bootstrap.php"
+           backupGlobals="true"
+           colors="true"
+           executionOrder="random"
+  >
   <testsuites>
     <testsuite name="Test Suite">
       <directory>tests</directory>
--- a/.php_cs
+++ b/.php_cs
@@ -7,6 +7,7 @@ $config = PhpCsFixer\Config::create()
         'array_syntax' => ['syntax' => 'short'],
         'declare_strict_types' => false,
         'concat_space' => ['spacing'=>'one'],
+        'php_unit_test_case_static_method_calls' => ['call_type' => 'self'],
         // 'ordered_imports' => true,
         // 'phpdoc_align' => ['align'=>'vertical'],
         // 'native_function_invocation' => true,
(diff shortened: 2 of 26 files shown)


Nyholm:
Oh, this is a massive PR. 

I assume this is auto generated. =)

It will be easier to enforce this with a CS rule. Lets wait for #2375 to be merged, then rebase this PR (or create a new) with updated CS rules and this auto generated fix.

Nyholm:
Hey @simPod 
Have a look at `master`. I added some PHP-CS-Fixer rules. You can modify the rules to make sure we always use a static call.

simPod:
@Nyholm phpstan has check for that. I'm not aware phpcs is able to check for this (?) It analyses only tokens, amirite? 

But surely I can rebase and fix it for now.

Nyholm:
Checkout "php_unit_test_case_static_method_calls" in https://github.com/FriendsOfPHP/PHP-CS-Fixer.

It might be easier to do a new PR.. I dont know..

.travis.yml
```diff
@@ -25,8 +25,7 @@ matrix:
     - php: hhvm
       dist: trusty
     - php: nightly
-  include:
-    - php: 5.5
+  include:- php: 5.5
```
Nyholm: Please revert

phpunit.xml.dist
```diff
@@ -1,7 +1,11 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<phpunit bootstrap="./tests/bootstrap.php"
-         backupGlobals="true"
-         colors="true">
+  <phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+           xsi:noNamespaceSchemaLocation="vendor/phpunit/phpunit/phpunit.xsd"
+           bootstrap="./tests/bootstrap.php"
+           backupGlobals="true"
+           colors="true"
+           executionOrder="random"
```
Nyholm: Why do we need to change this file?

.travis.yml
simPod: sorry, you're too fast :D

phpunit.xml.dist
simPod: I added schema and also random execution order is IMO good way to detect eventual state between tests.

Nyholm:
I resolved the conflicts with master. 

Can you fix the CS errors and push again please?

simPod:
BTW wouldn't it make more sense to use phpcs instead of that cs fixer?

Nyholm:
I dont think so at the moment. 
Try opening an issue with pros/cons.

gmponos:
Hello here.. sorry for digging this up. I just pulled the repo and saw the current PR.

Long time ago I tried to actually do some private benchmarks if there is an actual performance benefit from this. I remember also finding some posts os S. Bergman saying the same.

https://github.com/sebastianbergmann/phpunit/issues/1914#issuecomment-148407321
https://phpunit.de/manual/6.5/en/appendixes.assertions.html#appendixes.assertions.static-vs-non-static-usage-of-assertion-methods

ofc this is from 2015.. so are there any benchmarks for static/self?

simPod:
Overall, all code bases I fixed in this way run faster about something around 0-10 % (measured through CI). No benchmarks tho.

PR #2326, opened 2019-07-02
Minor perf improvement on PrepareBodyMiddleware

gmponos:
Minor perf improvement on PrepareBodyMiddleware by moving a check if the stream is seekable first.

The improvement is that it is better to check first if the stream is readable rather than checking the size of the stream.

### DIFF
--- a/src/PrepareBodyMiddleware.php
+++ b/src/PrepareBodyMiddleware.php
@@ -97,9 +97,13 @@ class PrepareBodyMiddleware
         // Always add if the body cannot be rewound, the size cannot be
         // determined, or the size is greater than the cutoff threshold
         $body = $request->getBody();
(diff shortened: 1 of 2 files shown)


Nyholm:
> The improvement is that it is better to check first if the stream is readable

Could you elaborate why this is true?

gmponos:
To be fair.. this clearly depends on the implementation of the Stream.. for instance check https://github.com/guzzle/psr7/blob/master/src/Stream.php#L61

It already contains if it is readable/seakable/writable but getting the size and the stats of a file looks more costly. https://github.com/guzzle/psr7/blob/master/src/Stream.php#L123

> this clearly depends on the implementation of the Stream.. 

With that said.. I don't think it's worth the trouble.

PR #2327, opened 2019-07-03
Fix failing tests

gmponos:
- Response status code must be an integer between 1xx and 5xx

### DIFF
--- a/src/RedirectMiddleware.php
+++ b/src/RedirectMiddleware.php
@@ -186,7 +186,7 @@ class RedirectMiddleware
         if ($options['allow_redirects']['referer']
             && $modify['uri']->getScheme() === $request->getUri()->getScheme()
         ) {
-            $uri = $request->getUri()->withUserInfo('', '');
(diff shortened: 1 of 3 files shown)


gmponos:
I am not sure what is the reason of the failing build

andrewnclark:
I'm fairly sure this test `testCreatesGenericErrorResponseException` can just be removed given the current state of the library. It's impossible to create a PSR7 Response with an invalid status code that is below 100 or gte 600 thanks the this method in its constructor: `$this->assertStatusCodeRange($status);`

Execution is stopped before the generic error response that is being tested for is ever created. The question here is whether are happy to let the current `InvalidArgumentException: Status code must be an integer value between 1xx and 5xx.` to bubble up as it does (and remove the rest) or whether we'd prefer to catch that Exception and wrap it in a generic `RequestException` (thus making the test pass).

Personally, I think it's safe to remove this particular test.

gmponos:
closing as this has been resolved from #2349

PR #2340, opened 2019-07-30
Consider body size only in requests which use body

freiondrej:
GET requests do not use body, therefore it does not make sense to check their body size.

The reason for this fix is a strange problem we encountered recently, where a server misinterpreted Content-Length: 0 of a GET request presumably with Content-Range, so it did not send any data. This fix solved the issue for us.

### DIFF
--- a/src/Handler/StreamHandler.php
+++ b/src/Handler/StreamHandler.php
@@ -41,7 +41,7 @@ class StreamHandler
 
             // Append a content-length header if body size is zero to match
             // cURL's behavior.
-            if (0 === $request->getBody()->getSize()) {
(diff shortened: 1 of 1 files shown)


sagikazarmark:
Although `GET` requests do not contain a body _usually_, it's not unprecedented. I know it's not RFC-compliant, but this would actually be a BC break for those clients.

Also, as the comment above indicates, it's there to match the behavior of cURL handler.

I would probably copy the stream handler without that change if I were you. I'll wait for other opinions though.

Nyholm:
I know it is super weird. But GET request can have a body.

PR #2341, opened 2019-07-31
Use composer command instead of php composer.phar

shojaeix:
The default command for composer is `composer` and maybe `php composer.phar` is outdated


sagikazarmark:
Thanks!

Issue #2342, opened 2019-08-01
Can we change this to let JSON pass in the second parameter control?


baicaiit:
Can we change this to let JSON pass in the second parameter control?
![image](https://user-images.githubusercontent.com/42630572/62263930-7a139d80-b450-11e9-8138-ed111dc134fa.png)

andrewnclark:
I can see how this would be useful and achievable by tweaking the options. I'll submit a PR if they're open to the idea.

andrewnclark:
I started work on something but it's possible to do what you need without changes so it's unlikely they will be merged. You request options array should look something like:

```php
[
    'body' => json_encode($data, JSON_UNESCAPED_UNICODE),
    'headers' => [
        'Content-Type' => 'application/json'
    ]
]
```

Nyholm:
Thank you @andrewnclark for your answer!

PR #2343, opened 2019-08-05
Allow json encode options

andrewnclark:
As requested in #2342 this PR will allow for JSON encoding options to used if they are provided. Otherwise, all previous behavior is retained.

Wasn't sure if this was of interest outside the initial issue but decided to throw a quick something together. 

The JSON option block will now accept the following:

```php
'json' => [
    '_options' => JSON_UNESCAPED_UNICODE,
    'body' => ['foo' => 'bar']
]
```

As well as the original array:
```php
'json' => ['foo' => 'bar']
```

I'll update the documentation accordingly if everyone is happy. Feedback is appreciated.

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -27,6 +27,7 @@ class Client implements ClientInterface
     /** @var array Default request options */
     private $config;
 
+
     /**
      * Clients accept an array of constructor parameters.
      *
@@ -321,10 +322,10 @@ class Client implements ClientInterface
 
         if (isset($options['json'])) {
 
-            if (! isset($options['json']['options'])) {
+            if (! isset($options['json']['_options'])) {
                 $options['body'] = \GuzzleHttp\json_encode(isset($options['json']['body']) ? $options['json']['body'] : $options['json']);
             } else {
-                $options['body'] = \GuzzleHttp\json_encode($options['json']['body'], $options['json']['options']);
+                $options['body'] = \GuzzleHttp\json_encode($options['json']['body'], $options['json']['_options']);
             }
 
             unset($options['json']);
--- a/src/Client.php
+++ b/src/Client.php
@@ -320,7 +320,13 @@ class Client implements ClientInterface
         }
 
         if (isset($options['json'])) {
-            $options['body'] = \GuzzleHttp\json_encode($options['json']);
+
+            if (! isset($options['json']['options'])) {
+                $options['body'] = \GuzzleHttp\json_encode(isset($options['json']['body']) ? $options['json']['body'] : $options['json']);
+            } else {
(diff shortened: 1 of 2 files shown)


andrewnclark:
I'm not keen on the underscore myself but going ahead without it could be a breaking change for anyone who has an `options` key in their existing JSON arrays.  I'm open to ideas though if anyone has any.

With regards to docs I can put them together fairly quickly should it get approval. Thanks for the input though.

andrewnclark:
Having took an hour away from the screen I realized that 'body' is just as likely, if not more so, to be an existing key so I introduced a JsonPayload object that will allow options to be provided.

This would allow either of the two options arrays to work with no backwards compatibility issues:

```php
'json' => [
    'foo' => 'bar'
]

'json' => new JsonPayload(['foo' => 'bar'], JSON_UNESCAPED_UNICODE)
```

gmponos:
> it could be a breaking change for anyone who has an options key in their existing JSON arrays

🤔 I haven't thought about this.. I am now having second thoughts as I believe it's not worth the trouble. Someone can json_encode his body and use the body parameter.

andrewnclark:
That is a valid point, a simple `'body' => json_encode($data, JSON_UNESCAPED_UNICODE)` would negate any of these changes, but we could make a similar call on the JSON option in it's entirety.

Thanks again for the input.

vaclavvanik:
Personally I would create new `$options` key called `json_options`.

sagikazarmark:
I wonder if this is actually useful. I'm not sure. I probably wouldn't use it, I'd just encode the body myself. The `json` option is a convenience feature, but creating a whole new "API" for that has some risks.

Personally, I would even consider removing the `json` option in a next major version, in favor of some sort of convenience wrapper. Like `$client->json()->get(...)` or something like that.

That being said, adding a `json_options` is probably a minor change, so if I haven't convinced you to just use `json_encode` instead, I'd probably accept a PR.

Nyholm:
I am a bit more defensive. 

I like the feature because it is cool and convenient. However, this open up a new config key, different values should be supported, what happens if you use `json_options` but not passing a `json` key? etc etc. 

There is a lot of Guzzle code we should write and maintain.. and all because application developers should not write: 
```php
['body'=>json_encode(...)]
```

Im 👎

andrewnclark:
I'm some what in agreement that this PR isn't really necessary. I can update the options to `json_options` and add some logic to validate that a the `json` option is provided when options are, but the json_encode approach feels like the correct way to do this.

Let me know what you all think.

Nyholm:
Im closing this PR due the reasons I already stated. If another maintainer feel differently, feel free to reopen.

PR #2344, opened 2019-08-05
Fix PHPUnit Configuration warning

andrewnclark:
Prevent the following warning being thrown by PHPUnit during testing:

```
  Warning - The configuration file did not pass validation!
  The following problems have been detected:

  Line 6:
  - Element 'testsuite': The attribute 'name' is required but missing.

  Test results may not be as expected.
```


sagikazarmark:
Thanks!

PR #2346, opened 2019-08-08
allow_redirects.referer is false by default

BenMorel:
According to both the [RedirectMiddleware source code](https://github.com/guzzle/guzzle/blob/6.3.3/src/RedirectMiddleware.php#L28) and the [online docs](http://docs.guzzlephp.org/en/stable/request-options.html), `referer` is `false` by default.

This PR fixes the PHP doc.


sagikazarmark:
Thanks @BenMorel !

sagikazarmark:
Breaking build is unrelated to this change, potentially fixed on master.

PR #2348, opened 2019-08-08
[CurlFactory] Prevent undefined offset when using array for ssl_key options

### LINKED ISSUE #2339: Undefined offset: 1 in CurlFactory.php

**Description**  
Currently this is tossing notices in my log files constantly.  I do have a letsencrypt SSL file in place, but guessing this is assuming that something (password?) is populated in `$sslKey[1]` and it's not in my case.  This is a production instance and haven't tried pulling updates yet (but the code appears unchanged in current release on a quick glance).

**How to reproduce**  
Code block:
```            $sslKey = $options['ssl_key'];
            if (is_array($sslKey)) {
                $conf[CURLOPT_SSLKEYPASSWD] = $sslKey[1];
                $sslKey = $sslKey[0];
            }
```
**Possible Solution**  
Not necessarily the best way to do this, but it will squelch all the warnings.  I don't have enough familiarity with the code base to suggest a better fix at this point ... but this does quiet the warnings in my case.
Code block:
```
            $sslKey = $options['ssl_key'];
            if (is_array($sslKey)) {
                $conf[CURLOPT_SSLKEYPASSWD] = @$sslKey[1];
                $sslKey = $sslKey[0];
            }
```
**Additional context**

andrewnclark:
Hopefully resolves #2339 where the `ssl_key` options are provided using the array format, but without a password as the second item in the array.

Still waiting conformation that this was the cause of the warnings, but this will avoid them going forward.

As always, feedback is appreciated.

### DIFF
--- a/src/Handler/CurlFactory.php
+++ b/src/Handler/CurlFactory.php
@@ -454,11 +454,12 @@ class CurlFactory implements CurlFactoryInterface
         }
 
         if (isset($options['ssl_key'])) {
-            $sslKey = $options['ssl_key'];
-            if (is_array($sslKey)) {
-                $conf[CURLOPT_SSLKEYPASSWD] = $sslKey[1];
-                $sslKey = $sslKey[0];
+            if (is_array($options['ssl_key']) && count($options['ssl_key']) === 2) {
+                list($sslKey, $conf[CURLOPT_SSLKEYPASSWD]) = $options['ssl_key'];
             }
+
+            $sslKey = isset($sslKey) ? $sslKey: $options['ssl_key'];
+
             if (!file_exists($sslKey)) {
                 throw new \InvalidArgumentException(
                     "SSL private key not found: {$sslKey}"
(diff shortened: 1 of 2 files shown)


andrewnclark:
Has since been confirmed that the cause was options defined as such:

```php
$options = [
    'ssl_key' => ['/some/file/path']
]
```

This PR will prevent further warnings but the alternative is to validate properly that the options are both provided in the array or use a single string, thus making the above invalid and prompting a new release and update instructions.

sagikazarmark:
For maximum compatibility, I think it should be possible to use the array format without a key (even if it's not intended to be supported). We can deprecate that behavior though, and remove it in the next major release.

@andrewnclark are you up for making that change?

sagikazarmark:
Also, can you please rebase your branch from master?

Nyholm:
Thank you for the reviews and for the fix

PR #2371, opened 2019-10-11
Added a note to Overview to mention cURL

vfsoraki:
Without cURL, concurrency is not available. See https://bit.ly/33fA7RU and https://bit.ly/2IFVG6o for some SO questions.


vfsoraki:
Do please reword my sentences. English is not my first language.

sagikazarmark:
Thanks @vfsoraki , I updated your changes a bit to be shorter.

PR #2373, opened 2019-10-18
Require latest version of guzzlehttp/psr7

Nyholm:
To avoid any issues with older versions.


Nyholm:
Thank you for the review.

PR #2375, opened 2019-10-18
Static analysis with Github actions

### LINKED ISSUE #2239: Automatic coding style checks and formatting
- [ ] Add PrettyCI
- [ ] Add php-cs-fixer integration

Related #2167, #2194

Nyholm:
This introduced PHPStan, PHP-cs-fixer and Roave BC checker. 

Will fix #2239

### DIFF
--- a/.github/workflows/bc.yml
+++ b/.github/workflows/bc.yml
@@ -1,0 +1,10 @@
+on: [push, pull_request]
+name: Roave
+jobs:
+  roave_bc_check:
+    name: BC Check
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
+      - name: Roave BC Check
+        uses: docker://nyholm/roave-bc-check-ga
--- a/.github/workflows/static.yml
+++ b/.github/workflows/static.yml
@@ -1,0 +1,22 @@
+on: [push, pull_request]
+name: Static analysis
+jobs:
+  phpstan:
+    name: PHPStan
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
+      - name: PHPStan
+        uses: docker://oskarstark/phpstan-ga
+        with:
+          args: analyze --no-progress
+
+  php-cs-fixer:
+    name: PHP-CS-Fixer
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
+      - name: PHP-CS-Fixer
+        uses: docker://oskarstark/php-cs-fixer-ga
+        with:
+          args: --dry-run --diff-format udiff
--- a/.php_cs
+++ b/.php_cs
@@ -1,0 +1,21 @@
+<?php
+
+$config = PhpCsFixer\Config::create()
+    ->setRiskyAllowed(true)
+    ->setRules([
+        '@PSR2' => true,
+        'array_syntax' => ['syntax' => 'short'],
+        'declare_strict_types' => false,
+        'concat_space' => ['spacing'=>'one'],
+        // 'ordered_imports' => true,
+        // 'phpdoc_align' => ['align'=>'vertical'],
+        // 'native_function_invocation' => true,
+    ])
+    ->setFinder(
+        PhpCsFixer\Finder::create()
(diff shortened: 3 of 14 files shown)


.github/workflows/bc.yml
```diff
@@ -0,0 +1,10 @@
+on: [push, pull_request]
+name: Roave
+jobs:
+  roave_bc_check:
+    name: BC Check
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
+      - name: Roave BC Check
+        uses: docker://nyholm/roave-bc-check-ga
```
Ocramius: Can this be versioned?

.github/workflows/static.yml
```diff
@@ -0,0 +1,22 @@
+on: [push, pull_request]
+name: Static analysis
+jobs:
+  phpstan:
+    name: PHPStan
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
+      - name: PHPStan
+        uses: docker://oskarstark/phpstan-ga
```
Ocramius: Can this be versioned?

.github/workflows/static.yml
```diff
@@ -0,0 +1,22 @@
...
+        with:
+          args: analyze --no-progress
+
+  php-cs-fixer:
+    name: PHP-CS-Fixer
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
+      - name: PHP-CS-Fixer
+        uses: docker://oskarstark/php-cs-fixer-ga
```
Ocramius: Similar question here: sember tags are kinda needed

.php_cs
```diff
@@ -0,0 +1,21 @@
+<?php
+
+$config = PhpCsFixer\Config::create()
+    ->setRiskyAllowed(true)
+    ->setRules([
+        '@PSR2' => true,
```
Ocramius: Is PSR-12 already viable?

phpstan.neon.dist
```diff
@@ -0,0 +1,9 @@
+parameters:
+    level: 1
```
Ocramius: This sounds like a really painfully low analysis level: is it maybe possible to use psalm with aggressive configuration and a baseline instead? That would at least prevent new bugs from sneaking through.

.php_cs
Nyholm: Not sure.

phpstan.neon.dist
Nyholm: Yeah, I would like to do that. Im just taking it slow with one PR at the time.

.github/workflows/bc.yml
Nyholm: It can, but why? I would like to have all the warnings possible. Even if roave BC check updates their major version.

phpstan.neon.dist
Ocramius: Fair enough - my idea was to replace phpstan for this kind of projects: no baseline means very ineffective static analysis

.github/workflows/bc.yml
Ocramius: While indeed nice to have latest and greatest, tools like CS and SA would then break for random contributors jumping in

sagikazarmark: I agree, pipelines should be as static as possible. What's working today, should work tomorrow as well. We need to update these versions regularly. Same for actions, like `actions/checkout@master`. It should be `@v1`.

phpstan.neon.dist
simPod: @Ocramius there's [baseline for PHPStan](https://medium.com/@ondrejmirtes/phpstans-baseline-feature-lets-you-hold-new-code-to-a-higher-standard-e77d815a5dff) now too. Is this still valid?

> my idea was to replace phpstan for this kind of projects

Ocramius: Yeah, my comment was put in before I knew phpstan added baseline support: I suggest going with `--level=max` and a baseline config.

rvanlaak:
This also fixed #2095 and #1971 👏

PR #2378, opened 2019-10-23
Decouple from the UriInterface implementation

alexeyshockov:
A small change to make Guzzle and the current `UriInterface` implementation from `psr7` package decoupled. The main goal currently is to be able to change the implementation (see guzzle/psr7#296) without any changes in `guzzle` main package itself.


Nyholm:
Im not sure what the win is..
Let's what for https://github.com/guzzle/psr7/pull/296.

Tobion:
Closing as I don't see what this brings. And https://github.com/guzzle/psr7/pull/296 is closed anyway

PR #2383, opened 2019-10-23
Do not fail on BC break failure

Nyholm:
This adds an extra entrypoint file that suppresses our two BC breaks.


Nyholm:
Thank you for the review

PR #2385, opened 2019-10-23
Some automated CS fixes

Nyholm:
They are all safe ;)

### DIFF
--- a/.php_cs
+++ b/.php_cs
@@ -7,13 +7,14 @@ $config = PhpCsFixer\Config::create()
         'array_syntax' => ['syntax' => 'short'],
         'declare_strict_types' => false,
         'concat_space' => ['spacing'=>'one'],
-        // 'ordered_imports' => true,
+        'ordered_imports' => true,
         // 'phpdoc_align' => ['align'=>'vertical'],
         // 'native_function_invocation' => true,
     ])
     ->setFinder(
         PhpCsFixer\Finder::create()
             ->in(__DIR__.'/src')
+            ->in(__DIR__.'/tests')
             ->name('*.php')
     )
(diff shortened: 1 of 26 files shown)


.php_cs
```diff
@@ -7,13 +7,14 @@ $config = PhpCsFixer\Config::create()
         'array_syntax' => ['syntax' => 'short'],
         'declare_strict_types' => false,
         'concat_space' => ['spacing'=>'one'],
-        // 'ordered_imports' => true,
+        'ordered_imports' => true,
```
Nyholm: This is just because we want it to look good

.php_cs
```diff
@@ -7,13 +7,14 @@ $config = PhpCsFixer\Config::create()
...
         // 'phpdoc_align' => ['align'=>'vertical'],
-        // 'native_function_invocation' => true,
+        'native_function_invocation' => true,
```
Nyholm: This is actually a security issue

.php_cs
```diff
@@ -7,13 +7,14 @@ $config = PhpCsFixer\Config::create()
...
     ])
     ->setFinder(
         PhpCsFixer\Finder::create()
             ->in(__DIR__.'/src')
+            ->in(__DIR__.'/tests')
```
Nyholm: This should be enabled because we do care about tests too

Nyholm:
I could not use native function calls because we are actually overwriting `curl_multi_setopt` and `curl_setopt_array` in our tests

sagikazarmark:
> They are all safe ;)

Promise? 😄

Nyholm:
Promise!

PR #2387, opened 2019-10-23
Use PHPStan baseline

Nyholm:
Follow up from @simPod and @Ocramius comments here:  https://github.com/guzzle/guzzle/pull/2375#discussion_r338275795

This will help use write less bugs in the future.

### DIFF
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -2,6 +2,6 @@ includes:
     - phpstan-baseline.neon
 
 parameters:
-    level: 7
+    level: max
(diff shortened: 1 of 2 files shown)


phpstan.neon.dist
```diff
@@ -1,9 +1,7 @@
+includes:
+    - phpstan-baseline.neon
+
 parameters:
-    level: 1
+    level: 7
```
Ocramius: I think PHPStan has level 8 too now?

PR #2392, opened 2019-10-26
Fixed branch alias

GrahamCampbell:


Nyholm:
Thank you.
It should be 6.5 right? That is the next release

PR #2393, opened 2019-10-27
Cover case where curl_multi_select return 0

### LINKED ISSUE #2390: curl_multi_select can return immediately ignoring timeout, with result 0 and error code 0
**Guzzle version(s) affected**: unknown

**Description**  
The implementation of curl_multi_select changed in PHP 7.1.23/7.2.11 - It uses curl_multi_wait instead of curl_multi_select. See https://bugs.php.net/bug.php?id=76480

When this happens, requests will typically succeed, but spike to 100% cpu while waiting for requests to complete (because nothing is sleeping, not even curl_multi_select)

- It might be possible that some settings in Guzzle prevent this, but unlikely.
- This issue being filed may be related to issues reported after 2018-09

**How to reproduce**  
This may only affect php >= 7.1.23 and curl >= 7.28.0. Use getAsync methods, where one endpoint takes a long time to respond (and another has a connect timeout). You may have to run this test repeatedly and log the number of times curl_multi_exec gets called.

**Possible Solution**  
Call usleep(250 microseconds - microseconds elapsed since previous call to curl_multi_select ended) if curl_multi_select returns <= 0? A proposed fix to a different library is https://github.com/rmccue/Requests/pull/284 - The symptoms seen prior to this is https://gist.github.com/TysonAndre/46a6829d9d8de00e50226890a6a9a63f (both curl_multi_select and curl_multi_exec return 0 but many times don't respect the configured timeout of 50 milliseconds)

- It already sleeps for a return value of -1.

**Additional context**  
I observed that repeatedly calling curl_multi_select and curl_multi_exec 

The guzzle implementation using curl_multi_select without handling a return value of 0 is:

https://github.com/guzzle/guzzle/blob/8f10d0e/src/Handler/CurlMultiHandler.php#L117-L130

Nyholm:
This will fix #2390

FYI @TysonAndre

### DIFF
--- a/src/Handler/CurlMultiHandler.php
+++ b/src/Handler/CurlMultiHandler.php
@@ -117,12 +117,15 @@ class CurlMultiHandler
         // Step through the task queue which may add additional requests.
         P\queue()->run();
 
-        if ($this->active &&
-            curl_multi_select($this->_mh, $this->selectTimeout) === -1
-        ) {
-            // Perform a usleep if a select returns -1.
-            // See: https://bugs.php.net/bug.php?id=61141
-            usleep(250);
+        if ($this->active) {
+            $fd = curl_multi_select($this->_mh, $this->selectTimeout);
+            if ($fd === -1 || $fd === 0) {
+                // Perform a usleep if a select returns -1.
+                // See: https://bugs.php.net/bug.php?id=61141
+                // Select will return 0 instead of -1 since PHP 7.1.23 and 7.2.11
+                // See: https://bugs.php.net/bug.php?id=76480
+                usleep(250);
+            }
         }
 
         while (curl_multi_exec($this->_mh, $this->active) === CURLM_CALL_MULTI_PERFORM);


src/Handler/CurlMultiHandler.php
```diff
@@ -117,12 +117,15 @@ public function tick()
...
         P\queue()->run();
 
-        if ($this->active &&
-            curl_multi_select($this->_mh, $this->selectTimeout) === -1
-        ) {
-            // Perform a usleep if a select returns -1.
-            // See: https://bugs.php.net/bug.php?id=61141
-            usleep(250);
+        if ($this->active) {
+            $fd = curl_multi_select($this->_mh, $this->selectTimeout);
```
TysonAndre: I'd rather have this patch ensure that this block of code sleeps for at least 250 microseconds, instead of sleeping for an additional 250 microseconds.

Before, it'd sleep for 1.000 seconds in the case that curl_multi_select normally handled no activity (and didn't unexpectedly exit earlier). After this, it'd sleep 1.00025 seconds.

- If a request took exactly  1.00010 seconds, it'd sleep unnecessarily for an extra 150 microseconds

```php
            $select_start = microtime(true);
            $fd = curl_multi_select($this->_mh, $this->selectTimeout);
            $elapsed_milliseconds = (microtime(true) - $select_start) * 1000000;
            if ($fd === -1 || $fd === 0) {
                // Perform a usleep if a select returns -1.
                // See: https://bugs.php.net/bug.php?id=61141
                // Select will return 0 instead of -1 for newer libcurl versions since PHP 7.1.23 and 7.2.11
                // See: https://bugs.php.net/bug.php?id=76480
                $sleep_time = 250 - $elapsed_milliseconds;
                if ($sleep_time >= 1) {
                    usleep($sleep_time);
                }
            }
```

TysonAndre:
Any thoughts on my suggestion?

GrahamCampbell:
Ping @Nyholm. :)

PR #2401, opened 2019-11-15
Automated native_function_invocation fixes

draco2003:
Reproduce with:
php php-cs-fixer --rules=native_function_invocation fix src --allow-risky=yes
(php-cs-fixer from : https://cs.symfony.com/)

Details:
https://veewee.github.io/blog/optimizing-php-performance-by-fq-function-calls/

### DIFF
--- a/.php_cs
+++ b/.php_cs
@@ -10,7 +10,7 @@ $config = PhpCsFixer\Config::create()
         'php_unit_test_case_static_method_calls' => ['call_type' => 'self'],
         'ordered_imports' => true,
         // 'phpdoc_align' => ['align'=>'vertical'],
-        // 'native_function_invocation' => true,
+        'native_function_invocation' => true,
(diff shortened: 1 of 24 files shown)


Nyholm:
I agree with this change. But we cannot do it since the tests rely on overwriting some custom functions.

gmponos:
How about using the `compiler_optimized` option?

https://mlocati.github.io/php-cs-fixer-configurator/#version:2.16|fixer:native_function_invocation

gmponos:
Not sure about this but I think namespacing every function has a misbenefit.. symfony is also using the compiler optimized... if I understood correctly this ruleset: https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.16/src/RuleSet.php#L195

Nyholm:
Fixed in #2444

PR #2402, opened 2019-11-18
Prevent trying to autoload Throwable

kronthto:
PR for comment in https://github.com/guzzle/guzzle/pull/2273#discussion_r346277358 .

It will prevent `interface_exists` from attempting to autoload the SPL interface, which is bound to fail.

With an autoloader implementation that doesn't properly handle not finding a class it *can* cause warnings (had this issue with the Magento1 autoloader).

The only risk I can think of is very abstract, as I don't think it's common that PHP5 environments would provide a polyfill for `\Throwable` in autoloaded code.

I didn't figure out how to write a relevant test for this change.

### DIFF
--- a/src/Exception/GuzzleException.php
+++ b/src/Exception/GuzzleException.php
@@ -3,7 +3,7 @@ namespace GuzzleHttp\Exception;
 
 use Throwable;
 
-if (interface_exists(Throwable::class)) {
+if (interface_exists(Throwable::class, false)) {
     interface GuzzleException extends Throwable
     {
     }


GrahamCampbell:
Hmmm. I'd say the autoloader should not be crashing the application like that. Warning for classes that don't exist is silly because testing if a class exists is so common.

riccardomessineo:
This is exactly the PR I was going to submit. Had the same issue with Yii.

sagikazarmark:
> The only risk I can think of is very abstract, as I don't think it's common that PHP5 environments would provide a polyfill for `\Throwable` in autoloaded code.

Are there any references supporting that claim? I'm not saying it's wrong, but I can't say from the top of my head that they don't.

kronthto:
`symfony/polyfill-php70` for one doesn't do it, the [`\Error` stub](https://github.com/symfony/polyfill-php70/blob/master/Resources/stubs/Error.php) extends `\Exception` and they don't provide a `\Throwable`.

That's of course not proving some don't.

sagikazarmark:
After thinking this through, I agree with @GrahamCampbell, an autoloader should not fail when a class does not exist.

My primary concern is breaking polyfills that do provide a `Throwable` stub.

An easy fix is providing a stub in an environment where the autoloader can't handle the situation, but I think it should be an edge case.

Nyholm:
According to the [PSR4 autoloader specification](https://www.php-fig.org/psr/psr-4/)

> 4. Autoloader implementations MUST NOT throw exceptions, MUST NOT raise errors of any level, and SHOULD NOT return a value.

I know it sucks that Magento decided to build an autoloader that does not comply with PSR-4. However, I know that newer versions of Magento will fix this issue. =)

I will close this issue since it is not related to Guzzle. Mark also provided a workaround.

PR #2405, opened 2019-11-19
Added additional check in withCookieHeader.

Miljan9602:
Added check for Host header. If check does not match domain from uri, cookie will not be added.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,5 +1,5 @@
 {
-    "name": "guzzlehttp/guzzle",
+    "name": "Miljan9602/guzzle",
     "type": "library",
     "description": "Guzzle is a PHP HTTP client library",
     "keywords": [
--- a/composer.json
+++ b/composer.json
@@ -1,5 +1,5 @@
 {
-    "name": "Miljan9602/guzzle",
+    "name": "miljan9602/guzzle",
     "type": "library",
     "description": "Guzzle is a PHP HTTP client library",
     "keywords": [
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -1,5 +1,5 @@
 {
-    "name": "guzzlehttp/guzzle",
+    "name": "miljan9602/guzzle",
```
Nyholm: Oh, please dont rename the package =)

nemanjan00:
We have no idea how exacly we would test for something like this. 

Problem happens when we want to send request to different reverse proxy, on purpose. 

We change address to `http://reverseproxyaddress/` and add header `Host: originalolaction` and, cookies for `reverseproxyaddress` are used, instead of those for `originallocation`

GrahamCampbell:
I don't understand the use case here. Why would a cookie for a different domain be added?

nemanjan00:
In some test cases, we want to pick specific reverse rpoxy instance. 

If we specify that reverse proxy in host, we need to specify Domain header, so reverse proxy knows what domain we are requesting, and, we also want to use cookies for that domain.

Nyholm:
Friendly ping

GrahamCampbell:
This is also missing tests.,

src/Cookie/CookieJar.php
```diff
@@ -279,10 +279,15 @@ public function withCookieHeader(RequestInterface $request)
         $scheme = $uri->getScheme();
         $host = $uri->getHost();
         $path = $uri->getPath() ?: '/';
+        $headerHost = null;
+
+        if (sizeof($request->getHeader('Host')) > 0) {
```
GrahamCampbell, line 284: When can there be no Host header?

GrahamCampbell (changes requested):

PR #2412, opened 2019-12-01
Update .gitattributes

reedy:


.gitattributes
```diff
@@ -1,10 +1,12 @@
...
-.gitattributes    export-ignore
-/.github/         export-ignore
-.gitignore        export-ignore
-/.travis.yml      export-ignore
-/build/           export-ignore
-/docs/            export-ignore
-/Makefile         export-ignore
-/phpunit.xml.dist export-ignore
-/tests/           export-ignore
+/.editorconfig         export-ignore
```
sagikazarmark: Some files, like editorconfig and gitignore were ignored from the entire tree, not just the root on purpose

reedy: None of the files exist in any subdirectory though

PR #2421, opened 2019-12-07
Implement PSR-18

Nyholm:
I drop old PHP versions. Update PHPUnit and Implement PSR-18

This will replace #2122

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,35 +2,31 @@ language: php
 
 sudo: false
 
+env:
+  COMPOSER_FLAGS="--no-interaction --prefer-source --dev"
+
 php:
-  - 5.6
-  - 7.0
-  - 7.1
   - 7.2
   - 7.3
   - 7.4snapshot
-  - nightly
 
 before_script:
   - curl --version
-  - composer install --no-interaction --prefer-source --dev
+  - composer install $COMPOSER_FLAGS
   - ~/.nvm/nvm.sh install v0.6.14
   - ~/.nvm/nvm.sh run v0.6.14
-  - if [[ "$TRAVIS_PHP_VERSION" != "7.0" || "$TRAVIS_PHP_VERSION" != "7.1" ]]; then echo "xdebug.overload_var_dump = 1" >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini || true; fi
+  - echo "xdebug.overload_var_dump = 1" >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini || true
 
 script: make test
 
 matrix:
-  allow_failures:
-    - php: hhvm
-      dist: trusty
-    - php: nightly
-  include:
-    - php: 5.5
-      dist: trusty
-    - php: hhvm
-      dist: trusty
   fast_finish: true
+  allow_failures:
+    - php: nightly
+      env: COMPOSER_FLAGS="--no-interaction --prefer-source --dev --ignore-platform-reqs"
+  include:
+    - php: nightly
+      env: COMPOSER_FLAGS="--no-interaction --prefer-source --dev --ignore-platform-reqs"
 
 before_deploy:
   - make package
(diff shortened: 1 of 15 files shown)


gmponos:
Is this PR only for PSR-18.. ? For instance there is one more deprecation on one exception... We will do the separately after 6.5 tag?

.travis.yml
```diff
@@ -2,35 +2,31 @@ language: php
...
+env:
+  COMPOSER_FLAGS="--no-interaction --prefer-source --dev"
+
 php:
-  - 5.6
-  - 7.0
-  - 7.1
   - 7.2
   - 7.3
   - 7.4snapshot
```
gmponos: I think snapshot is not needed anymore? Am I right?

Nyholm: It was needed 4 days ago

Nyholm:
I will release 6.5 and then start working on 7.0. So this will be the first PR I merge after 6.5 is released.

src/Client.php
```diff
@@ -130,6 +131,18 @@ public function send(RequestInterface $request, array $options = [])
         return $this->sendAsync($request, $options)->wait();
     }
 
+    /**
+     * The HttpClient PSR (PSR-18) specify this method.
+     */
+    public function sendRequest(RequestInterface $request): ResponseInterface
```
gmponos: Thinking out loud here.. without having the answer on my head.. 

Should this be final?

.travis.yml
Nyholm: Nice!

Nyholm:
This is just the first PR. 
We should add more integration tests and add PHP7.2 code. But one step at the time

README.md
```diff
@@ -79,12 +79,15 @@ composer update
...
 [guzzle-3-repo]: https://github.com/guzzle/guzzle3
 [guzzle-4-repo]: https://github.com/guzzle/guzzle/tree/4.x
 [guzzle-5-repo]: https://github.com/guzzle/guzzle/tree/5.3
 [guzzle-6-repo]: https://github.com/guzzle/guzzle
+[guzzle-7-repo]: https://github.com/guzzle/guzzle
 [guzzle-3-docs]: http://guzzle3.readthedocs.org
-[guzzle-5-docs]: http://guzzle.readthedocs.org/en/5.3/
-[guzzle-6-docs]: http://guzzle.readthedocs.org/en/latest/
+[guzzle-5-docs]: http://docs.guzzlephp.org//en/5.3/
+[guzzle-6-docs]: http://docs.guzzlephp.org//en/6.5/
```
sagikazarmark:
```suggestion
[guzzle-6-docs]: http://docs.guzzlephp.org//en/6.x/
```

If the branch is  6.x.

src/functions.php
```diff
@@ -133,7 +133,7 @@ function default_user_agent()
     static $defaultAgent = '';
 
     if (!$defaultAgent) {
-        $defaultAgent = 'GuzzleHttp/' . Client::VERSION;
+        $defaultAgent = 'GuzzleHttp/Guzzle';
```
sagikazarmark: Maybe add `7.x` here?

README.md
Nyholm: I created a 6.5 branch already. 

My idea is that Guzzle 7 is a so easy upgrade so there is no reason to add more features to 6.x

src/functions.php
Nyholm: Then we still need to worry about the version. For what benefit?

GrahamCampbell: You could always do the trick php-cs-fixer does to detect its version?

GrahamCampbell: https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v2.16.1/src/ToolInfo.php#L61-L72

Nyholm:
Thank you for the reviews.

src/functions.php
Nyholm: That is pretty cool. But Im not sure we want the version at all. 

Anyhow. A followup PR may fix this.

PR #2424, opened 2019-12-08
Add IDN support for redirects also

gmponos:


gmponos:
if the corrent PR is a solution for issue #2424 then we should consider centralizing this logic..

src/RedirectMiddleware.php
```diff
@@ -190,7 +191,41 @@ public function modifyRequest(
...
         }
 
-        $modify['uri'] = $this->redirectUri($request, $response, $protocols);
+        $uri = $this->redirectUri($request, $response, $protocols);
+
+        if ($uri->getHost() && isset($options['idn_conversion']) && ($options['idn_conversion'] !== false)) {
+            $idnOptions = ($options['idn_conversion'] === true) ? IDNA_DEFAULT : $options['idn_conversion'];
+
+            $asciiHost = idn_to_ascii($uri->getHost(), $idnOptions, INTL_IDNA_VARIANT_UTS46, $info);
+            if ($asciiHost === false) {
```
alexeyshockov: What do you think about extracting this logic to a function/helper class? Looks like we have at least 2 places now where it's used, worth extracting.

gmponos: ofc

> if the corrent PR is a solution for issue #2424 then we should consider centralizing this logic..

gmponos:
btw I am targeting 6.5 as a bugfix.

src/functions.php
```diff
@@ -344,3 +346,44 @@ function _current_time()
...
     return function_exists('hrtime') ? hrtime(true) / 1e9 : microtime(true);
 }
+
+
+/**
+ * @param int $options
+ * @return UriInterface
+ * @internal
+ */
+function _idn_uri_convert(UriInterface $uri, $options = IDNA_DEFAULT)
```
alexeyshockov: What do you think about changing `IDNA_DEFAULT` constant to `0` here? I'm think when `intl` is not available it can cause an error...

See roundcube/roundcubemail#6244 for details, also related to #2448.

src/functions.php
```diff
@@ -344,3 +346,44 @@ function _current_time()
...
+{
+    if ($uri->getHost()) {
+        $asciiHost = idn_to_ascii($uri->getHost(), $options, INTL_IDNA_VARIANT_UTS46, $info);
```
alexeyshockov: It make sense to use `$variant = defined('INTL_IDNA_VARIANT_UTS46') ? INTL_IDNA_VARIANT_UTS46 : null;` workaround here, because it seems that the constant is not always available.

src/functions.php
gmponos: Done.

src/functions.php
gmponos: Done.

gmponos:
@alexeyshockov thank you for the review.. can you check again?

src/functions.php
gmponos: Done.

src/functions.php
```diff
@@ -344,3 +346,47 @@ function _current_time()
...
+ *
+ * @return UriInterface
+ *
+ * @internal
+ */
+function _idn_uri_convert(UriInterface $uri, $options = 0)
+{
+    if ($uri->getHost()) {
+        $variant = defined('INTL_IDNA_VARIANT_UTS46') ? INTL_IDNA_VARIANT_UTS46 : 0;
+        $asciiHost = idn_to_ascii($uri->getHost(), $options, $variant, $info);
```
alexeyshockov: Just checked locally, it raises a deprecation error (`idn_to_ascii(): INTL_IDNA_VARIANT_2003 is deprecated`).

The only way is to call the function without the third argument at all, like [here](https://github.com/TYPO3/TYPO3.CMS/commit/3a2b54a1b36b99b525cc04eccf20fd018a24e4a6#diff-cb4f296a39c2f11c2784e0e940745b82R193-R199):
```
$asciiHost = defined('INTL_IDNA_VARIANT_UTS46')
                ? idn_to_ascii($uri->getHost(), $idnOptions, INTL_IDNA_VARIANT_UTS46, $info)
                : idn_to_ascii($uri->getHost(), $idnOptions);
```

gmponos: thanks again..  done..

alexeyshockov: OK, so it's night and too many versions to keep in the head.

It works on my machine, because I have PHP 7.4 with already right defaults.

On 7.3 it raises a warning by default, if you don't specify `INTL_IDNA_VARIANT_UTS46` manually. So there is basically no way to get rid of this deprecation error in PHP 7.2+ if you have an old version of ICU library. Looks like we have to disable `idn_conversion` by default in this case (PHP 7.2+ and `INTL_IDNA_VARIANT_UTS46` not defined).

alexeyshockov: @gmponos, now I think it's better to open a separate PR for this change. Please take a look at #2453.

gmponos: better we avoid the back and forth :)

alexeyshockov: Agree, but now it seems really out of scope here

sagikazarmark:
So, how does this relate to #2454

Should the two be shipped in 6.5.1?

alexeyshockov:
> So, how does this relate to #2454 ?

Initially (when the issue with `INTL_IDNA_VARIANT_UTS46` was discovered) I thought it can be fixed quickly here, in this PR. After additional information from affected devs and some checks I ended up with creating a separate PR for this change.

So the last changes (connected to `INTL_IDNA_VARIANT_UTS46`) should be removed from here. Sorry again for confusion, @gmponos.

> Should the two be shipped in 6.5.1?

Don't see reasons not to do that. Or ship #2454 first and this one after, in `6.5.2`. Whatever works for you.

sagikazarmark:
Thanks @alekseykuleshov !

If @gmponos can make the required changes we can release these fixes together.

gmponos:
I will later today or tomorrow...

gmponos:
@alexeyshockov I merged your changes from 6.5 branch.

Would appreciate if you could re-validate this.

sagikazarmark:
@gmponos Can you please enable the intl extension in GH actions ci.yaml? I just saw that there is a condition for the intl extension being loaded.

gmponos:
done on https://github.com/guzzle/guzzle/pull/2478 since current branch has tests under travis

PR #2429, opened 2019-12-08
Use anonymous class

gmponos:


Nyholm:
Could you see why Travis is failing?

gmponos:
hmm.. it's not worth the trouble and one of the tests is now removed...

Nyholm:
Hm. Okey.

PR #2431, opened 2019-12-08
Phpunit 8 deprecated expectException annotations

gmponos:


Nyholm:
Thank you. 
Can you see why Travis fails?

gmponos:
Now it is passing :)

FYI will replace assertContains on next PR.

PR #2438, opened 2019-12-08
Add GitHub Actions

sagikazarmark:


.travis.yml
```diff
@@ -10,6 +10,12 @@ php:
   - 7.3
   - 7.4
 
+branches:
```
Nyholm: You dont need to add this. 

You can safely remove the travis.yml. The test will still run on these branches since there is a travis.yml there.

sagikazarmark: DPne

PR #2439, opened 2019-12-08
Deprecate functions

Nyholm:
These global functions are sure handy, but it is a code spell. 
Lets deprecate them and provide a better solution for Guzzle 8.

### DIFF
--- a/src/functions.php
+++ b/src/functions.php
@@ -37,6 +37,8 @@ function uri_template($template, array $variables)
  *
  * @return string Returns a string containing the type of the variable and
  *                if a class is provided, the class name.
+ *
+ * @deprecated Will be removed in 8.0.0
  */
 function describe_type($input)
 {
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Odd to deprecate them without a replacement yet?

Nyholm:
I think all of these are internal.

Yeah, you might be correct. There is no way for a user to **not** use deprecated code. 

I just wanted a PR to say: "This is not the way we should write code in the future".

GrahamCampbell:
> These global functions are sure handy, but it is a code spell.

They are not global, actually. They are namespaced?

sagikazarmark:
What would be a better alternative? TBH I never understood that why static functions of a class is a better approach than just plain functions. It's unlikely that you will mock these functions.

Nyholm:
Excellent question. =)

Static functions on a class will **show** that there are a dependency here. With these functions in the current state will hide complexity and dependencies away for the reader.

Nyholm:
Im closing this. I should write an issue instead.

PR #2440, opened 2019-12-08
Remove UriTemplate

Nyholm:
It is never used :/


GrahamCampbell:
Probably was used by the AWS SDK?

Nyholm:
Thank you
I just checked, AWS SDK is not using `UriTemplate`

GrahamCampbell:
Probs fine to kill it then. :)

gmponos:
We forgot it in the previous PRs.. but should we start adding notes to [UPGRADING.md](https://github.com/guzzle/guzzle/blob/master/UPGRADING.md) ?

Also what is the alternative here if someone has been using this?

For instance some one must have in his composer `ext-uri-template: "*"`?

Do you believe it would be out of the scope of guzzle keeping this on a repo as `guzzlehttp/uri-template` ?

Nyholm:
This class and function was never used. So I dont see any reasons for Guzzle org to keep it around. 

------

Im just finishing a PR. Then I will write changelog and upgrade docs.

gmponos:
> This class and function was never used. So I don't see any reasons for Guzzle org to keep it around.

For me this depends on the scope of the library. Even if it is not used internally in the project, it does not exactly justify the removal. As a package you have classes to be used out of it.. not inside of it.. in the same way that `Client` class is not used anywhere inside the project... It is meant to be used outside of it.

So UriTemplate initially might have been created with this purpose.

Tried to search about the reasons that lead adding this class but I couldn't find anything from the history: https://github.com/guzzle/guzzle/commits/master/src/UriTemplate.php

Nyholm:
> For me this depends on the scope of the library.

That is very true. The Guzzle library is a HTTP client, not a HTTP (or PSR7) utilities. 

I am not sure why `UriTemplate` was created or if it has been used. 

-----

I suggest to remove this class. If there is a need for it in the future, lets create a new package with this class.

sagikazarmark:
The UriTemplate predates the PSR-7 package, but I'm not sure what its original purpose was. I think we can remove it from here. We can always add it back to PSR-7 if needed (the PSR-7 package isn't strictly about PSR-7 anyway, whether it's good or bad is a different question).

Nyholm:
Thank you

jeremeamia:
Just an FYI: The history on this was that it was used by https://github.com/guzzle/guzzle-services in the guzzle 2/3 era.

GrahamCampbell:
Thanks @jeremeamia. We actually decided to keep the class available, but just not in the guzzle core: https://github.com/guzzle/uri-template. :)

PR #2442, opened 2019-12-08
Added type annotation

### LINKED ISSUE #2441: Add PHP 7.2 type annotations
We should add them on all code where it is possible.

Nyholm:
This will fix #2441

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -141,26 +141,6 @@ parameters:
 			path: src/Cookie/CookieJar.php
 
 		-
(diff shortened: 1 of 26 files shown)


gmponos:
As a first step I believe it is fine.. what we can do as next steps is add return types and typehints to...
- `anonymous functions`
- `Tests`
- Then enable php-cs-rule https://mlocati.github.io/php-cs-fixer-configurator/#version:2.15|fixer:phpdoc_to_return_type

Nyholm:
Yeah, This is a first step. I know one can go further... But I'll leave that work for a future PR. Potentially by someone else. =)

Thank you for the review.

PR #2222, opened 2018-11-25
Require ext-json

gmponos:
You have some `json_decode` functions at various places... 

PHPstorm suggests to require `ext-json`

I have also added to composer config to sort packages.

PR #2443, opened 2019-12-08
Update our extension requirements

Nyholm:
From https://www.php.net/manual/en/json.installation.php

> As of PHP 5.2.0, the JSON extension is bundled and compiled into PHP by default.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,12 +22,12 @@
     ],
     "require": {
         "php": "^7.2.5",
-        "ext-json": "*",
         "guzzlehttp/promises": "^1.0",
         "guzzlehttp/psr7": "^1.6.1",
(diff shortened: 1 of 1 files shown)


gmponos:
Well.. I was the one that added the json extension here https://github.com/guzzle/guzzle/pull/2222.. 

FYI: https://github.com/moneyphp/money/pull/530#issuecomment-444901364

TBH it clearly depends on the maintainer.. the only thing I would argue is leave it here for not having to decline future PRs adding it back 😄

Nyholm:
Oh, I see. 

I though that there were no json "extension" and that it was in core. 
I see now that it is just a default extension.

Nyholm:
Thank you

PR #2444, opened 2019-12-08
Fixed native_function_invocation CS

Nyholm:
This will replace #2401

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -67,7 +67,7 @@ class Client implements ClientInterface, \Psr\Http\Client\ClientInterface
     {
         if (!isset($config['handler'])) {
             $config['handler'] = HandlerStack::create();
-        } elseif (!is_callable($config['handler'])) {
+        } elseif (!\is_callable($config['handler'])) {
             throw new \InvalidArgumentException('handler must be a callable');
         }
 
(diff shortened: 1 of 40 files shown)


gmponos:
Shouldn't the php_cs file contain changes?

Nyholm:
No, because we cannot use this all over the place. 

See `bootstrap.php`: We rely on overriding these functions in the tests. 

```php
// Override curl_setopt_array() and curl_multi_setopt() to get the last set curl options
namespace GuzzleHttp\Handler {
    function curl_setopt_array($handle, array $options)
    {
        if (!empty($_SERVER['curl_test'])) {
            $_SERVER['_curl'] = $options;
        } else {
            unset($_SERVER['_curl']);
        }
        return \curl_setopt_array($handle, $options);
    }

    function curl_multi_setopt($handle, $option, $value)
    {
        if (!empty($_SERVER['curl_test'])) {
            $_SERVER['_curl_multi'][$option] = $value;
        } else {
            unset($_SERVER['_curl_multi']);
        }
        return \curl_multi_setopt($handle, $option, $value);
    }
}

```

gmponos:
I would have approved this.. but I am still wondering why symfony (hope I have understood correctly) does not prefix all the functions.. 

I am not saying we should follow it just because they do it.. but I am wondering if there is a performance reason that we are not aware or if it is for BC reason (easier MRs)

Nyholm:
I don’t 100% know. But I’m sure it is because Symfony wants to avoid merge conflicts when merging bug fixes between the 3-4 active branches.

GrahamCampbell:
Yes, probably. Also, once you prefix a function, it makes it impossible to mock, so things like `microtime` can no longer be mocked by the strategy of defining a function in the other namespace of the same name, if it has been prefixed with a slash.

Nyholm:
Note also that "impossible to mock" means that it is more secure sine no evil package maintainer may change the functionality of native methods.

GrahamCampbell:
> Note also that "impossible to mock" means that it is more secure sine no evil package maintainer may change the functionality of native methods.

Well, yeh. But also, impossible for the package's own tests to mock functions, if they wanted to.

Nyholm:
Thank you for the reviews and discussion.

PR #2445, opened 2019-12-08
Move php-http/client-integration-tests to require-dev

Nyholm:
oops :(


Nyholm:
Wow, Github actions is way quicker than Travis =)

PR #2446, opened 2019-12-09
Fix documented return type in GuzzleHttp\Pool::promise()

jaikdean:
Our static analysis is complaining that the return type of `GuzzleHttp\Pool::promise()` is documented incorrectly as `GuzzleHttp\GuzzleHttp\Promise\Promise`. It looks like there was a missing backslash to move to the root namespace.

### DIFF
--- a/src/Pool.php
+++ b/src/Pool.php
@@ -72,7 +72,7 @@ class Pool implements PromisorInterface
     /**
      * Get promise
      *
-     * @return \GuzzleHttp\Promise\Promise
+     * @return PromiseInterface
      */
     public function promise()
     {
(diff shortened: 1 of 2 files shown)


jaikdean:
I'm unsure why the BC check is failing, as the errors are unrelated to anything changed in this PR. I'm not familiar with the CI setup here so don't know if the failure is expected or a known issue.

src/Pool.php
```diff
@@ -72,7 +72,7 @@ public function __construct(
     /**
      * Get promise
      *
-     * @return GuzzleHttp\Promise\Promise
```
Nyholm: Could you just do `PromiseInterface`?

jaikdean:
@Nyholm I wasn't actually testing `master`, this bug is in the current release version :) Should I retarget this PR at the `6.5` branch?

PR #2311, opened 2019-06-13
Fix logging with a string rejection

thewilkybarkid:
Promises aren't always rejected with an exception. I haven't changed `MessageFormatter` as that'd be a BC break.

PR #2447, opened 2019-12-09
Fix logger error when mock file is not found

ankahla:
Fix error:
```
Argument 3 passed to GuzzleHttp
MessageFormatter::format() must be an instance of Exception or null, string given, called in /var/www/vendor/guzzlehttp/guzzle/src/Middleware.php on line 199
```

### DIFF
--- a/src/Middleware.php
+++ b/src/Middleware.php
@@ -197,7 +197,8 @@ final class Middleware
                         $response = $reason instanceof RequestException
                             ? $reason->getResponse()
                             : null;
-                        $message = $formatter->format($request, $response, $reason);
+                        $exception = $reason instanceof RequestException ? $reason : null;
+                        $message = $formatter->format($request, $response, $exception);
                         $logger->notice($message);
                         return \GuzzleHttp\Promise\rejection_for($reason);
                     }


src/Middleware.php
```diff
@@ -197,7 +197,8 @@ function ($reason) use ($logger, $request, $formatter) {
                         $response = $reason instanceof RequestException
                             ? $reason->getResponse()
                             : null;
-                        $message = $formatter->format($request, $response, $reason);
+                        $exception = $reason instanceof RequestException ? $reason : null;
+                        $message = $formatter->format($request, $response, $exception);
```
gmponos: I have a question. Why `$reason` is string.

So if reason shouldn't be a string we could typehint like this:

`function (\Exception $reason) use ($logger, $request, $formatter) {`

if reason could be a string then I would suggest to improve this block of code with the anonymous function because we are checking twice about the exception

Also note that the formatter does not accept only `RequestException` but `Exceptions` in general.

ankahla: `$reason` is mixed (there is no typehint in the project) so theoretically it can be anything, but it is a string if we create a `RejectedPromise` with a string. example https://github.com/csarrazi/guzzle-cache-middleware/blob/master/src/MockMiddleware.php#L47

The formatter accept only null|Exception as third argument so we have to check the type of `$reason` before calling the formatter.

what optimisation do you suggest ?

Nyholm:
I think #2311 is trying to solve the same issue. 

They are solving it in a different way. What is the pros/cons which each solution?

ankahla:
I don't know why the BC check fail
I prefer the other solution #2311 because here I throw away the reason of failure if it is not an exception object

Nyholm:
BC fails becuase we already merged BC breaking changes to master. They will go to 7.0. 

Could you review/test and comment on #2311 and then close this PR?

ankahla:
I approuved #2311

PR #2449, opened 2019-12-10
Added return types to many anonymous functions

gmponos:


src/HandlerStack.php
```diff
@@ -34,8 +34,6 @@ class HandlerStack
      * @param callable $handler HTTP handler function to use with the stack. If no
```
GrahamCampbell: This doc is wrong. Should be `?callable` or `callable|null`. I don't know which style you prefer.

gmponos: Nice catch.. fixed.

GrahamCampbell: 🚀

PR #2453, opened 2019-12-12
Better defaults for PHP installations with old ICU lib

### LINKED ISSUE #2448: INTL_IDNA_VARIANT_UTS46 not defined having ICU < 4.6
**Guzzle version(s) affected**: 6.5.0

**Description**  
`IDN conversion failed` exception being thrown when dealing with domain name having regular ASCII characters (no IDN conversion required at all). Exception is thrown at https://github.com/guzzle/guzzle/blob/6.5.0/src/Client.php#L241

**How to reproduce**  
Requires old version of ICU on e.g. PHP 7.2 (these two things are independent - unfortunately).

**Possible Solution** 
Omit `INTL_IDNA_VARIANT_UTS46` in case it's not available - at https://github.com/guzzle/guzzle/blob/6.5.0/src/Client.php#L221

**Additional References**

See the following references and how it was handled in the TYPO3 CMS project
* https://github.com/TYPO3/TYPO3.CMS/commit/3a2b54a1b36b99b525cc04eccf20fd018a24e4a6#diff-cb4f296a39c2f11c2784e0e940745b82R193-R199
* https://forge.typo3.org/issues/87953 (additional description & system details)

alexeyshockov:
Fixes #2448 and #5326

### DIFF
--- a/docs/request-options.rst
+++ b/docs/request-options.rst
@@ -561,7 +561,7 @@ idn_conversion
 :Types:
     - bool
     - int
-:Default: ``true`` if ``intl`` extension is available, ``false`` otherwise
+:Default: ``true`` if ``intl`` extension is available (and ICU library is 4.6+ for PHP 7.2+), ``false`` otherwise
 :Constant: ``GuzzleHttp\RequestOptions::IDN_CONVERSION``
 
 .. code-block:: php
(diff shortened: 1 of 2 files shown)


alexeyshockov:
@Nyholm, please take a look

alexeyshockov:
Sorry, not the right branch. Just opened another one to `6.5`... See #2454.

PR #2454, opened 2019-12-12
Better defaults for PHP installations with old ICU lib

### LINKED ISSUE #2448: INTL_IDNA_VARIANT_UTS46 not defined having ICU < 4.6

alexeyshockov:
Fixes #2448

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -274,7 +274,7 @@ class Client implements ClientInterface
             && (
                 defined('INTL_IDNA_VARIANT_UTS46')
                 ||
-                version_compare(PHP_VERSION, '7.2.0') < 0
+                PHP_VERSION_ID < 70200
             );
 
         // Use the standard Linux HTTP_PROXY and HTTPS_PROXY if set.
--- a/src/Client.php
+++ b/src/Client.php
@@ -218,7 +218,8 @@ class Client implements ClientInterface
         if ($uri->getHost() && isset($config['idn_conversion']) && ($config['idn_conversion'] !== false)) {
             $idnOptions = ($config['idn_conversion'] === true) ? IDNA_DEFAULT : $config['idn_conversion'];
 
-            $asciiHost = idn_to_ascii($uri->getHost(), $idnOptions, INTL_IDNA_VARIANT_UTS46, $info);
+            $idnaVariant = defined('INTL_IDNA_VARIANT_UTS46') ? INTL_IDNA_VARIANT_UTS46 : 0;
+            $asciiHost = idn_to_ascii($uri->getHost(), $idnOptions, $idnaVariant, $info);
             if ($asciiHost === false) {
                 $errorBitSet = isset($info['errors']) ? $info['errors'] : 0;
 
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
But what if the function was implemented by a pollyfill?

alexeyshockov:
@GrahamCampbell, don't see any issues with a polyfill. If it implements `idn_to_ascii()`, then all should be fine. [`symfony/polyfill` implements both the function and `INTL_IDNA_VARIANT_UTS46` constant](https://github.com/symfony/polyfill/blob/master/src/Intl/Idn/bootstrap.php#L37), so `idn_conversion` option will be `true` by default.

GrahamCampbell:
Maybe guzzle should always load the polyfill?

alexeyshockov:
I think it will not help here. The original issue with `6.5.0` is about old version of ICU library (`intl` extension is available, just "partially"). So `symfony/polyfill` will not work in this case, because it discriminates things by just checking for `idn_to_ascii()` existence.

Creating a polyfill for exactly this case seems to be very challenging, and IMO Guzzle should not force any polyfills usage to the end developer.

src/Client.php
```diff
@@ -267,7 +268,14 @@ private function configureDefaults(array $config)
...
 
         // idn_to_ascii() is a part of ext-intl and might be not available
-        $defaults['idn_conversion'] = function_exists('idn_to_ascii');
+        $defaults['idn_conversion'] = function_exists('idn_to_ascii')
+            // Old ICU versions don't have this constant, so we are basically stuck (see https://github.com/guzzle/guzzle/pull/2424
+            // and https://github.com/guzzle/guzzle/issues/2448 for details)
+            && (
+                defined('INTL_IDNA_VARIANT_UTS46')
+                ||
+                version_compare(PHP_VERSION, '7.2.0') < 0
```
brandonkelly: This would be faster:

```suggestion
                PHP_VERSION_ID < 70200
```

johncarlson21:
guys this needs to be merged like now.. after my update, now I have a non working script. I'm going to manually implement this fix.

Mark-H:
This PR indeed fixes the issue I reported in #2458.

sagikazarmark:
I assume this has to be forward ported to master as well, right?

alexeyshockov:
@sagikazarmark, correct

jonnott:
...so when is 6.5.1 shipping? ;)

jonnott:
Can 6.5.1 be expedited ... ? Currently anyone running CentOS 6 has broken code if using Guzzle :(

sagikazarmark:
@jonnott you can downgrade to 6.4 for the time being. 6.5.1 will be released today or tomorrow.

sagikazarmark:
@alexeyshockov can you send this (or the appropriate version) patch to master as well?

jonnott:
> @jonnott you can downgrade to 6.4 for the time being. 6.5.1 will be released today or tomorrow.

I temporarily added 'idn_conversion'=>false to all my Guzzle\Client instances instead!

PR #2461, opened 2019-12-15
Disallow null uri as value

gmponos:
I am not sure if this is something we want for v7 or we should raise a deprecation notice on 7 and target this PR on 8


sagikazarmark:
We should probably include change log or upgrade guide or both in these PRs, otherwise we will forget these removed deprecations.

gmponos:
I already started another PR about upgrade guide.. will push it there... what is your suggested procedure that we must follow?

We can merge that one and then start doing update of the CHANGELOG/UPGRADE per PR?

related PR #2473

Nyholm:
We can merge this PR now. Good thing that you added this PR to the "Breaking changes in 7" project.

PR #2462, opened 2019-12-15
Added visiblity to all constants and php-cs rule for visibility

gmponos:


src/Handler/CurlFactory.php
```diff
@@ -15,8 +15,8 @@
  */
 class CurlFactory implements CurlFactoryInterface
 {
-    const CURL_VERSION_STR = 'curl_version';
-    const LOW_CURL_VERSION_NUMBER = '7.21.2';
+    protected const CURL_VERSION_STR = 'curl_version';
+    protected const LOW_CURL_VERSION_NUMBER = '7.21.2';
```
gmponos: Made this protected since class is not final someone might need them if he extends `CurlFactory` for whatever reason

src/RequestOptions.php
```diff
@@ -31,7 +31,7 @@ final class RequestOptions
      *   response that was received, and the effective URI. Any return value
      *   from the on_redirect function is ignored.
      */
-    const ALLOW_REDIRECTS = 'allow_redirects';
+    public const ALLOW_REDIRECTS = 'allow_redirects';
```
gmponos: These do not add much value.. changes are made for consistency.

src/Handler/CurlFactory.php
Nyholm: Why not public?

gmponos: I don't have a good argument.. So let's make them public and maybe we can revise for the visibility on v8

gmponos: I don't have a good argument.. So let's make them public and maybe we can revise for the visibility on v8

gmponos:
I will also need to add a phpcs rule for this.

src/Handler/CurlFactory.php
gmponos: Done.

gmponos:
I've added also the php-cs rule

PR #2464, opened 2019-12-15
Remove deprecated request options

gmponos:
Are these considered deprecated that should be removed for v7?


tests/ClientTest.php
```diff
@@ -191,7 +191,7 @@ public function testRewriteSaveToToSink()
         $mock = new MockHandler([new Response(200, [], 'foo')]);
         $client = new Client(['handler' => $mock]);
         $client->get('http://foo.com', ['save_to' => $r]);
-        self::assertSame($r, $mock->getLastOptions()['sink']);
+        self::assertSame($r, $mock->getLastOptions()['save_to']);
```
sagikazarmark: I guess these tests can be removed.

gmponos: Done.. my PhpStorm could not find these options for some reason :/

PR #2466, opened 2019-12-15
Fixes from typehints of the code

gmponos:


gmponos:
Hello...

Initially I had committed only this until this commit https://github.com/guzzle/guzzle/pull/2466/commits/19061715b240a83be3409b60871521c20346645a but then I thought that it is not worth creating a separate PR for this.. So I pushed more changes here.

Summary:

- Add more typehints
- Add more docblocks.
- Clears more phpstan issues.

src/MessageFormatter.php
```diff
@@ -61,7 +61,7 @@ public function __construct(?string $template = self::CLF)
      *
      * @param RequestInterface  $request  Request that was sent
      * @param ResponseInterface $response Response that was received
```
gmponos: I should have changed also the response typehint...

GrahamCampbell:
Add an (array) cast at line 72 to fix the phpstan warning.

gmponos:
@GrahamCampbell I believe that casting this to an array will hide a possible error.

The error was reported because for some weird reason that I am still investigating I've run the baseline command and it automatically removed those ignore errors although it was on a code that I haven't touched.

I tried the changes here to affect in the least possible way the behavior. So I believe we should see this on a different PR.

phpstan-baseline.neon
```diff
@@ -160,6 +160,11 @@ parameters:
 			count: 1
 			path: src/Cookie/SetCookie.php
 
+		-
+			message: "#^Property GuzzleHttp\\\\Cookie\\\\SetCookie\\:\\:\\$data \\(array\\) does not accept array\\|null\\.$#"
```
GrahamCampbell: :-1: Add a cast in line 72 of the code.

GrahamCampbell:
> The error was reported because for some weird reason that I am still investigating

it's because that function call may return null as well as an array, at line 72.

gmponos:
>> The error was reported because for some weird reason that I am still investigating

> it's because that function call may return null as well as an array, at line 72.

@GrahamCampbell That was not my full sentence.

Let me elaborate.. I've run the baseline command locally and for some reason it has removed from the baseline file the ignore error above.. and then I committed the file with the ignored error removed... Hence it was reported on the CI. 

Baseline file has tons of errors that you can comment in any PR. Same here.. The comment is out of the scope of my changes. 

Never the less you could say it is an error that it must be fixed.. why not do it now? Problem is that I don't know if having to cast with array hides any future error that should be reported.. Personally I would feel more confident having a check if it is null to throw an exception. But again this is out of the scope of the changes I wanted to apply here... I hope this is clear.. 

I promise that I will be the next first thing that I will try to fix from the baseline.. or maybe you can do it..  but personally I would go with the exception solution.. If @sagikazarmark or @Nyholm also feel that the array cast can be added here safely let me know...

sagikazarmark:
Sorry, I'm quite out of context here, so I will leave that to @gmponos and @Nyholm

Nyholm:
I agree it is an issue. Since @gmponos dont want to fix the existing issue in this PR I guess it is up for grabs for everybody.

PR #2467, opened 2019-12-16
Fix documented return type for Pool::promise() and Client::__call()

jaikdean:
Our static analysis is complaining that the return type of `GuzzleHttp\Pool::promise()` and `GuzzleHttp\Client::__call()` are documented incorrectly.

This replaces #2446 as that targeted the wrong branch.

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -1126,16 +1126,6 @@ parameters:
 			path: src/Pool.php
 
 		-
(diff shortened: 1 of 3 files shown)


jaikdean:
@gmponos I'm not sure that link is working properly for me, sorry. All I can see on there is the comment from @Nyholm about changing from `GuzzleHttp\Promise\Promise` to `GuzzleHttp\Promise\PromiseInterface`, which is already covered in this PR.

gmponos:
I am from my mobile.. so I might be wrong... but I though that the comment was about using a use statement...

jaikdean:
@gmponos Ah, I see. I've updated the PR to use `use` statements rather than FQCNs.

nickdnk:
Please merge and release :)

PR #2471, opened 2019-12-18
Be strict about array_replace that may return null

gmponos:
@GrahamCampbell as promised this fixes this comment : https://github.com/guzzle/guzzle/pull/2466#discussion_r358159204

Are we good with this?


GrahamCampbell:
I don't think the code can ever fail the way we're calling it?

Nyholm:
It looks like we are using both a belt and suspenders here...

Better safe than sorry I guess.

PR #2473, opened 2019-12-18
UPGRADE file for version 7

gmponos:


gmponos:
This is a draft for review we can update it if more things come up.

UPGRADING.md
```diff
@@ -1,6 +1,37 @@
...
+Moreover:
+- All internal native functions calls of Guzzle are now prefixed with a slash. 
+This change makes it impossible for the users of the library to change the behavior of any of these functions
+
+Before:
+`curl_version()`
+
+After:
+`\curl_version()`
+
```
alexeyshockov:
```suggestion

- [PSR-18](https://www.php-fig.org/psr/psr-18/) support.
```

I think it's worth to mention this

UPGRADING.md
```diff
@@ -1,6 +1,37 @@
...
+ 
+- Class `GuzzleHttp\UriTemplate` is removed.
+- Class `GuzzleHttp\Exception\SeekException` is removed.
+- Constant `GuzzleHttp\ClientInterface::VERSION` is removed.
+- Function `GuzzleHttp\Exception\RequestException::getResponseBodySummary` is removed. Use `\GuzzleHttp\Psr7\get_message_body_summary`
+as an alternative
+- Request option `exception` which was an alias of `http_errors` is removed.
+- Request option `save_to` which was an alias of `sink` is removed.
+
+For the full diff you can check [here](https://github.com/guzzle/guzzle/compare/6.5.0..master)
```
Nyholm: This links needs to be updated. Write (https://github.com/guzzle/guzzle/compare/6.5.0..7.0.0)

UPGRADING.md
gmponos: WDYT if on the first paragraph we had this: 
```
In order Guzzle to take advantage of the new features of PHP, and support [PSR-18](https://www.php-fig.org/psr/psr-18/) 
we dropped the support of PHP 5. The minimum supported PHP version is now PHP 7.2.
```

The reason is that Upgrade guide should have only the details that you need to update your version. The rest should be on the changelog as we currently have them...

PR #2476, opened 2019-12-19
Remove an impossible condition

TysonAndre:
The param type hint ensures the value is always a string inside the
method.

(Detected by phan and the phpstan baseline)

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -111,13 +111,8 @@ parameters:
 			path: src/Cookie/CookieJar.php
 
 		-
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
> (Detected by phan)

Was also detected by phpstan, as your PR shows. ;)

TysonAndre:
> Was also detected by phpstan, as your PR shows. ;)

I realized that after the first commit - I haven't worked with phpstan baselines much. There was one false positive in phpstan due to phpstan treating phpdoc and real types the same way.

PR #2480, opened 2019-12-20
Composer normalize

sagikazarmark:
Add composer normalize check

*Please don't squash this PR*


sagikazarmark:
It's a composer plugin, so I don't think it would work with a phar. It's also a dev dependency.

PR #2482, opened 2019-12-21
Deprecate Client::__call()

Nyholm:
I propose to merge this after 7.0.0 release.

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -84,6 +84,8 @@ class Client implements ClientInterface, \Psr\Http\Client\ClientInterface
      * @param array  $args
      *
      * @return PromiseInterface|ResponseInterface
+     *
+     * @deprecated Client::__call() will be removed in guzzlehttp/guzzle:8.0. Use the actual function instead of calling a magic one.
      */
     public function __call($method, $args)
     {


gmponos:
IMHO I wouldn't like the helper functions to be removed.. Instead I would advise to be strictly declared and create them inside the client.

alexeyshockov:
Agree with @gmponos, `->get()`, `->post()`,.. are very useful IMO and should stay. Implementation via `__call()` should be changed to usual methods though.

GrahamCampbell:
I think we should add these functions to the ClientInterface, since people tend to want to call them anyway, even when they only have something implementing the interface.

GrahamCampbell:
I've created https://github.com/guzzle/guzzle/pull/2513, which I think is a good solution. By providing a trait, we have effectively not really added more stuff to interface, from the perspective of someone looking to implement the interface for themselves. All they need to do is implement the old interface, and then "use" the trait.

gmponos:
> I think we should add these functions to the ClientInterface

I just saw that I had upvoted this... my bad..  IMHO I am not in favor of moving these functions to an interface... Neither having them as trait and allowing developers to use it.

GrahamCampbell:
Developers use the dynamic methods anyway, even when they only know they have a `ClientInterface`. This way, we're just making it "official". Critically, we are making Guzzle easy to use.

GrahamCampbell:
Posted the question to twitter, in an unbiased way: https://twitter.com/GrahamJCampbell/status/1211338797314646020.

GrahamCampbell:
Feedback so far:

image

Nyholm:
Yeah. I saw that. 
Let’s deprecate the magic function and add proper ones. Just like you’ve done. 

Thank you.

GrahamCampbell:
Excellent. I really think keeping these functions will help Guzzle appeal to new comers, and also help exiting users upgrade to Guzzle 7 when it comes out. 🚀

gmponos:
Just to be clear here... if I started the guzzle client from zero.. I would definetely not have these functions.. the only reason that I said that these should stay is that they are being used by the community and it will bring a lot of headaches... not because the community uses them makes it valid...

I hope that promoting PSR 18 into the client developers will go away from these functions..

Promoting these functions to an interface makes them more official and in a way suggested...

So yes in the future I would like to revise this and remove them...

At the moment I dont see neither a strong argument for removing neither a strong argument (at least for me) to promote them...

GrahamCampbell:
I guess we kinda have different visions. I am interesting in making the package as easy to use as possible. To me, this doesn't mean just saying it's an implementation of the PSR HTTP client spec. Guzzle has to differentiate itself from the crowd to stay the most popular choice, so it really doesn't make sense to throw away niceties that attract new users.

GrahamCampbell:
I think we only have a strong argument for leaving them in.

sagikazarmark:
I don't see why we should remove those functions.

There are two major use cases for an HTTP Client:

1a. Write a nice wrapper around it for your HTTP endpoint, handle deserialization into DTOs, etc.
1b. Generate that wrapper
2. Just give me whatever is available on that URL.

I can't count the number of PHP scripts (actual scripts, not full applications) I wrote just to download a file and process it somehow. I never bothered using PSR-7/PSR-18 related stuff in those cases. Just installed Guzzel, `->get()`, done.

There is essentially no maintenance cost.

I think this PR was based on a confusion anyway, at least we talked about this internally and we only wanted to remove the magic function, not the whole functionality. Can't find the discussion though.

alexeyshockov:
> I can't count the number of PHP scripts (actual scripts, not full applications) I wrote just to download a file and process it somehow. I never bothered using PSR-7/PSR-18 related stuff in those cases. Just installed Guzzel, ->get(), done.

This is 100% true, Guzzle has managed to combine ease of use and powerful features (line promises and concurrent execution) in one package. Ease of use is what makes it the most popular client there.

gmponos:
> I guess we kinda have different visions. I am interesting in making the package as easy to use as possible. To me, this doesn't mean just saying it's an implementation of the PSR HTTP client spec. Guzzle has to differentiate itself from the crowd to stay the most popular choice, so it really doesn't make sense to throw away niceties that attract new users.

Everyone wants this.. Who will say the opposite.. That's not the point... Someone can use the same argument for not removing the `getConfig` because it makes it easy for him and attractive.. or even worse use this argument to add `json` or `toString` method to `GuzzleHttp\Psr7\Response`.. 

Just because something makes it easy and attractive does not make it correct...  There are more factors that need to be taken into consideration.

> I think this PR was based on a confusion anyway

Yes I believe the same.. what I disagree is moving them to an abstract layer/interface.

sagikazarmark:
> Someone can use the same argument for not removing the `getConfig` because it makes it easy for him and attractive

Get config leaks internal concerns. There is nothing obviously wrong with HTTP methods in the client.

> what I disagree is moving them to an abstract layer/interface.

The `ClientInterface` is not really an abstraction in this case. It does not abstract implementation details away, it's not there to allow providing alternative implementations. It probably once was though: before the `MockHandler` I think we had a `MockClient`.

 I think the `ClientInterface` is only there to allow mocking the client. From this perspective not adding the new methods to the interface would be a mistake.

With PSR-18 in the picture I don't really see a major requirement to keep our interface/keep our interface "pure".

There is no obvious best solution in this case.

GrahamCampbell:
@sagikazarmark's comments were exactly my thinking for both PRs. I want Guzzle to be easy to use and attractive to new comers, but also to encourage good code by not exposing its internals.

gmponos:
closing as we all agree that we wont remove the helper function and the discussion continues here: https://github.com/guzzle/guzzle/pull/2513

PR #2483, opened 2019-12-21
Merge 6.5 to master

gmponos:
Only this once ... and for the last time I believe that pushing 6.5 to master is worth it.

If it's not worth it let's create another PR.. but most probably it will be a soft reset of this PR.


src/functions.php
```diff
@@ -312,3 +314,47 @@ function _current_time()
 {
     return \function_exists('hrtime') ? \hrtime(true) / 1e9 : \microtime(true);
 }
+
+
```
Nyholm: Remove extra line

gmponos: fixed.

Nyholm:
Is there anything else you want to include in this PR?

PR #2485, opened 2019-12-21
Fix phpstan complains about false errors

gmponos:


gmponos:
For some weird reason PHPStan was complaining about these on master: https://github.com/guzzle/guzzle/runs/359402989

We had this PR merged before in order to fix them: https://github.com/guzzle/guzzle/pull/2471

also this error was removed: https://github.com/guzzle/guzzle/pull/2485/files#diff-cbcd91b5fa0a393ebd972f66cc47dc90L538

which initially it seems to have been cause again by an array_replace here: https://github.com/guzzle/guzzle/blob/master/src/Handler/StreamHandler.php#L298

The most interesting part is that I had this weird change in behavior locally too.. I am not sure if this is related in anyway with github actions or with PHPStan.. Also in GH actions I can not see the PHPStan version. So I am not sure what version it is used on each case...

src/Cookie/SetCookie.php
```diff
@@ -69,12 +69,7 @@ public static function fromString(string $cookie): self
      */
     public function __construct(array $data = [])
     {
-        $replaced = \array_replace(self::$defaults, $data);
-        if ($replaced === null) {
```
sagikazarmark: Why did you remove this check?

Nyholm: If you revert this change, what will happen then?

gmponos: Phpstan will complain :/

gmponos: You can check master failing build.

sagikazarmark: PHPStan probably things `array_replace` returns an array, not null. Removing the check is not a good idea IMHO, because the function *can* return null, in which case we should throw an exception.

gmponos: OK.. I will leave the exception there.. My major problem here is that this error was reported two days ago.. and two days ago we merged this throw exception. What happened now with PHPStan and it false reports that array_replace returns only `array`?

gmponos: > Removing the check is not a good idea IMHO

reverted..

Nyholm:
CI is green, code is great.  =)

PR #2494, opened 2019-12-23
Created JsonEncoder class to overwrite json encoding with di

fhink:
This allows users to implement a custom encoder that could implement the symfony serializer for example.

### DIFF
--- a/src/Encoder/JsonEncoderInterface.php
+++ b/src/Encoder/JsonEncoderInterface.php
@@ -6,17 +6,17 @@ interface JsonEncoderInterface
 {
     /**
      * @param mixed $value
-     * @param int $options
-     * @param int $depth
+     * @param int   $options
+     * @param int   $depth
      * @return string
      */
     public function encode($value, int $options = 0, int $depth = 512): string;
 
     /**
      * @param string $json
-     * @param bool $assoc
-     * @param int $depth
-     * @param int $options
+     * @param bool   $assoc
+     * @param int    $depth
+     * @param int    $options
      * @return mixed
      */
     public function decode(string $json, bool $assoc = false, int $depth = 512, int $options = 0);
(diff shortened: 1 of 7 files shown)


sagikazarmark:
Thanks for the PR!

The automatic JSON encoding is there to make using Guzzle easier, but it shouldn't be used for advanced use cases. Encoding should be done by the user. Therefore I don't think this feature should land in Guzzle as it is.

gmponos:
I agree with @sagikazarmark you can easily do this:

```
$client->get('www.example.com', [
    'body' => $this->encoder->encode($yourData)
]);
```

fhink:
Thats why I made this PR. Currently it is being encoded automatically when using the JSON option.

GrahamCampbell:
Seems kinda odd to have special treatment of JSON but not others?

src/Encoder/JsonEncoderInterface.php
```diff
@@ -0,0 +1,23 @@
...
+namespace GuzzleHttp\Encoder;
+
+interface JsonEncoderInterface
+{
+    /**
+     * @param mixed $value
+     * @param int   $options
+     * @param int   $depth
+     * @return string
+     */
```
GrahamCampbell: Missing `@throws`.

src/Encoder/JsonEncoderInterface.php
```diff
@@ -0,0 +1,23 @@
...
+    public function encode($value, int $options = 0, int $depth = 512): string;
+
+    /**
+     * @param string $json
+     * @param bool   $assoc
+     * @param int    $depth
+     * @param int    $options
+     * @return mixed
+     */
```
GrahamCampbell: Same.

fhink:
Yes it is inconsistent. Perhaps a better fix would be to remove the json_encode for the 'JSON' key so it works the same as the 'body' key, like gmponos' example.

alexeyshockov:
I'm also against it, if you need a custom encoder, just use and provide the results to `body` option directly.

Also this topic was already discussed in #2343.

sagikazarmark:
Thanks @fhink for the contribution, but I don't see it fit in Guzzle at the moment.

PR #2495, opened 2019-12-23
Widen the Exception type to Throwable

gmponos:
I should have these changes included on this: #2493 but I thought about later.

Please review thoroughly the code.. there might be more places that need a change. 

Thank you

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -342,7 +342,7 @@ class Client implements ClientInterface, \Psr\Http\Client\ClientInterface
 
         try {
             return Promise\promise_for($handler($request, $options));
-        } catch (\Exception $e) {
+        } catch (\Throwable $e) {
             return Promise\rejection_for($e);
         }
     }
--- a/src/Client.php
+++ b/src/Client.php
@@ -342,7 +342,7 @@ class Client implements ClientInterface, \Psr\Http\Client\ClientInterface
 
         try {
             return Promise\promise_for($handler($request, $options));
-        } catch (\Throwable $e) {
+        } catch (\Exception $e) {
             return Promise\rejection_for($e);
         }
     }
--- a/src/Handler/CurlFactory.php
+++ b/src/Handler/CurlFactory.php
@@ -569,7 +569,7 @@ class CurlFactory implements CurlFactoryInterface
                 if ($onHeaders !== null) {
                     try {
                         $onHeaders($easy->response);
-                    } catch (\Throwable $e) {
+                    } catch (\Exception $e) {
                         // Associate the exception with the handle and trigger
                         // a curl header write error by returning 0.
                         $easy->onHeadersException = $e;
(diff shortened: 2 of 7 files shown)


GrahamCampbell:
Not necessarily. It might not make sense to catch throwables, actually. Throwables that are not exceptions are errors in the code, and probably should totally crash everything, on purpose.

gmponos:
>  It might not make sense to catch throwables, actually

Yes for some of the cases above I have second thoughts also.. can you be more specific.. because I don't think that this applies everywhere.. 

For instance on my current PR I believe that MockHandler should accept `Throwables`.

sagikazarmark:
@GrahamCampbell 's comment refers to _catching_ `\Throwable`s, not _accepting_ them and I kinda agree, although in some cases those caught throwables are actually wrapped in another exception and thrown again.

gmponos:
> comment refers to catching \Throwables

Yes.. I had second thoughts too. So shall I revert every `} catch (Throwable) {` I did?

sagikazarmark:
No, I don't think you should.

For example: it's probably a good thing, that promises can now catch errors too, and return them as rejection.

But I'm not sure catching errors and returning them as `RequestException` is a good idea.

gmponos:
OK.. then can we have specific comments per case? :)

src/Client.php
```diff
@@ -342,7 +342,7 @@ private function transfer(RequestInterface $request, array $options): PromiseInt
 
         try {
             return Promise\promise_for($handler($request, $options));
-        } catch (\Exception $e) {
+        } catch (\Throwable $e) {
```
sagikazarmark: I would revert this for now.

src/Handler/CurlFactory.php
```diff
@@ -569,7 +569,7 @@ private function createHeaderFn(EasyHandle $easy): callable
                 if ($onHeaders !== null) {
                     try {
                         $onHeaders($easy->response);
-                    } catch (\Exception $e) {
+                    } catch (\Throwable $e) {
```
sagikazarmark: I would revert this for now.

src/Handler/MockHandler.php
```diff
@@ -79,7 +79,7 @@ public function __invoke(RequestInterface $request, array $options): PromiseInte
             }
             try {
                 $options['on_headers']($response);
-            } catch (\Exception $e) {
+            } catch (\Throwable $e) {
```
sagikazarmark: I would revert this for now.

src/Handler/StreamHandler.php
```diff
@@ -115,7 +115,7 @@ private function createResponse(
         if (isset($options['on_headers'])) {
             try {
                 $options['on_headers']($response);
-            } catch (\Exception $e) {
+            } catch (\Throwable $e) {
```
sagikazarmark: I would revert this for now.

PR #2496, opened 2019-12-23
Deprecate functions and create internal ones

gmponos:
Some of these functions seem to be for internal use. This is a replacement in sort of a way of https://github.com/guzzle/guzzle/pull/2439

### DIFF
--- a/src/functions.php
+++ b/src/functions.php
@@ -13,8 +13,24 @@ use Psr\Http\Message\UriInterface;
  *
  * @return string Returns a string containing the type of the variable and
  *                if a class is provided, the class name.
+ *
+ * @deprecated
  */
 function describe_type($input): string
+{
+    @trigger_error(sprintf('Function %s is deprecated and will be removed on guzzle 8.0', __FUNCTION__), E_USER_DEPRECATED);
+    return _describe_type($input);
+}
+
+/**
+ * Debug function used to describe the provided value type and class.
+ *
+ * @return string Returns a string containing the type of the variable and
+ *                if a class is provided, the class name.
+ *
+ * @internal
+ */
+function _describe_type($input): string
 {
     switch (\gettype($input)) {
         case 'object':
@@ -55,8 +71,25 @@ function headers_from_lines($lines): array
  * @param mixed $value Optional value
  *
  * @return resource
+ *
+ * @deprecated
  */
 function debug_resource($value = null)
+{
+    @trigger_error(sprintf('Function %s is deprecated and will be removed on guzzle 8.0', __FUNCTION__), E_USER_DEPRECATED);
+    return _debug_resource($value);
+}
+
+/**
+ * Returns a debug stream based on the provided variable.
+ *
+ * @param mixed $value Optional value
+ *
+ * @return resource
+ *
+ * @internal
+ */
+function _debug_resource($value = null)
 {
     if (\is_resource($value)) {
         return $value;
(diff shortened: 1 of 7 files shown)


src/functions.php
```diff
@@ -75,8 +108,27 @@ function debug_resource($value = null)
  * @throws \RuntimeException if no viable Handler is available.
  *
  * @return callable Returns the best handler for the given system.
+ *
+ * @deprecated
  */
 function choose_handler(): callable
+{
+    @trigger_error(sprintf('Function %s is deprecated and will be removed on guzzle 8.0', __FUNCTION__), E_USER_DEPRECATED);
```
alexeyshockov: I'm not sure about this one, because:
1. it's already in use in [_public_ documentation](https://github.com/guzzle/guzzle/blob/e889719ced2eefd498844550d69a99e681ea4c43/docs/handlers-and-middleware.rst)
2. I definitely see a use case described above, when you want to create an empty `HandlerStack` with the default handler, IMO it's legit.

src/functions.php
```diff
@@ -101,8 +153,21 @@ function choose_handler(): callable
 
 /**
  * Get the default User-Agent string to use with Guzzle
+ *
+ * @deprecated
  */
 function default_user_agent(): string
+{
+    @trigger_error(sprintf('Function %s is deprecated and will be removed on guzzle 8.0', __FUNCTION__), E_USER_DEPRECATED);
```
alexeyshockov: Not sure about this. Though it's kind of internal, but others can only read from it anyway, not modify. If we make the function private, there will be no way to just read the default agent.

I try to think about it as a constant, in this case IMO it would be a public constant like
```php
class Client
{
    public const DEFAULT_USER_AGENT = 'GuzzleHttp';
}
```

So for me it's a constant that is calculated at the first access (based on the environment). Thinking this way, I'm not sure why we should hide/protect it.

sagikazarmark:
1. Why do we have to rename the functions? Marking them internal with a different name is the same as just marking them internal in the next major release
2. Why bother at all? I don't see any value in messing around with these functions for the sake of making them internal.

gmponos:
> Why bother at all?

Not bound by BC promise.

> Why do we have to rename the functions? Marking them internal with a different name is the same as just marking them internal in the next major release

- I was thinking about applying this on 7.1 after the 7.0 is released..
- I would like to have a pattern of having internal functions with an underscore prefix. I wouldn't like to rename them moving to 7.0 will cause overhead.. 

If I had to choose between doing the changes on 7.0 or leave them be.. I would leave them...

sagikazarmark:
> Not bound by BC promise.

I don't see a major risk in having to change these functions. This is not an interface, not an abstraction, these are functions. If we need to change them, we can just introduce new functions.

There is no harm in that, it won't make the API obscure.

sagikazarmark:
I don't think this is necessary. We can get back to it later if we need to do it.

GrahamCampbell:
This PR should be closed now, or updated to mark the methods on the Utils class as internal. The old functions will already be marked as deprecated in 7.1, but their deprecation message will need to say that there is no replacement, for the ones that are being made internal.

GrahamCampbell:
TBH, `describeType` and `debugResource` are the only two here, that I think we should change to be internal.

GrahamCampbell:
We maybe need to wait until 8.0 until formally adding the internal annotation though, since surely we should treat adding the internal annotation to something that used to be public, as a BC break?

Nyholm:

gmponos:
Closing I don't think it's worth fixing conflicts.. we can rediscuss this..

PR #2497, opened 2019-12-23
Method annotation is not enforcing return type argument

gmponos:
Since the class is not final and can be extended the method annotation does not enforce the return type.


tests/Exception/BadResponseExceptionTest.php
```diff
@@ -0,0 +1,24 @@
...
+        $prev = new \Exception();
+        $response = new Response();
+        $e = new BadResponseException('foo', $req, $response, $prev);
+        self::assertSame($req, $e->getRequest());
+        self::assertSame($response, $e->getResponse());
+        self::assertTrue($e->hasResponse());
+        self::assertSame('foo', $e->getMessage());
+        self::assertSame($prev, $e->getPrevious());
+    }
+}
```
sagikazarmark: Missing newline

gmponos: fixed..

phpstan-baseline.neon
```diff
@@ -200,6 +200,11 @@ parameters:
 			count: 1
 			path: src/Exception/BadResponseException.php
 
+		-
+			message: "#^Method GuzzleHttp\\\\Exception\\\\BadResponseException\\:\\:getResponse\\(\\) should return Psr\\\\Http\\\\Message\\\\ResponseInterface but returns Psr\\\\Http\\\\Message\\\\ResponseInterface\\|null\\.$#"
+			count: 1
+			path: src/Exception/BadResponseException.php
+
```
gmponos: I added this to baseline

IMHO it is a false positive.

sagikazarmark: Might even be a bug

gmponos: No it is not... phpstan only looks at the docblock..  and does not check the whole flow of the code in order to understand that the response can never be null...

sagikazarmark: What if you add a docblock to the function then?

PR #2499, opened 2019-12-23
Adding types on MockHandler fixes various issues

gmponos:
This might conflict with #2495 but I will handle it..


gmponos:
@sagikazarmark and @Tobion commit ac28f4e failed  https://github.com/guzzle/guzzle/pull/2499/checks?check_run_id=361496418

The message is weird.. Should the `CurlMultiHandlerTest` use the `MockHandler`

PR #2500, opened 2019-12-23
[8.0] Remove the exponentialDelay function from RetryMiddleware

gmponos:
I don't think this is needed to be public.. I guess it is a legacy thing that stayed public because there was no anonymous functions, or we were not so much used to them back then...

### DIFF
--- a/tests/RetryMiddlewareTest.php
+++ b/tests/RetryMiddlewareTest.php
@@ -1,4 +1,5 @@
 <?php
+
 namespace GuzzleHttp\Tests;
 
 use GuzzleHttp\Client;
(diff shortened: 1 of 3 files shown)


tests/RetryMiddlewareTest.php
```diff
@@ -71,10 +72,36 @@ public function testCanRetryExceptions()
...
+        self::assertCount(3, $calls);
+        self::assertSame(2000, $h->getLastOptions()['delay']);
+
+        $calls = [];
+        $h->append(new \Exception(), new \Exception(), new \Exception(), new Response());
+        $p = $c->send(new Request('GET', 'http://test.com'), []);
+
+        self::assertSame(200, $p->getStatusCode());
+        self::assertCount(4, $calls);
+        self::assertSame(4000, $h->getLastOptions()['delay']);
```
gmponos: My main problem here is that these tests are causing actual delays.. any workaround would be helpful here.

sagikazarmark: You could override the sleep function, or create an injectable/overrideable clock that falls back to the system clock. But I think this is fine for now as it is.

Nyholm: Or.. we could just keep it public.

The tests were easier to read and write and did not take forever to run =)

sagikazarmark: Or...we could not test it. 😄 It's a trivial sleep after all...

Nyholm: There has recently been a bug in this trivial code. I think the tests should be there.

gmponos: Let's not have this delay v7. Will add it to v8 milestone...

GrahamCampbell:
@gmponos Are you able to resolve the merge conflicts please?

PR #2508, opened 2019-12-28
Cleanup upgrade docs

Nyholm:
I've looked at the [diff](https://github.com/guzzle/guzzle/compare/6.5.0..master) and I found this document to be up to date. 

The only thing we can argue that is missing is the changes introduced in https://github.com/guzzle/guzzle/pull/2495. But I am happy not including them.

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -3,36 +3,44 @@ Guzzle Upgrade Guide
 
 6.0 to 7.0
 ----------
-In order Guzzle to take advantage of the new features of PHP, we dropped the support of PHP 5.
-The minimum supported PHP version is now PHP 7.2.
 
-Wherever it was possible type hints and return types for functions and methods are applied. 
+In order to take advantage of the new features of PHP, Guzzle dropped the support
+of PHP 5. The minimum supported PHP version is now PHP 7.2. Type hints and return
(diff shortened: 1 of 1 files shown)


UPGRADING.md
```diff
@@ -12,27 +12,31 @@ Please make sure:
...
+```php
+// Before:
+`curl_version()`;
+
+// After:
+`\curl_version()`;
+```
 
-For the full diff you can check [here](https://github.com/guzzle/guzzle/compare/6.5.0..7.0.0)
+For the full diff you can check [here](https://github.com/guzzle/guzzle/compare/6.5.0..master)
```
Nyholm: This change should be reverted when we released 7.0.0. 

(Sorry @gmponos for suggesting to change this in your original PR)

gmponos: No problem.. should be changed to 6.5.2 also we need to add changes of 6.5.2 to master

gmponos: > also we need to add changes of 6.5.2 to master
https://github.com/guzzle/guzzle/pull/2509

Nyholm: They are already included.

Nyholm:
Thank you for merging

PR #2509, opened 2019-12-28
Changes from 6.5.2 added to master

gmponos:


src/functions.php
```diff
@@ -316,17 +316,15 @@ function _current_time()
...
 
 /**
- * @param int $options
- *
- * @return UriInterface
- *
  * @internal
  */
-function _idn_uri_convert(UriInterface $uri, $options = 0)
+function _idn_uri_convert(UriInterface $uri, int $options = 0): UriInterface
```
GrahamCampbell: This won't work on PHP 5.5 and 5.6.

GrahamCampbell: In fact, it will cause a parse error.

GrahamCampbell: Oh, I missread the target as being Guzzle 6! Never mind...

gmponos: Sorry the title might be confusing... but this PR is targeting master where we support only 7.. I am migrating latest changes of v6.5 to master..

Issue #2505, opened 2019-12-28
Tests warnings


GrahamCampbell:
Some things to look at:

```
 1) GuzzleHttp\Tests\MessageFormatterTest::testCreatesWithClfByDefault
readAttribute() is deprecated and will be removed in PHPUnit 9.
getObjectAttribute() is deprecated and will be removed in PHPUnit 9.

2) GuzzleHttp\Tests\MiddlewareTest::testLogsWithStringError
Using assertContains() with string haystacks is deprecated and will not be supported in PHPUnit 9. Refactor your test to use assertStringContainsString() or assertStringContainsStringIgnoringCase() instead.

3) GuzzleHttp\Tests\Handler\CurlMultiHandlerTest::testCanSetSelectTimeout
readAttribute() is deprecated and will be removed in PHPUnit 9.
getObjectAttribute() is deprecated and will be removed in PHPUnit 9.

4) GuzzleHttp\Tests\Handler\CurlMultiHandlerTest::testUsesTimeoutEnvironmentVariables
readAttribute() is deprecated and will be removed in PHPUnit 9.
getObjectAttribute() is deprecated and will be removed in PHPUnit 9.

5) GuzzleHttp\Test\Handler\CurlFactoryTest::testClosesIdleHandles
readAttribute() is deprecated and will be removed in PHPUnit 9.
getObjectAttribute() is deprecated and will be removed in PHPUnit 9.
```

GrahamCampbell:
Also, from the js server:

```
 (node:18577) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.
```

sagikazarmark:
The js server is not important. I upgraded node under the server, but it should be fine.

gmponos:
Closing as the PR is merged. :) 

Thank you @GrahamCampbell

GrahamCampbell:
After https://github.com/guzzle/guzzle/pull/2517 is merged too, yes.

PR #2512, opened 2019-12-29
Fixed deprecation warnings in tests

GrahamCampbell:
This will be necessary when we later update phpunit to "^8|^9", which will be necessary to test on PHP 8+.

### DIFF
--- a/tests/Helpers.php
+++ b/tests/Helpers.php
@@ -1,0 +1,34 @@
+<?php
+namespace GuzzleHttp\Tests;
+
+class Helpers
+{
+    public static function readObjectAttribute(object $object, string $attributeName)
+    {
+        $reflector = new \ReflectionObject($object);
+
+        do {
+            try {
+                $attribute = $reflector->getProperty($attributeName);
+
+                if (!$attribute || $attribute->isPublic()) {
+                    return $object->$attributeName;
+                }
+
+                $attribute->setAccessible(true);
+
+                try {
+                    return $attribute->getValue($object);
+                } finally {
+                    $attribute->setAccessible(false);
+                }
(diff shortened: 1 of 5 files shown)


sagikazarmark:
Thanks @GrahamCampbell ! I assume it fixes #2505

gmponos:
Please read this https://github.com/sebastianbergmann/phpunit/issues/3339

Phpunit has deprecated these methods because this is the wrong way of testing stuf.

GrahamCampbell:
It doesn't mean we shouldn't do it, ever. I think our current usage is fine.

GrahamCampbell:
I intentionally didn't make it a trait because of the method name collision.

tests/Helpers.php
```diff
@@ -0,0 +1,34 @@
...
+                if (!$attribute || $attribute->isPublic()) {
+                    return $object->$attributeName;
+                }
+
+                $attribute->setAccessible(true);
+
+                try {
+                    return $attribute->getValue($object);
+                } finally {
+                    $attribute->setAccessible(false);
```
Nyholm: I dont see why this is needed.

gmponos:
Hmm good but i still prefer the name... helper seems like a pool that everyone can add anything...

tests/Helpers.php
GrahamCampbell: Just copied from the original implementation. Could remove it I suppose.

Nyholm: Okey. 
I thought it did not matter.
I'll see if I can find the reason why it was added.

GrahamCampbell: I think it's probably not needed too. Doesn't really matter though. Only test code, and it doesn't hurt.

PR #2513, opened 2019-12-29
Add get, post, ... to the client interface, removing the need for dynamic dispatch

GrahamCampbell:
I think this is pragmatic solution (replacing #2482) to allow people to still call the short hand methods they want to, but also make the interface easy to implement, by providing a trait that implements these extra methods by calling `request` and `requestAsync`.

---

Closes #2482.

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -11,11 +11,6 @@ parameters:
 			path: src/Client.php
 
 		-
-			message: "#^Method GuzzleHttp\\\\Client\\:\\:__call\\(\\) has parameter \\$args with no value type specified in iterable type array\\.$#"
-			count: 1
-			path: src/Client.php
-
-		-
 			message: "#^Method GuzzleHttp\\\\Client\\:\\:sendAsync\\(\\) has parameter \\$options with no value type specified in iterable type array\\.$#"
 			count: 1
 			path: src/Client.php
@@ -27,12 +22,12 @@ parameters:
 
 		-
 			message: "#^Method GuzzleHttp\\\\Client\\:\\:requestAsync\\(\\) has parameter \\$options with no value type specified in iterable type array\\.$#"
-			count: 1
+			count: 2
 			path: src/Client.php
 
 		-
 			message: "#^Method GuzzleHttp\\\\Client\\:\\:request\\(\\) has parameter \\$options with no value type specified in iterable type array\\.$#"
-			count: 1
+			count: 2
 			path: src/Client.php
 
 		-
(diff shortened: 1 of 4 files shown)


sagikazarmark:
I like it, thanks @GrahamCampbell . I would probably postpone this to 7.1, because it's technically not a BC break.

Or is it? Looking at the code, methods like `->purge` also seem to work.

Why are there so much phpstan baseline rule changes?

alexeyshockov:
+1 for this

I think it’s a BC break after all, because Guzzle could be used for WebDAV, for example, which uses additional HTTP verbs. So it should be in `7.0`.

gmponos:
> Or is it? Looking at the code, methods like ->purge also seem to work.

People that have a custom implementation of  `ClientInterface` will break for them.. The functions are also added to the interface.. Same goes for the PHPStan question.

> People that have a custom implementation of  `ClientInterface` will break for them.. 

If we had BC roave it would be reported also as a BC

alexeyshockov:
So, if we agree that it's a BC break, then it would be better:
1. not remove `__call()` completely in this PR, but only deprecate it
2. merge this PR after `7.0`, to make it a part of `7.1` (as @sagikazarmark suggested)
3. prepare a different PR to remove `__call()` completely, for `8.0`

@GrahamCampbell, what do you think?

GrahamCampbell:
The 7.1.0 milestone is incorrect, since this is a major breaking change to the interface. Needs to be 7.0.0 if we are considering this for the 7.x series.

sagikazarmark:
Let's keep the discussion here (we talked about this on Slack as well).

Currently we are discussing whether these methods should be added to the interface or not. There are pro/contra opinions.

**If** we decide not to add these methods to the interface, then it's not a breaking change.

Personally I see no harm in adding these methods to the interface.

On the other hand, we could keep the magic method and continue supporting HTTP verbs, like `purge` and `list`. From that perspective, removing the magic method is also a BC break.

As I said, I don't have a strong opinion whether we should add the http verb methods to the interface. If you could list a few pro/contra arguments I could probably pick a side.

Nyholm:
I believe the 7.1 milestone is correct. 

After 7.0 is released, I think we should deprecate `__call()` and add these methods (implemented with a trait). 
That will make it easy to upgrade, and we will keep the changelog to 7.0 small. 

**About adding these method in an interface:**
That is a separate discussion, they have never been in an interface, nobody expects them to be in the interface. These methods are easy and convenient for application authors. However, if you are writing a library I dont mind forcing you to write `request('GET', ...)`.

GrahamCampbell:
> nobody expects them to be in the interface.

Not true. There is loads of code out there that typehints ClientInterface, but then calls ->get, because developers do what looks nice...

Nyholm:
> >nobody expects them to be in the interface.

> Not true. There is loads of code out there that typehints ClientInterface, but then calls ->get, because developers do what looks nice...

Okey, you might be right. 

But "looks nice" and "being correct" is not the same thing. I dont think we need syntactic sugar in our `ClientInterface`

GrahamCampbell:
A strongly feel we should add these methods to the client interface. That is the only actual objective of this PR. It makes no practical difference otherwise, if we use __call or not in the client.

Reasons for adding:

1. People already call the methods as if they exist on the interface!
2. Guzzle should be easy to use and appealing to new commers.
3. Makes tests mocking the interface easy.

Reasons for not adding:

1. To encourage PSR-style usage.
2. Makes the interface harder to implement.

The reasons against are weak for the following reasons:

1. People will do what they want anyway... this is why people already call ->get on things that implement ClientInterface. It's fine to let people know that PSR is around, but not to force them to use it.
2. People implementing the interface can pull in the trait I added, so it's no harder to use. The other way people use the interface is test mocking. This is no harder, in fact easier now.

GrahamCampbell:
If this does go into 7.0.0, I wonder if we should also make the Client class final?

alexeyshockov:
Not related directly, but: does `ClientInterface` makes sense at all? There is PSR-18 for that already, that defines the client interface. If we deprecate/delete `->getConfig()` method (another discussion), it will even less different from the PSR's one. IMO it makes sense to deprecate the whole interface, because it doesn't bring any practical value anymore.

gmponos:
> If this does go into 7.0.0, I wonder if we should also make the Client class final?

we will add final annotations on 7.1

GrahamCampbell:
Adding a final annotation is a major breaking change.

GrahamCampbell:
Just to be clear. In a minor release we absolutely cannot ever:

1. Add a new method to an interface.
2. Change the signature of a method in an interface.
3. Add the final modifier to a class.
4. Delete a public method of a class.
5. Delete a protected method of a non-final class.

GrahamCampbell:
> Delete a protected method of a non-final class.

Unless we made it clear the method/class was "internal".

gmponos:
> Adding a final annotation is a major breaking change.

How come.? Annotations is not code.. people can still extend the classes?

> if we use __call or not in the client.

It does.. since the client class is not final adding the methods enforces typehints.. Method annotations do not..  as I said annotation comments are not code..

GrahamCampbell:
> How come.? Annotations is not code.. people can still extend the classes?

I was talking about code. I'd be massively against just documenting final without enforcing it.

GrahamCampbell:
> It does.. since the client class is not final adding the methods enforces typehints.. Method 

Well, yeh. If the class is not final, it makes no difference. It certainly makes no difference from the perspective of anyonen using the code compositionally.

gmponos:
About the final created this.. let's not diverge further the discussion.

https://github.com/guzzle/guzzle/issues/2521

Please note about adding the final modifier or removing the interface.. that we have already stopped adding new features to version 7 as we want the developers to have an easy update to version 7.

We have already introduced a lot of changes with PHP 7 and typehints. We had more features in our mind that we could implement on version 7 but we postponed them for later. 

ofc we can discuss new things that someone might have a different PoV and believes that it fits 7.0.

Back on the discussion about the interface.. I am with @Nyholm on this..

GrahamCampbell:
Adding these methods to the interface will have no bad consequences surely. Only good ones? We don't expect there to be many custom implementations of the interface, and the upgrading path is super easy if there are: just load the trait. There are only advantages:

1. Mocking the interface in tests works now.
2. Code that typehints the interface can call the nice methods.

GrahamCampbell:
If we don't add these methods, why do we even bother having an interface....

sagikazarmark:
It seems that we are not reaching a consensus, not to mention that we are actually discussing more than one thing:

- making classes final
- dropping the interface

The goal for Guzzle 7 was as easy migration from 6 as possible. That's why we tried to minimize breaking changes: apart from PHP 7 hints and removed deprecations I hardly think we broke anything.

Therefore I propose we delay this to Guzzle 8. The current solution worked for years, so I don't see the gain, compared to the pain this can cause.

GrahamCampbell:
Ok, then... should we merged this **without the interface additions** into Guzzle 7.0.0? Do we have a consensus on that, at least?

Nyholm:
Sorry no. 

One major goal of Guzzle 7 is to minimize the number of changes. It is weird, I know. But last time Guzzle released a major version half the PHP community got upset. So that is why changes are kept to a minimum. 

However, Guzzle 7 comes with the PSR-18 interface. So in the ideal world, most libraries are using PSR-17 and PSR-18. That is why I don't mind doing lots of breaking changes in Guzzle 8. I also dont mind releasing Guzzle 8 within 12 months. 

This will give users an easy upgrade path between Guzzle 6 and 7 AND it will allow us to make BC changes fairly soon. 

-------

Could you remove all BC breaks in this PR and add a deprecation warning on `__call()`, then I will be happy to merge.

GrahamCampbell:
Ok, so how about this for a plan:

1. We add the extra methods to the client BEFORE the 7.0.0 release, but leave __call untouched.
2. We deprecate __call in 7.1.0, but leave its behaviour untouched.
3. We remove __call in 8.0.0.
4. No changes to the ClientInterface. To be discussed later.

GrahamCampbell:
https://github.com/guzzle/guzzle/pull/2482 should be re-opened, for merge into 7.1.0, I think?

PR #2515, opened 2019-12-29
[7.0] Avoid the getConfig function

GrahamCampbell:
A precursor to https://github.com/guzzle/guzzle/pull/2516.

### DIFF
--- a/tests/ClientTest.php
+++ b/tests/ClientTest.php
@@ -64,13 +64,14 @@ class ClientTest extends TestCase
             'headers'  => ['bar' => 'baz'],
             'handler'  => new MockHandler()
         ]);
-        $base = $client->getConfig('base_uri');
-        self::assertSame('http://foo.com', (string) $base);
-        self::assertInstanceOf(Uri::class, $base);
-        self::assertNotNull($client->getConfig('handler'));
-        self::assertSame(2, $client->getConfig('timeout'));
-        self::assertArrayHasKey('timeout', $client->getConfig());
-        self::assertArrayHasKey('headers', $client->getConfig());
+        $config = Helpers::readObjectAttribute($client, 'config');
+        self::assertArrayHasKey('base_uri', $config);
+        self::assertInstanceOf(Uri::class, $config['base_uri']);
+        self::assertSame('http://foo.com', (string) $config['base_uri']);
+        self::assertArrayHasKey('handler', $config);
+        self::assertNotNull($config['handler']);
+        self::assertArrayHasKey('timeout', $config);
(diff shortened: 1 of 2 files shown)


tests/ClientTest.php
```diff
@@ -64,13 +64,14 @@ public function testClientHasOptions()
...
             'handler'  => new MockHandler()
         ]);
-        $base = $client->getConfig('base_uri');
-        self::assertSame('http://foo.com', (string) $base);
-        self::assertInstanceOf(Uri::class, $base);
-        self::assertNotNull($client->getConfig('handler'));
-        self::assertSame(2, $client->getConfig('timeout'));
-        self::assertArrayHasKey('timeout', $client->getConfig());
-        self::assertArrayHasKey('headers', $client->getConfig());
+        $config = Helpers::readObjectAttribute($client, 'config');
```
gmponos: That's what I feared with having this https://github.com/guzzle/guzzle/pull/2512 merged...

This was wrong before.. it is wrong now.. and now it is even worse because in kind of a way we promote this.

Test should not care what the property of the config is.. Tests should care that if I create a client with a `base_uri` that then a request is sent having the base uri...

GrahamCampbell: Well, I'm making it no worse than it already is. This PR is necessary if we don't want getConfig to be part of the public interface of the client, which I think is a good move.

We can look to refactor the tests later, and maybe even remove the helper class too. Maybe we should create an issue to address that?

gmponos: > This was wrong before.. it is wrong now..

btw I meant that the test was wrong.. and no you are not making it worse.. I believe that the current test should be replaced with separate ones.. 

for instance a test that checks that the base_url is applied correctly when a request is made.. etc etc.. and this brings me to my next question.. 

> We can look to refactor the tests later

Why should we do it later.. Can't we do it now?

> Maybe we should create an issue to address that

I don't think it's worth an issue.. Yes, reflection is needed in Curl tests because of the complexity of testing native functions of curl etc etc.. but I don't think ti should be used lightly everywhere else...

GrahamCampbell: > Why should we do it later.. Can't we do it now?

Because I don't want to implement it now (because it will take some time for no gain in src), but I do want to deprecate the getConfig function. Someone else is welcome to implement a more full fix to avoid the helper if they want, but I feel that is probably best done as another PR, after this one.

sagikazarmark: This PR is not about refactoring tests. While it might not be perfect, it serves its purpose.

Nyholm:
Thank you

PR #2516, opened 2019-12-29
[7.1] Deprecate the getConfig method

### LINKED ISSUE #2514: [RFC] Deprecate getConfig
I feel like the client configuration should be immutable, and should not be exposed to the outside, either to be read, or to be changed. Should we mark `getConfig` on the client interface as deprecated in 7.x, and remove it in 8.0.0?

GrahamCampbell:
This depends on https://github.com/guzzle/guzzle/pull/2515 being merged first.

---

Closes #2514.

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -189,6 +189,8 @@ class Client implements ClientInterface, \Psr\Http\Client\ClientInterface
      * @param string|null $option The config option to retrieve.
      *
      * @return mixed
+     *
+     * @deprecated Client::getConfig will be removed in guzzlehttp/guzzle:8.0.
      */
     public function getConfig(?string $option = null)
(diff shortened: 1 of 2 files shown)


gmponos:
I suggest to move this discussion for 7.1

sagikazarmark:
Like the idea, but I agree we should move this to 7.1 with the rest of the deprecations.

GrahamCampbell:
Sure. It would be great if #2515 could be merged earlier, so that the docs don't show using a function we plan to deprecate, and neither does our test code. That's why I separated it into two PRs. ;)

Nyholm:
I agree to deprecate this function.

Nyholm:

PR #2517, opened 2019-12-29
[6.5] Fixed deprecation warning in node test server

GrahamCampbell:
Will allow us to run on node 12 on the master branch when merged upwards.

### DIFF
--- a/tests/server.js
+++ b/tests/server.js
@@ -178,7 +178,7 @@ var GuzzleServer = function(port, log) {
         that.responses = JSON.parse(request.body);
         for (var i = 0; i < that.responses.length; i++) {
           if (that.responses[i].body) {
-            that.responses[i].body = new Buffer(that.responses[i].body, 'base64');
+            that.responses[i].body = new Buffer.from(that.responses[i].body, 'base64');
           }
         }
         if (that.log) {


gmponos:
> master branch when merged upwards.

Hm.. I am not sure about the process that we will follow.. maybe two different PRs might be needed.

GrahamCampbell:
6.5 is only 5 commits behind master. Should be possible to do a `6.5 -> master` merge. I could probably do it as a PR if you want?

GrahamCampbell:
https://github.com/guzzle/guzzle/pull/2519

tests/server.js
```diff
@@ -178,7 +178,7 @@ var GuzzleServer = function(port, log) {
         that.responses = JSON.parse(request.body);
         for (var i = 0; i < that.responses.length; i++) {
           if (that.responses[i].body) {
-            that.responses[i].body = new Buffer(that.responses[i].body, 'base64');
+            that.responses[i].body = new Buffer.from(that.responses[i].body, 'base64');
```
guilliamxavier, line 181: you should use either `new Buffer(` or `Buffer.from(`, not `new Buffer.from(` (even if it seems to work...)

guilliamxavier: #2638

PR #2519, opened 2019-12-29
Merge 6.5->master

GrahamCampbell:
I just did a `git merge 6.5` on the `master` branch, and sorted out the conflicts. There were actually no changes not already present on master, other than my most recent fix to the js server.

### DIFF
--- a/tests/server.js
+++ b/tests/server.js
@@ -178,7 +178,7 @@ var GuzzleServer = function(port, log) {
         that.responses = JSON.parse(request.body);
         for (var i = 0; i < that.responses.length; i++) {
           if (that.responses[i].body) {
-            that.responses[i].body = new Buffer(that.responses[i].body, 'base64');
+            that.responses[i].body = new Buffer.from(that.responses[i].body, 'base64');
           }
         }
         if (that.log) {


gmponos:
Hard to follow the linkage 😛 we could keep the discussion on previous...

> There were actually no changes not already present on master

> https://github.com/guzzle/guzzle/pull/2509

Master and 6.5 have diverged and I think the decision is to be diverged... we could merge this.. but I think this is the direction...

GrahamCampbell:
> Master and 6.5 have diverged

Barely. That's why it was so easy for me to do the branch merge just now. :)

GrahamCampbell:
By which I mean, master has gone forward a lot, but 6.5 has not gone in a different direction.

gmponos:
>  has not gone in a different direction.

because we merged #2509.. the of 6.5 exists on master but the history is different.. because of #2509

Anyway that's not the important thing here..

The important thing is that maybe we could have the discussion here about what the direction will be.. having a PR from 6.5 to master each time a new thing reaches 6.5 is a little bit noise... Having developers submitting duplicate PR is the same noise.

Also we need to see how we will backport bugs.. from master to 6.5 (check https://github.com/guzzle/guzzle/pull/2311#issuecomment-569420116)

IIRC another thing that "some" ppl might care is that contribution history is counted only on master :P
https://help.github.com/en/github/setting-up-and-managing-your-github-profile/why-are-my-contributions-not-showing-up-on-my-profile

GrahamCampbell:
Better to merge bug fixes into the lowest branch, then merge upwards. This is what Symfony and Laravel do.

> IIRC another thing that "some" ppl might care is that contribution history is counted only on master :P

Doesn't matter if PRs are not merged into master, as long as the branch they are merged into is then merged up into master, for commits showing in the contribution graph for a repo.

sagikazarmark:
The problem with merging 6.5 into master is that 6.5 contains and will always contain PHP 5.x specific fixes, whereas we probably don't need those on master or need a PHP 7 specific version of them.

There is a risk that manual conflict resolution might not always catch these things (for example: there were some PHP 5.x specific fixes related to IDN/ICU).

This PR seems okay, but generally I'm against merging 6.5 into master.

PR #2520, opened 2019-12-30
Add lowest build to matrix

sagikazarmark:


sagikazarmark:
> You cannot use include to add new jobs to a build matrix.

PR #2523, opened 2019-12-30
Remove additional backticks in upgrading document

BackEndTea:
It was already inside of a code block, implying that it would be a shell call 
from php, while it is just a function call.


BackEndTea:
That was quick, thanks @sagikazarmark

PR #2524, opened 2019-12-30
Update InvalidArgumentException message for request with array body

### LINKED ISSUE #1594: GET request with array body returns an error
Version : Composer package v. 6.1

I'm not very familiar with Guzzle yet but there seems to be a smal inconsistency within the Guzzle client. I'm making the following query:

``` php
$client = new GuzzleHttp\Client( .... );
$response = $client->request(
    'GET',
    'http://httpbin.org',
    [
        RequestOptions::BODY => [
            'hello' => 'world'
        ]
    ]
);
```

This code throws the following exception: 

> Passing in the "body" request option as an array to send a POST request has been deprecated. Please use the "form_params" request option to send a application/x-www-form-urlencoded request, or a the "multipart" request option to send a multipart/form-data request.

Now this is clearly not a POST request so either the exception message is wrong or an additional check is required to check for the client method being used before throwing the exception. I would make a PR but I'm not sure which one's right.

Thank you for your great work! :+1:

gmponos:
Closes #1594 

Two things.. 

1. GET requests can have a body (sounds weird).
2. This PR returns the exception instead of throwing.. this helps various tools do better analysis when the code stops.

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -163,7 +163,7 @@ class Client implements ClientInterface, \Psr\Http\Client\ClientInterface
         // Merge the URI into the base URI.
         $uri = $this->buildUri(Psr7\uri_for($uri), $options);
         if (\is_array($body)) {
-            $this->invalidBody();
+            throw $this->invalidBody();
         }
         $request = new Psr7\Request($method, $uri, $headers, $body, $version);
         // Remove the option so that they are not doubly-applied.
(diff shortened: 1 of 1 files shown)


Nyholm:
I dont see how this PR solves (1.)

Could you elaborate?

gmponos:
The message [here](https://github.com/guzzle/guzzle/pull/2524/files#diff-ff73e042e738204c6da009e2ed19f783R474) says about `POST` although `form_params` can be used with `GET` also.

Nyholm:
Oh, I see. you just updated the message. I was expecting some code that allowed GET requests to have a body.
 =)

Issue #2186, opened 2018-10-31
Future of guzzle with PSR-17 and PSR-18


gmponos:
Hello,

I just saw the [tweet](https://twitter.com/TobiasNyholm/status/1057562838217179136) of @Nyholm about the first tag of psr/http-client. I am super excited about this. Great job you guys.

I am starting this issue in order to share some thoughts about guzzle and PSRs

1. First of all will there be a new package of guzzle implementing PSR-17?
2. Will you continue supporting options? I know that PSR-18 does not allow options. So what I am asking is if guzzle will continue supporting or if you would consider supporting something like this:

```
$request = $factory->createRequest('get', 'http://www.test.com');
$request = GuzzleHttp\function\apply_options($request, $options);
$client->send($request);
```

if not... what will be the main advantage of someone using guzzle instead of httplug?

3. Would you consider starting a new branch accepting contributions? I am willing to help here.
4. Also what will happen with async functions? Will you continue supporting them without options?

sagikazarmark:
> First of all will there be a new package of guzzle implementing PSR-17?

Implementing PSR-17 is already in progress and the code will be part of the PSR-7 package. The naming is a little bit unfortunate, but it provides the most flexibility in our opinion.

> Will you continue supporting options?

Guzzle will eventually implement PSR-18 directly. We will try to limit the number of breaks to as small as possible. Also, we will keep Guzzle's rich API.

> if not... what will be the main advantage of someone using guzzle instead of httplug?

HTTPlug consists of a number of different components:

1. HTTP Client contract which will mainly be replaced by PSR-18
2. Adapter packages which provide a thin layer between clients limited to the standard (compared to Guzzle's rich API)
3. A number of HTTP related components relying on HTTPlug/PSR-18

With the standard out adapter packages will probably become obsolete eventually, once clients themselves implement the standard, but the rest of the ecosystem will remain (plugins, message tools, etc).

> Would you consider starting a new branch accepting contributions? I am willing to help here.

We are still evaluating options and concentrate on PSR-17 first, but we will certainly accept contributions. I'll let you know once we start working on PSR-18/PHP 7.x implementation.

> Also what will happen with async functions? Will you continue supporting them without options?

As mentioned earlier, Guzzle's API will remain as it is for most of it's part and we will keep all the features, so I don't expect major changes here.

gmponos:
Let me see if I understood correctly.

[...]

PR #2525, opened 2019-12-30
PoC for PSR-17

gmponos:
Current PR is only meant for discussion. It's purpose is not to be merged.

The main goals of this are the following:

- We want to support PSR-17
- We want developers to use the rich and useful RequestOptions of guzzle in order to create a request and send it use PSR-18 method.

How:

The biggest change here is that we extract `applyOptions` function into a separate function. By extracting the applyOptions to a separate function developers are able to create classes like in the following example:

```
class MyHandler
{

    /**
     * @var ClientInterface
     */
    private $client;

    /**
     * @var RequestFactoryInterface
     */
    private $factory;

    public function __construct(Psr\Http\Client\ClientInterface $client, RequestFactoryInterface $factory)
    {
        $this->client = $client;
        $this->factory = $factory;
    }

    public function sendEmail(string $email)
    {
        $request = $this->factory->createRequest('POST', 'www.sendemail.com');
        $options = [RequestOptions::QUERY => ['email' => $email]];
        $request = apply_http_options($request, $options);
        $this->client->sendRequest($request);
    }
}
```

in order to support this we need to merge #1890 

After merging this it will allow us to remove these three lines of codes: https://github.com/guzzle/guzzle/pull/2525/files#diff-ff73e042e738204c6da009e2ed19f783L161

These options are re-applied twice and they might have a performance impact since we are initializing two Streams, when user passes body as an option.

Related mentions before:

https://github.com/guzzle/guzzle/pull/1914
https://github.com/guzzle/guzzle/issues/1886

> The bad thing I'm experiencing with my team is the more verbose way to create an HTTP request. They seem to find it complicated. 

https://github.com/guzzle/guzzle/issues/2186#issuecomment-522053117


gmponos:
I have updated the description. If possible I would like some comments :)

sagikazarmark:
I find it weird that we encourage people to use standards like PSR-17, but we also suggest using our `RequestOptions`.

What's the point in using PSR-17 then? People will hardly use an entirely different PSR-7/PSR-17 package with our options.

gmponos:
> People will hardly use an entirely different PSR-7/PSR-17 package with our options.

I will revert this.. Most probably ppl will not use PSR-7/PSR-17/PSR-18 if they have to build a request in a verbose way when they can simply use `$client->get('uri', ['query' => ['key' => 'value']]);`

Extracting `apply_options` makes the pain less.

GrahamCampbell:
> so I ended up creating a wrapper https://gist.github.com/gnumoksha/b35d97613dd729315d3dc1ba0449544a

Nice. :P That code is an example of where the new ClientTrait would be well-used. ;)

GrahamCampbell:
https://github.com/guzzle/guzzle/blob/master/src/ClientTrait.php

GrahamCampbell:
#3135 has now been merged, replacing #1890.

gmponos:
Will try to re-initiate this.

Thanks @GrahamCampbell .. Please give me an ETA of 5 days.. if you don't hear from me please proceed with any other implementation

GrahamCampbell:
OK, in order to enable this to happen in Guzzle 8, some additional pre-work is required:

* https://github.com/guzzle/psr7/pull/655
* https://github.com/guzzle/psr7/pull/657
* https://github.com/guzzle/psr7/pull/658
* https://github.com/guzzle/guzzle/pull/3388

I've applied this to PSR7 v2 and Guzzle v7, so people can benefit from some of these improvements sooner and notice any issues sooner.

GrahamCampbell:
And finally, the PR to implement this in Guzzle 8: https://github.com/guzzle/guzzle/pull/3433.

PR #2527, opened 2019-12-31
Add object typehint to json_decode

gmponos:


src/functions.php
```diff
@@ -261,7 +261,7 @@ function is_host_in_noproxy(string $host, array $noProxyArray): bool
  * @param int    $depth   User specified recursion depth.
  * @param int    $options Bitmask of JSON decode options.
  *
- * @return array|string|int|float|bool|null
+ * @return object|array|string|int|float|bool|null
```
alexeyshockov: BTW, why not `mixed`?

GrahamCampbell: Because `resource` cannot be returned.

PR #2529, opened 2019-12-31
[7.0] Added real client methods

GrahamCampbell:
So, maybe we have consensus for this plan now:

1. We merge this PR before the 7.0.0 release.
2. We merge https://github.com/guzzle/guzzle/pull/2482 for the 7.1.0 release.
3. We remove __call in 8.0.0.

---

It remains to be discussed what to do with the ClientInterface.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Changed
 
 * Avoid the getenv function when unsafe [#2531](https://github.com/guzzle/guzzle/pull/2531)
+* Added real client methods [#2529](https://github.com/guzzle/guzzle/pull/2529)
 
 ### Removed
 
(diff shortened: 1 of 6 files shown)


sagikazarmark:
Adding these methods isn't any kind of break, so we are free to do it in 7.0

GrahamCampbell:
> Adding these methods isn't any kind of break, so we are free to do it in 7.0

Adding these is a break, if people are extending the class.

We have already made a BC break since the 7.0.0 beta1, so more breaks are fine. I assume we will stop making breaks once we get to either RC or stable?

GrahamCampbell:
I've added a note to the upgrading guide.

sagikazarmark:
> Adding these is a break, if people are extending the class.

Well, only if someone implemented these methods with a different signature, but I'm not so concerned about that. Also, extending the `Client` class is not something that people should do (even if it's not marked as final).

But yeah, strictly speaking it CAN be a break in certain cases, but it's not a break in behavior.

PR #2530, opened 2019-12-31
[6.5] Avoid the unsafe getenv function

GrahamCampbell:
#### TLDR: `getenv` is not thread safe!

The `genv` and `putenv` functions are not threadsafe, so environment variables set in one thread using `putenv` may leak into another via `getenv`. This not just a "theoretical" bug. It actually happens in real apps and servers, and was the result of hundreds of bug reports on the Laravel framework over the years, where people were confused by variables not being what they expected!

---

EDIT: In order to minimize breakage, for CLI apps, I've allowed to still read from `getenv` for that.

### DIFF
--- a/src/functions.php
+++ b/src/functions.php
@@ -406,7 +406,7 @@ function _getenv($name)
         return (string) $_SERVER[$name];
     }
 
-    if (PHP_SAPI === 'cli' && ($value = getenv($name) !== false)) {
+    if (PHP_SAPI === 'cli' && ($value = getenv($name)) !== false) {
         return (string) $value;
     }
 
--- a/src/functions.php
+++ b/src/functions.php
@@ -392,3 +392,23 @@ function _idn_uri_convert(UriInterface $uri, $options = 0)
 
     return $uri;
 }
+
+/**
+ * @param string $name
+ *
+ * @return string|null
+ *
+ * @internal
+ */
+function _getenv($name)
+{
+    if (isset($_SERVER[$name])) {
+        return (string) $_SERVER[$name];
+    }
+
+    if (PHP_SAPI === 'cli' && ($value = getenv($name) !== false)) {
+        return (string) $value;
+    }
+
+    return null;
+}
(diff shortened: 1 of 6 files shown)


gmponos:
I have approved the PR.. never the less just for knowledge sharing.. I understand the use case for a framework that developers set sensitive info on env etc etc...

But for guzzle what is the risk.. ?

gmponos:
Also isn't this a BC? Should a developer before he used `putenv` to set these parameters to set them differently now?

I remember I had a bug when I changed this before.. but maybe I don't remember correctly

GrahamCampbell:
I think I need to email you privately about this one...

gmponos:
> I think I need to email you privately about this one...

You can find me on symfony slack and php-http slack as well...

src/functions.php
```diff
@@ -392,3 +392,23 @@ function _idn_uri_convert(UriInterface $uri, $options = 0)
...
+ *
+ * @internal
+ */
+function _getenv($name)
+{
+    if (isset($_SERVER[$name])) {
+        return (string) $_SERVER[$name];
+    }
+
+    if (PHP_SAPI === 'cli' && ($value = getenv($name) !== false)) {
```
gmponos: I like the concept but please help me understand this.. if someone has been using putenv (let's say on a boostrap file of his own) wont this break for him?

gmponos: https://3v4l.org/l4h7T

GrahamCampbell: Yes. That is intentional. Because calling `putenv` is what is not safe to do. This is the break we have to make in order to fix things. People should be loading stuff into the server super global, and not using `putenv` if they are bootstrapping something to run in a multithreaded environment, such as php-fpm.

GrahamCampbell:
It looks like Symfony never used to call `getenv` because of the thread safety issues. That PR you reference is so that they don't call `putenv`.

GrahamCampbell:
At the very least, in 6.x, we should at least look at `$_SERVER` first, and if set, return straight away, which mitigates must of the risk.

PR #2537, opened 2020-01-06
Avoid PHP bug #77946

phansys:
PHP bug [#77946](https://bugs.php.net/bug.php?id=77946) is fixed since [PHP 7.2.22](https://github.com/php/php-src/commit/c8c183eb62b666b5e9c92ca2cbf13f5464ae3aa9).

Related to #2292.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,7 @@
         }
     ],
     "require": {
-        "php": "^7.2.5",
+        "php": "~7.2.22 || ^7.3.9",
         "ext-json": "*",
         "guzzlehttp/promises": "^1.0",
         "guzzlehttp/psr7": "^1.6.1",
--- a/tests/Handler/CurlFactoryTest.php
+++ b/tests/Handler/CurlFactoryTest.php
@@ -1,9 +1,11 @@
 <?php
 namespace GuzzleHttp\Test\Handler;
 
+use GuzzleHttp\Exception\ConnectException;
 use GuzzleHttp\Handler;
 use GuzzleHttp\Handler\CurlFactory;
 use GuzzleHttp\Handler\EasyHandle;
+use GuzzleHttp\Promise\PromiseInterface;
 use GuzzleHttp\Psr7;
 use GuzzleHttp\Tests\Helpers;
 use GuzzleHttp\Tests\Server;
@@ -755,5 +757,23 @@ class CurlFactoryTest extends TestCase
         $easyHandle->handle = \curl_init();
 
         self::assertEmpty($factory->release($easyHandle));
+    }
+
+    public function testCurlErrorMessage(): void
+    {
+        $factory = new Handler\CurlFactory(1);
+        $request = new Psr7\Request('GET', 'http://' . uniqid() . uniqid());
+        $fn = static function (Psr7\Request $request, array $options) use (&$fn, $factory): PromiseInterface {
+            $easy = $factory->create($request, $options);
+            curl_exec($easy->handle);
+            $easy->errno = curl_errno($easy->handle);
+
+            return Handler\CurlFactory::finish($fn, $easy, $factory);
+        };
+
+        $this->expectException(ConnectException::class);
+        $this->expectExceptionMessage('Could not resolve host: ');
+
+        $fn($request, [])->wait();
     }
 }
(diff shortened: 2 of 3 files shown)


GrahamCampbell:
Hmm. Can we not work around the bug. Also, Guzzle 6 will be affected by this?

phansys:
> Hmm. Can we not work around the bug. Also, Guzzle 6 will be affected by this?

The fix for Guzzzle 6 is provided at #2292.

GrahamCampbell:
I think it would be better to delay the PHP version bump all the way to that, to Guzzle 8. Does the default travis 7.2 setup even run 7.2.22 or newer?

phansys:
> I think it would be better to delay the PHP version bump all the way to that, to Guzzle 8. Does the default travis 7.2 setup even run 7.2.22 or newer?

I think the default Travis setup for PHP 7.2 is using `PHP 7.2.15 (cli) (built: Feb 11 2019 23:22:17) ( ZTS )`.

GrahamCampbell:
So it would be a bad idea to include this change in Guzzle 7, since it would hurt adoption. Guzzle 7 is meant to be easy to upgrade to.

GrahamCampbell:
> I think the default Travis setup for PHP 7.2 is using PHP 7.2.15 (cli) (built: Feb 11 2019 23:22:17) ( ZTS ).

Yes. I just tried it, now:

image

composer.json
```diff
@@ -26,7 +26,7 @@
         }
     ],
     "require": {
-        "php": "^7.2.5",
+        "php": "^7.2.22",
```
GrahamCampbell: This should be `~7.2.22|^7.3.9`, since PHP 7.3.0-7.3.8 is also affected by the same bug as 7.2.0-7.2.11.

phansys: You're totally right. The constraint was updated.
Thank you.

composer.json
```diff
@@ -26,7 +26,7 @@
         }
     ],
     "require": {
-        "php": "^7.2.5",
+        "php": "~7.2.22 || ^7.3.9",
```
Nyholm: This syntax is way wrong. Please read https://developer.happyr.com/always-use-caret-instead-of-tilde

Nyholm: Im also not super happy about changing the PHP requirements. It is set to 7.2.5 because of that is the lowest, most supported PHP version on major Linux LTSes.

phansys: IMO, the [Tilde Version Range](https://getcomposer.org/doc/articles/versions.md#tilde-version-range-) operator is the right choice for the desired version resolution. The constraint will respect SemVer, since it only allows to update to any minor or patch version from 7.x branch.
If you have any doubt about this behavior, please check https://semver.mwl.be/ ("php" is not recognized as a package there since it only handles the packages published in Packagist, but you can test with any other package and the result should be the same).

phansys: IMO, regardless how available is a

phansys:
> Thank you.
> 
> The PHP error is an edge case, ie most users will not be effected. I dont believe we should care about it.

Thanks for the feedback. I just left some concerns about your comments.

composer.json
Nyholm: Im sorry, you are correct.

Nyholm: Are you saying the Guzzle is unusable when running 7.2.5?

phansys: I'm not saying it is unusable, just that a feature provided by a dependency (the CURL extension) is not behaving as expected.
I noticed the issue since a piece of code in a private project was checking the error message (an edge case, like you mentioned).

Nyholm:
Since this is a bug in an edge case. I vote to close this issue. 
There will be many bugs in PHP and I don't believe we should update our patch version as soon as PHP releases a new version. 

I believe a vast majority of Guzzle users can use Guzzle running 7.2.5. So updating the PHP requirements would hurt more than it helps.

sagikazarmark:
@phansys Thanks a lot for the reports and PRs, nice catch.

I have to agree with the others though. For Guzzle 7 adoption has a higher priority at the moment, not to mention that this isn't actually a fix in Guzzle, but making Guzzle unusable on a whole bunch of PHP versions.

The real solution is upgrading, but this issue is not severe enough IMHO to justify forcing an upgrade.

To make not of the issue, we could add a comment in the code with links to these PRs and the PHP bug ticket, but other than that I don't think this issue needs fixing.

Thanks again!

Issue #1870, opened 2017-07-04
Leading spaces in request headers are not handled correctly.


katmsft:
| Q            | A
| ------------ | ---
| Bug?         | yes
| New Feature? | no
| Version      | 6.2.1

#### Actual Behavior

If there are leading spaces in request headers, Guzzle will merge the header into previous header, causing an error. Also Guzzle does not trim or throw exception for trailing spaces in header's field name.

#### Expected Behavior

In [HTTP Spec 3.2.4](http://greenbytes.de/tech/webdav/draft-ietf-httpbis-p1-messaging-22.html#rfc.section.3.2.4.p.3), it is stated that, 

> No whitespace is allowed between the header field-name and colon.

So trailing spaces in header's field name should cause Guzzle to trim it, or better throw an exception, and leading spaces in field name should not cause header to be merged into previous header.

#### Steps to Reproduce

Code:
```
$headers0 = ['   doublespace   ' => '   doublespace   '];
$headers1 = ['nonspace' => 'nonspace'];
$headers2 = ['   frontspace' => '   frontspace'];
$headers3 = ['trailspace   ' => 'trailspace   '];
$request0 = new Request('GET', 'http://www.google.com', $headers0);
$request1 = new Request('GET', 'http://www.google.com', $headers1);
$request2 = new Request('GET', 'http://www.google.com', $headers2);
$request3 = new Request('GET', 'http://www.google.com', $headers3);
$client = new Client();

$result = $client->send($request0);
$result = $client->send($request1);
$result = $client->send($request2);
$result = $client->send($request3);
```

Result:
```
GET http://www.google.com/ HTTP/1.1
Host: www.google.com
User-Agent: GuzzleHttp/6.2.1 curl/7.49.1 PHP/5.5.38 doublespace   : doublespace

GET http://www.google.com/ HTTP/1.1
Host: www.google.com
User-Agent: GuzzleHttp/6.2.1 curl/7.49.1 PHP/5.5.38
nonspace: nonspace

GET http://www.google.com/ HTTP/1.1
Host: www.google.com
User-Agent: GuzzleHttp/6.2.1 curl/7.49.1 PHP/5.5.38 frontspace: frontspace

GET http://www.google.com/ HTTP/1.1
Host: www.google.com
User-Agent: GuzzleHttp/6.2.1 curl/7.49.1 PHP/5.5.38
trailspace   : trailspace
```

katmsft:
[...]

PR #2540, opened 2020-01-09
Added test for checking header field that contains white spaces

adrianosferreira:
Reported in https://github.com/guzzle/guzzle/issues/1870


adrianosferreira:
I'm closing this PR as it seems this kind of test is already covered in Node:

```
TypeError [ERR_INVALID_HTTP_TOKEN]: Header name must be a valid HTTP token [" key "]
```

Issue #2507, opened 2019-12-28
Guzzle exceptions


Nyholm:
I dont believe we have the best structure of our exceptions. 

The exception hierarchy currently look like this: 

```
\InvalidArgumentException
-- InvalidArgumentException

\RuntimeException
-- TransferException 
-- -- RequestException
-- -- -- TooManyRedirectsException
-- -- -- ConnectException
-- -- -- BadResponseException
-- -- -- -- ClientException
-- -- -- -- ServerException
```

I suggest we update it to: 

```
\InvalidArgumentException
-- InvalidArgumentException

\RuntimeException
-- RequestException
-- TooManyRedirectsException
-- ConnectException
-- BadResponseException
-- -- ClientException
-- -- ServerException
```

GrahamCampbell:
Maybe we should keep `TransferException`. Useful for catching stuff that went wrong with the http request, but not other errors, probably caused by bad code.

Nyholm:
I did not see a point in keeping both `ConnectException` and `TransferException`. I agree with you that we should have such exception. I think it is an issue with naming. In this proposal I think  `ConnectException`, `TransferException` and  `NetworkException` are all doing the same thing.

GrahamCampbell:
What about:

```
\InvalidArgumentException
-- InvalidArgumentException

\RuntimeException
-- RuntimeException
-- -- TransferException
-- -- -- RequestException
-- -- -- TooManyRedirectsException
-- -- -- ConnectException
-- -- -- BadResponseException
-- -- -- -- ClientException
-- -- -- -- ServerException
```

GrahamCampbell:
The naming of `TransferException` matches what HTTPlug does, and their `RequestException` extends it.

Nyholm:
Hehe. Im not super happy with HTTPlug's exceptions either =)

What is the point of `TransferException`? All exceptions implements the `GuzzleException` interface so you can use it if you want to "catch everything".

GrahamCampbell:
But what if you don't want to catch Guzzle exceptions not caused by the actual HTTP request-response stuff?

Nyholm:
> But what if you don't want to catch Guzzle exceptions not caused by the actual HTTP request-response stuff?

Im not sure I 100% understand. Could you rephrase that without the double negative? And what is the "HTTP request-response stuff"? Do you mean the network layer?

sagikazarmark:
[...]

PR #2541, opened 2020-01-09
PSR-18: Separate ConnectException from RequestException

davidgrayston:
Guzzle version affected: `7.0.0-beta.1`

_Relates to #2507 / https://github.com/php-http/httplug/issues/155_

### Description
Update exception hierarchy to conform with https://www.php-fig.org/psr/psr-18/meta/
>The domain exceptions NetworkExceptionInterface and RequestExceptionInterface define a contract very similar to each other. The chosen approach is to not let them extend each other because inheritance does not make sense in the domain model. A RequestExceptionInterface is simply not a NetworkExceptionInterface.

### Changed
- `ConnectException` now extends `TransferException` instead of `RequestException`
  > Note: request, response and handler context methods have been moved into traits for reuse - there might be a neater/preferable way

#### New hierarchy:
```
\RuntimeException
-- TransferException
-- -- ConnectException (Psr\Http\Client\NetworkExceptionInterface)
-- -- RequestException (Psr\Http\Client\RequestExceptionInterface)
-- -- -- TooManyRedirectsException
-- -- -- BadResponseException
-- -- -- -- ClientException
-- -- -- -- ServerException
```

I've tried to do this with as few breaking changes as possible - you could instead introduce a new exception that implements `Psr\Http\Client\RequestExceptionInterface` which would extend `RequestException` like this _(which would be non-breaking)_:
```
\RuntimeException
-- TransferException
-- -- RequestException
-- -- -- ConnectException (Psr\Http\Client\NetworkExceptionInterface)
-- -- -- RequestFailedException (Psr\Http\Client\RequestExceptionInterface)
-- -- -- -- TooManyRedirectsException
-- -- -- -- BadResponseException
-- -- -- -- -- ClientException
-- -- -- -- -- ServerException
```


src/Exception/ConnectException.php
```diff
@@ -10,22 +12,32 @@
...
         array $handlerContext = []
     ) {
-        parent::__construct($message, $request, null, $previous, $handlerContext);
+        parent::__construct($message, 0, $previous);
+        $this->setRequest($request);
+        $this->setHandlerContext($handlerContext);
     }
 
+    /**
+     * @deprecated 8.0.0
```
GrahamCampbell: We can't mark this as deprecated until the 7.1.0 release, so these annotations will need to come in a followup PR after this one, if this is merged into what will be 7.0.0.

src/Exception/ConnectException.php
```diff
@@ -10,15 +12,19 @@
  *
  * Note that no response is present for a ConnectException
  */
-class ConnectException extends RequestException implements NetworkExceptionInterface
+class ConnectException extends TransferException implements NetworkExceptionInterface
```
gmponos: If this gets approved we will need to update the UPGRADE guide and also the changelog.. Added as a note here in order not to forget.

davidgrayston: I've added a note to _UPGRADING.md_ in 3edf0a8

davidgrayston:
@gmponos I can remove use of traits if preferable?

I was also wondering if the following methods could be removed with a note added to `UPGRADING.md`?
- `GuzzleHttp\Exception\ConnectException::getResponse()`
- `GuzzleHttp\Exception\ConnectException::hasResponse()`

src/Exception/RequestException.php
```diff
@@ -11,15 +13,11 @@
  */
 class RequestException extends TransferException implements RequestExceptionInterface
 {
-    /** @var RequestInterface */
-    private $request;
+    use RequestAwareTrait, HandlerContextAwareTrait;
```
GrahamCampbell: If we are following PSR-12 and/or Symfony code style guidelines, then this needs to be:

```php
use RequestAwareTrait;
use HandlerContextAwareTrait;
```

davidgrayston: Fixed in 8ad0f9d

davidgrayston:
Would you prefer the PHPDoc for `getHandlerContext()` be duplicated in each exception, or could this be moved into a handler context interface, with each implementation using inheritdoc?

Nyholm:
You can duplicate it. 
It is "just exceptions" and it is just twice.

Nyholm:
Generally, change log updates should not be done in a PR. That will lead to merge conflicts.

Tobion:
With these changes, we should also change RequestException to make the response non-optional. The response was only optional for ConnectException. But now ConnectException does not extend RequestException anymore. All sub-classes of RequestException have a response now AFAIK.

Tobion:
Looks like TooManyRedirectsException does not have a response but should have. The response is just not forwarded from what I see.

Nyholm:
Thank you @Tobion

sergiz:
The docs haven't been updated and still display old hiearchy: https://docs.guzzlephp.org/en/stable/quickstart.html#exceptions

PR #2542, opened 2020-01-10
fix problem double backslash

dhanyn10:
fix double backslash `\\` that automatically added when we use forward slash `/`. in my case it cause error in data processing.
i have a data that must be stay same when it sended to a third party API, it's contains a special character like forward slash.


sagikazarmark:
If you need such behavior, I suggest you construct the `body` itself instead of passing data in the json field.

PR #2546, opened 2020-01-11
[7.0] Avoid functions due to global install conflicts

GrahamCampbell:
Don't call the GuzzleHttp "functions"; instead use class static methods. The functions themselves could be marked as deprecated in 7.1.0.

---

This is the full version of https://github.com/guzzle/guzzle/pull/2548, which was only applied to the new internal functions added on or after 6.4.0, which are directly causing problems for people.

### DIFF
--- a/src/Utils.php
+++ b/src/Utils.php
@@ -1,0 +1,380 @@
+<?php
+namespace GuzzleHttp;
+
+use GuzzleHttp\Exception\InvalidArgumentException;
+use GuzzleHttp\Handler\CurlHandler;
+use GuzzleHttp\Handler\CurlMultiHandler;
+use GuzzleHttp\Handler\Proxy;
+use GuzzleHttp\Handler\StreamHandler;
+use Psr\Http\Message\UriInterface;
+
+final class Utils
+{
+    /**
+     * Debug function used to describe the provided value type and class.
+     *
+     * @return string Returns a string containing the type of the variable and
+     *                if a class is provided, the class name.
+     */
+    public static function describeType($input): string
+    {
+        switch (\gettype($input)) {
+            case 'object':
+                return 'object(' . \get_class($input) . ')';
+            case 'array':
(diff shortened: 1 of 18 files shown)


Nyholm:
I like this PR. But you have kept the functions. That means there could still be conflicts, right?

GrahamCampbell:
I kept them to make sure that upgrading to guzzle 7 is easy. We should deprecate in 7.1.0, and remove in 8.0.0, perhaps?

GrahamCampbell:
There will be no conflicts within the internals, at least, so upgrading to Guzzle 7 won't result in Guzzle 7 calling globally installed guzzle 6 functions.

src/Utils.php
```diff
@@ -0,0 +1,380 @@
...
+namespace GuzzleHttp;
+
+use GuzzleHttp\Exception\InvalidArgumentException;
+use GuzzleHttp\Handler\CurlHandler;
+use GuzzleHttp\Handler\CurlMultiHandler;
+use GuzzleHttp\Handler\Proxy;
+use GuzzleHttp\Handler\StreamHandler;
+use Psr\Http\Message\UriInterface;
+
+final class Utils
```
Nyholm: Please me more defensive here. 

Mark it as `@internal` and  write a comment to explain the intent of the class.

GrahamCampbell: I don't think this class is internal only? We do intend for people to call these class methods, instead of the functions.

Nyholm: Sure, It is just me that want to get rid of the functions AND eventually this class. 

We can discuss that in a separate PR or issue.

GrahamCampbell:
@Nyholm I also have a lighter version of this PR for 6.5.x, which only moves the internal-marked functions to a Utils class: https://github.com/guzzle/guzzle/pull/2548. In particular, this fixes problems when an older version of 6.x is installed globally, but 6.5.x is installed locally.

PR #2548, opened 2020-01-12
[6.5] Don't use internal functions

### LINKED ISSUE #2511: Call to undefined function GuzzleHttp_idn_uri_convert()
Hello - after the December 2019 updates, we are receiving the following error. Our client code calls into Guzzle at #2, so we don't call GuzzleHttp_idn_uri_convert() directly:

[28-Dec-2019 17:57:20 UTC] PHP Fatal error: Uncaught Error: Call to undefined function GuzzleHttp_idn_uri_convert() in wp-content/plugins/q4vrplugin/vendor/guzzlehttp/guzzle/src/Client.php:220
Stack trace:
#0 guzzlehttp/guzzle/src/Client.php(155): GuzzleHttp\Client->buildUri(Object(GuzzleHttp\Psr7\Uri), Array)

#1 guzzlehttp/guzzle/src/Client.php(183): GuzzleHttp\Client->requestAsync('GET', 'units', Array)

#2 plugin/Models/ApiClient.php(97): GuzzleHttp\Client->request('GET', 'units', Array)

Please let me know what additional information I can provide to assist.

Thanks!

Jim

GrahamCampbell:
Since these functions are internal, this should not be an issue. We maybe could even delete those functions in 7.0.0, outright. A more complete version is available at https://github.com/guzzle/guzzle/pull/2546, targetted at 7.0, since it does the same for the non-internal functions.

---

Fixes #2511.

### DIFF
--- a/src/Utils.php
+++ b/src/Utils.php
@@ -1,0 +1,67 @@
+<?php
+namespace GuzzleHttp;
+
+use GuzzleHttp\Exception\InvalidArgumentException;
+use Psr\Http\Message\UriInterface;
+
+final class Utils
+{
+    /**
+     * Wrapper for the hrtime() or microtime() functions
+     * (depending on the PHP version, one of the two is used)
+     *
+     * @return float|mixed UNIX timestamp
+     *
+     * @internal
+     */
+    public static function currentTime()
+    {
+        return function_exists('hrtime') ? hrtime(true) / 1e9 : microtime(true);
+    }
+
+    /**
+     * @param int $options
+     *
+     * @return UriInterface
+     * @throws InvalidArgumentException
+     *
+     * @internal
+     */
+    public static function idnUriConvert(UriInterface $uri, $options = 0)
+    {
+        if ($uri->getHost()) {
+            $idnaVariant = defined('INTL_IDNA_VARIANT_UTS46') ? INTL_IDNA_VARIANT_UTS46 : 0;
+            $asciiHost = $idnaVariant === 0
+                ? idn_to_ascii($uri->getHost(), $options)
+                : idn_to_ascii($uri->getHost(), $options, $idnaVariant, $info);
+            if ($asciiHost === false) {
+                $errorBitSet = isset($info['errors']) ? $info['errors'] : 0;
+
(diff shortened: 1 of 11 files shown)


src/Utils.php
```diff
@@ -0,0 +1,67 @@
...
+    {
+        if ($uri->getHost()) {
+            $idnaVariant = defined('INTL_IDNA_VARIANT_UTS46') ? INTL_IDNA_VARIANT_UTS46 : 0;
+            $asciiHost = $idnaVariant === 0
+                ? idn_to_ascii($uri->getHost(), $options)
+                : idn_to_ascii($uri->getHost(), $options, $idnaVariant, $info);
+            if ($asciiHost === false) {
+                $errorBitSet = isset($info['errors']) ? $info['errors'] : 0;
+
+                $errorConstants = array_filter(array_keys(get_defined_constants()), function ($name) {
```
phansys:
```suggestion
                $errorConstants = array_filter(array_keys(get_defined_constants()), static function ($name) {
```

src/Utils.php
```diff
@@ -0,0 +1,67 @@
...
+                    }
+                }
+
+                $errorMessage = 'IDN conversion failed';
+                if ($errors) {
+                    $errorMessage .= ' (errors: ' . implode(', ', $errors) . ')';
+                }
+
+                throw new InvalidArgumentException($errorMessage);
+            } else {
```
phansys: IMO, this `else` clause is superfluous and can be safely omitted.

src/Utils.php
GrahamCampbell: Not making any code changes. We can do this on 7.x. This is strictly a bug fix PR for 6.5.x.

src/Utils.php
GrahamCampbell: Not making any code changes. We can do this on 7.x. This is strictly a bug fix PR for 6.5.x.

GrahamCampbell: All I'm doing is moving the code around, and not changing it.

Nyholm:
Thank you again @GrahamCampbell

PR #2549, opened 2020-01-12
[7.1] Avoid global functions in dependencies

GrahamCampbell:
EDIT: The new aim of this PR is avoid the use of global functions provided by our dependencies. This means using their new Utils classes, and bumping their minimum versions (staying within the same major release series).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,8 +28,8 @@
     "require": {
         "php": "^7.2.5",
         "ext-json": "*",
-        "guzzlehttp/promises": "^1.0",
-        "guzzlehttp/psr7": "^1.6.1",
+        "guzzlehttp/promises": "^2.0",
+        "guzzlehttp/psr7": "^2.0",
         "psr/http-client": "^1.0"
     },
     "require-dev": {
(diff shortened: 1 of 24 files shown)


Nyholm:
Could we support version 1 and 2 at the same time?

GrahamCampbell:
Depends what we want to do with the functions.

GrahamCampbell:
@Nyholm I wonder if I should add `Utils` classes to the 1.x versions of these two libraries, and deprecate the functions, and then remove them in 2.x? That way, we could easily support both versions, without needing 3.x to finally delete the functions?

GrahamCampbell:
Or maybe even, to minimize BC, only add the `Utils` classes, but don't change the 1.x internals to calling them yet, and don't deprecate the functions yet. Then make the internal changes in 2.0, deprecations in 2.1, and removals in 3.0?

GrahamCampbell:
> Could we support version 1 and 2 at the same time?

New plan: only update to `guzzlehttp/promises:^1.4` and `guzzlehttp/psr7:^1.7` in this PR, using the new `Utils` classes.

Nyholm:
Sound good!

GrahamCampbell:

PR #2551, opened 2020-01-13
[8.0] Remove the getConfig function

GrahamCampbell:
This depends on https://github.com/guzzle/guzzle/pull/2516, which in turn depends on https://github.com/guzzle/guzzle/pull/2515.


GrahamCampbell:

Issue #2498, opened 2019-12-23
Finalize the list with php-cs-fixer rules


gmponos:
I would like to propose and we can discuss about a final list with rules for php-cs-fixer:

```
        '@PSR2' => true,
        'array_syntax' => ['syntax' => 'short'],
        'blank_line_after_opening_tag' => true,
        'concat_space' => ['spacing' => 'one'],
        'declare_strict_types' => false,
        'function_to_constant' => ['functions' => ['php_sapi_name']],
        'fully_qualified_strict_types' => true,
        'header_comment' => false,
        'lowercase_cast' => true,
        'is_null' => ['use_yoda_style' => false],
        'magic_method_casing' => true,
        'modernize_types_casting' => true,
        'multiline_comment_opening_closing' => true,
        'native_constant_invocation' => true,
        'no_alias_functions' => true,
        'no_alternative_syntax' => true,
        'no_blank_lines_after_phpdoc' => true,
        'no_empty_comment' => true,
        'no_empty_phpdoc' => true,
        'no_extra_blank_lines' => true,
        'no_leading_import_slash' => true,
        'no_spaces_around_offset' => true,
        'no_superfluous_phpdoc_tags' => ['allow_mixed' => true],
        'no_trailing_comma_in_singleline_array' => true,
        'no_unset_cast' => true,
        'no_unused_imports' => true,
        'no_useless_else' => true,
        'no_useless_return' => true,
        'no_whitespace_in_blank_line' => true,
        'ordered_imports' => true,
        'php_unit_mock' => ['target' => 'newest'],
        'php_unit_dedicate_assert_internal_type' => ['target' => 'newest'],
        'php_unit_expectation' => ['target' => 'newest'],
        'php_unit_no_expectation_annotation' => ['target' => 'newest'],
        'php_unit_test_annotation' => ['style' => 'prefix'],
        'php_unit_test_case_static_method_calls' => ['call_type' => 'self'],
        'php_unit_mock_short_will_return' => true,
        'phpdoc_align' => ['align' => 'vertical'],
        'phpdoc_line_span' => ['method' => 'multi', 'property' => 'multi'],
        'phpdoc_scalar' => true,
        'phpdoc_separation' => true,
        'phpdoc_single_line_var_spacing' => true,
        'phpdoc_trim' => true,
        'phpdoc_trim_consecutive_blank_line_separation' => true,
        'phpdoc_types' => true,
[...]

PR #2553, opened 2020-01-13
Configure and apply "static_lambda" CS rule

phansys:


GrahamCampbell:
👎 for merging this right now. We should discuss if we want this change in https://github.com/guzzle/guzzle/issues/2498, and then apply all the code style fixes in one go, to avoid creating noise that will cause conflicts with other PRs, more than once.

phansys:
>  for merging this right now. We should discuss if we want this change in #2498, and then apply all the code style fixes in one go, to avoid creating noise that will cause conflicts with other PRs, more than once.

I agree. Please, be free to close if you consider this must be addressed in other PR.

Nyholm:
Im closing this since we wait for #2498

Issue #2539, opened 2020-01-07
Handlers default timeout discrepancy


kkopachev:
This issue is derived from #1940 

**Guzzle version(s) affected**: 6.3.0+
**PHP version**: 7+
**cURL version**: irrelevant

**Description**  
Default timeout/connect_timeout for `CurlHandler` is not set, meaning it will wait forever. On the other hand, `StreamHandler` with default settings will wait [default-socket-timeout ini setting](http://php.net/manual/en/filesystem.configuration.php#ini.default-socket-timeout) and does not look like there is a way to make it wait forever, since it [only cares about positive numbers](https://github.com/guzzle/guzzle/blob/db9e5/src/Handler/StreamHandler.php#L426-L431)

Documentation [says default timeout of 0 is to wait indefinitely](http://docs.guzzlephp.org/en/stable/request-options.html#timeout), which is not true for stream handler.

**Possible Solution**  
My proposal is to configure default curl timeout to reasonable number.
If that's not possible, then at least make StreamHandler consistent with CurlHandler implementation and make it possible to wait forever and use that as default.

I'd be happy to open PR, but not sure on course of action here.

PR #2559, opened 2020-01-18
[WIP] [7.0] Default connect_timeout on curl to 300 seconds

GrahamCampbell:
Closes https://github.com/guzzle/guzzle/pull/1524.

### DIFF
--- a/docs/request-options.rst
+++ b/docs/request-options.rst
@@ -278,9 +278,9 @@ connect_timeout
 ---------------
 
 :Summary: Float describing the number of seconds to wait while trying to connect
-        to a server. Use ``0`` to wait indefinitely (the default behavior).
+        to a server. Use ``0`` to wait indefinitely.
 :Types: float
-:Default: ``0``
+:Default: ``300``
 :Constant: ``GuzzleHttp\RequestOptions::CONNECT_TIMEOUT``
 
 .. code-block:: php
--- a/src/Handler/CurlFactory.php
+++ b/src/Handler/CurlFactory.php
@@ -418,9 +418,16 @@ class CurlFactory implements CurlFactoryInterface
             }
         }
 
+        // treat 0 or less as "no timeout" and default to 300 seconds
         if (isset($options['connect_timeout'])) {
-            $timeoutRequiresNoSignal |= $options['connect_timeout'] < 1;
-            $conf[CURLOPT_CONNECTTIMEOUT_MS] = $options['connect_timeout'] * 1000;
+            if ($options['connect_timeout'] <= 0) {
+                $timeoutRequiresNoSignal = true;
+                $conf[CURLOPT_CONNECTTIMEOUT_MS] = 0;
+            } else {
+                $conf[CURLOPT_CONNECTTIMEOUT_MS] = $options['connect_timeout'] * 1000;
+            }
+        } else {
+            $conf[CURLOPT_CONNECTTIMEOUT_MS] = 300000;
         }
 
         if ($timeoutRequiresNoSignal && \strtoupper(\substr(PHP_OS, 0, 3)) !== 'WIN') {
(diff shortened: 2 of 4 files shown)


Nyholm:
Awesome. Thank you!

GrahamCampbell:
Just pushed another change here.

Nyholm:
Im trying to find where we define the default timeout for the stream. 

Could you show me? I want to verify that we have the same timeout.

GrahamCampbell:
The stream handler doesn't have a connect timeout, since it's not supported by the handler. It only has a "timeout".

Nyholm:
Yeah, but we dont sent at default value for the "timeout", are we?

GrahamCampbell:
Ah, yeh. What do we want for a timeout value?

GrahamCampbell:
Default "timeout" to "infinity", and "connect timeout" to "300 seconds"?

Nyholm:
Is this correct?

### Curl: 
- Connect timeout: 300s (changed in this PR)
- Timeout: infinity
- Read timeout: default_socket_timeout

### Stream:
- Connect timeout: 300s (cannot be configured)
- Timeout: infinity
- Read timeout: default_socket_timeout

GrahamCampbell:
No. I don't think the stream handler has a concept of a connect timeout. I'm not too sure.

GrahamCampbell:
There is `stream_set_timeout`, which defaults to 60 seconds. That is what controls how long we wait between receiving the last byte, and timing out, I think. Maybe that controls the connect timeout too?

Nyholm:
The goal of this PR should be that this code:

```php
// Will block for X seconds
$client->get('https://no-url.com');

```

should be blocking for X seconds. No matter if one use the curl handler or the socket handler. Since curl is way more configurable. We should look at what we can configure in the socket handler, then set sensible defaults and mimic those settings in curl. 

So the first thing we need to figure out how socket behave.

kkopachev:
This ticket related to #2539

kkopachev:
For streams:
looks like when using `fopen` with `http` handler, `connect_timeout` defaults to `default_socket_timeout`. Then if context resource is passed with configured `http => timeout` value, it is used as `read timeout`. 

Here is my path of thought:

`fsockopen` (unlike `fopen`) can receive connect [timeout as parameter](https://www.php.net/manual/en/function.fsockopen.php) which defaults to [default_socket_timeout](https://github.com/php/php-src/blob/d1764ca33018f1f2e4a05926c879c67ad4aa8da5/ext/standard/fsock.c#L34). It is only used as connect_timeout. Then it passed to [php_stream_xport_create](https://github.com/php/php-src/blob/d1764ca33018f1f2e4a05926c879c67ad4aa8da5/ext/standard/fsock.c#L72-L81) as parameter, which handles socket connection and etc and returns stream.

`fopen` with http handler reads timeout value [from context options](https://github.com/php/php-src/blob/5d6e923d46a89fe9cd8fb6c3a6da675aa67197b4/ext/standard/http_fopen_wrapper.c#L195), defaulting to [default_socket_timeout](https://github.com/php/php-src/blob/5d6e923d46a89fe9cd8fb6c3a6da675aa67197b4/ext/standard/http_fopen_wrapper.c#L208) ini setting. Then it passes timeout value to the same [php_stream_xport_create](https://github.com/php/php-src/blob/5d6e923d46a89fe9cd8fb6c3a6da675aa67197b4/ext/standard/http_fopen_wrapper.c#L213-L215) and opens the stream. After successful opening and connection, it [sets read timeout](https://github.com/php/php-src/blob/5d6e923d46a89fe9cd8fb6c3a6da675aa67197b4/ext/standard/http_fopen_wrapper.c#L218) to the same timeout value which was passed as context option. 
From my understanding, it is possible to change read_timeout for open stream by calling `stream_set_timeout`. However, that might be too late already, since php handled sending request headers, received  response headers and at the beginning of data stream (or possibly read some of it already into buffer).

mfn:
?

GrahamCampbell:
Closed because the implementation was not right, and it's not clear what the right implementation is. There is already an open issue tracking this bug, so closing this doesn't hide that issue.

PR #2563, opened 2020-01-18
[WIP] [7.0] Suggest ext-curl, and require 7.21.2 for running the tests

GrahamCampbell:


GrahamCampbell:
Hmm, does the extension version not correspond to the lib version?

Nyholm:
Im not sure you can use extension versions at all.

GrahamCampbell:
Yeh. I think you're right. Will just add the suggest ext-curl.

PR #2565, opened 2020-01-19
[7.0] Updated installation docs

GrahamCampbell:


docs/overview.rst
```diff
@@ -48,7 +48,7 @@ existing composer.json file:
 
     {
       "require": {
-         "guzzlehttp/guzzle": "~6.0"
+         "guzzlehttp/guzzle": "^7.0"
```
Nyholm: If you change this to 6 and `composer require guzzlehttp/guzzle:^6.0`, then I can merge right away. Otherwise this PR have to wait until 7.0 is released.

GrahamCampbell: Why? Aren't these the docs for master, and not 6.x. The 6.x docs are on the 6.x branch?

Nyholm: You are correct. But that is not how users work I’m afraid. 

I want to avoid issues like “I cannot install this”

GrahamCampbell: Can we update "stable" to point to 6.x, and default the docs to showing "stable", so the "latest" docs are not shown by default to users?

GrahamCampbell: Just like how the symfony docs `master` is the unreleased dev stuff, but the website shows the latest stable.

Nyholm: No, I would rather focus on getting a new release out by the end of the month. And I would like the experience to be as good as possible for as many users as possible.

GrahamCampbell: Then, we can just wait, and merge this closer to the release?

Nyholm:
Could you please rebase this PR?

GrahamCampbell:
Doesn't need rebasing.

GrahamCampbell:
Done. Ready to merge. :)

PR #2567, opened 2020-01-21
[6.5] Remove obsolete statement in `handler` option docs

guilliamxavier:
Fixes https://github.com/guzzle/guzzle/pull/1852#issuecomment-501727140

(Of course you can also decide to change the code to deprecate/forbid per-request override of  `handler`, but that would be a BC break, thus not before 7.0/8.0)


gmponos:
I do agree with this change.. but I would feel more comfortable if someone else from maintainers had a look as well..

Nyholm:
Thank you!

Issue #1919, opened 2017-09-18
Add `bearer` as an auth option


ghost:
Is there any reason why this is still not an option for the auth config options?

piotr-cz:
Can be done as adding new case [in Client](https://github.com/guzzle/guzzle/blob/6.3.0/src/Client.php#L346-L358):

```php
case 'bearer':
    $modify['set_headers']['Authorization'] = 'Bearer ' . $value[0];
    break;
```

usage:
```php
$response = $this->httpClient->request('GET', 'http://www.google.com', [
    GuzzleHttp\RequestOptions::AUTH => ['<token>', null, 'Bearer']
]);
```

gmponos:
closing https://github.com/guzzle/guzzle/pull/2568#issuecomment-650544967

PR #2568, opened 2020-01-21
Add "Bearer" as an auth option

hamzahejja:
- This PR addresses issue #1919 
- Adds Bearer Token Authorization to Client Requests.
- Added needed test.

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -405,6 +405,11 @@ class Client implements ClientInterface, \Psr\Http\Client\ClientInterface
                     $options['curl'][CURLOPT_HTTPAUTH] = CURLAUTH_NTLM;
                     $options['curl'][CURLOPT_USERPWD] = "$value[0]:$value[1]";
                     break;
+                case 'bearer':
+                    // Ensure that we don't have the Authorization header in
+                    // different case and set the new value wih the token
+                    $modify['set_headers'] = Psr7\_caseless_remove(['Authorization'], $modify['set_headers']);
+                    $modify['set_headers']['Authorization'] = 'Bearer ' . $value[0];
             }
         }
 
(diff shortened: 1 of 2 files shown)


Nyholm:
I don’t see how this is a more convenient way than just setting the authorization header yourself. If we feel that we really want to include this, then I’m fine with the implementation. (Good job btw)

I’ve added the milestone 7.1.

hamzahejja:
@Nyholm I agree with what you've said, it makes sense and is sufficient to only set the `Authorization` header with the bearer token. The reason why I implemented this was per issue #1919 which was **Accepting PRs**. I Appreciate the fast feedback!

twoixter:
@Nyholm FWIW, I found this PR while wondering why `RequestOptions::AUTH` didn't recognise `bearer` as built-in, as it is widely used because of OAuth 2.0 nowadays. I mean, it's not just a user-defined method in which you're expected to manually add the `Authorization` header.

It is a great feature for tools that saves configuration options so that I don't have to `switch (...)` or similar just for the sake of adding the header or using `RequestOptions::AUTH`.

src/Client.php
```diff
@@ -405,6 +405,11 @@ private function applyOptions(RequestInterface $request, array &$options): Reque
                     $options['curl'][CURLOPT_HTTPAUTH] = CURLAUTH_NTLM;
                     $options['curl'][CURLOPT_USERPWD] = "$value[0]:$value[1]";
                     break;
+                case 'bearer':
+                    // Ensure that we don't have the Authorization header in
+                    // different case and set the new value wih the token
```
davidgrayston, line 410:
```suggestion
                    // different case and set the new value with the token
```
Noticed this minor typo

Nyholm:

gmponos:
Hi.. @hamzahejja thank you for the contribution.. 

This was my bad marking the original issue as `accept PRs`

For now this seems like a not needed change. As @Nyholm said developers can easily set the `Authorization` header

PR #2576, opened 2020-01-31
Handle createResponse()-exceptions in curl-header-function correctly

### LINKED ISSUE #2266: Invalid HTTP response can crash Guzzle-Event-Loop
**Guzzle version(s) affected**: 6.5.1

**Description**  
If a (broken) server responds with a "HTTP/1.1 404: Not Found" header (note the wrong colon!) **after being redirected**, the promise is not normally rejected, but the `$promise->wait();` call throws an Exception ("Status code must be an integer value.") - and exits the event-loop completely (very bad when having concurrent requests).

**How to reproduce**  
Start two "HTTP test servers" ...

... one on port 18000 that redirects to port 18001:
```
cat <<EOF | nc -N -l -p 18000
HTTP/1.1 301 Moved Permanently
Connection: close
Location: http://localhost:18001/

EOF
```

... and one on port 18001 that sends the broken header with the colon after "404":
```
cat <<EOF | nc -N -l -p 18001
HTTP/1.1 404: Not Found
Connection: close

EOF
```

... then execute this script:
```php
$url = 'http://localhost:18000/';

try {
    $client->requestAsync('GET', $url)->then(function ($response) {
        echo 'Response Status-Code: ' . $response->getStatusCode() . PHP_EOL;
    }, function ($e) {
        echo 'Normal Error: ' . $e->getMessage() . PHP_EOL;
    })->wait();
} catch (\Exception $e) {
    echo 'Unexpected Error (should never happen): ' . $e->getMessage() . PHP_EOL;
}
```

It will run into the "Unexpected Error"-catch.

**Possible Solution**  
The problem is, that `$easy->createResponse();` in https://github.com/guzzle/guzzle/blob/master/src/Handler/CurlFactory.php#L569 can throw exceptions, but is not inside a try-catch.

Maybe it is the best, to just expand the try-catch for "$onHeaders"-callback so that the `$easy->createResponse();` is covered, too?

not-implemented:
Invalid status codes like "404:" (with a colon) are accepted by curl, but createResponse() throws an exception, that should not break the curl event loop.

I fixed it by extending the try/catch block around createResponse(). Then the exception is thrown as "An error was encountered during the on_headers event".

This fixes #2266

### DIFF
--- a/src/Handler/CurlFactory.php
+++ b/src/Handler/CurlFactory.php
@@ -562,16 +562,16 @@ class CurlFactory implements CurlFactoryInterface
             $value = trim($h);
             if ($value === '') {
                 $startingResponse = true;
-                $easy->createResponse();
-                if ($onHeaders !== null) {
-                    try {
+                try {
+                    $easy->createResponse();
+                    if ($onHeaders !== null) {
                         $onHeaders($easy->response);
-                    } catch (\Exception $e) {
-                        // Associate the exception with the handle and trigger
-                        // a curl header write error by returning 0.
-                        $easy->onHeadersException = $e;
-                        return -1;
                     }
+                } catch (\Exception $e) {
+                    // Associate the exception with the handle and trigger
+                    // a curl header write error by returning 0.
+                    $easy->onHeadersException = $e;
+                    return -1;
                 }
             } elseif ($startingResponse) {
                 $startingResponse = false;


not-implemented:
@GrahamCampbell ... may I ask you for a review on this issue? :-) ... Or could you link someone adequate?

See #2266 for details.

not-implemented:
@Nyholm ... may I ask you for a review on this issue? :-) I rebased the pull-request to be up2date.

This is still a very annoying problem when connecting to foreign broken HTTP servers.

See #2266 for details.

Nyholm:
Just a question: Would #2591 also fix this issue?

strarsis:
The Response constructor performs a status code check and throws a plain InvalidArgumentException.

Could it throw something like a guzzle [BadResponseException](https://github.com/guzzle/guzzle/blob/c4c348a4f1ecd86ed33e88adbd32741c74cd322b/src/Exception/BadResponseException.php) so the response reference/body can still be used by the client?

not-implemented:
Yes, it seems that #2591 addresses the same problem in a better way, so this can be closed as duplicate. Thanks!

PR #2581, opened 2020-02-13
Run CI with PHP 7.3

phil-davis:
and re-arrange Travis for PHP 5.4 and 5.4 to use `trusty` so that those old PHP versions are found.

This demonstrates that unit tests pass on PHP 7.3
That will make people feel happier that `guzzle5` can be used up to PHP  7.3

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,11 @@
 language: php
 
 php:
-  - 5.4
-  - 5.5
   - 5.6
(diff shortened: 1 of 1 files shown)


phil-davis:
Would this be OK to merge, just to demonstrate that guzzle 5 is still working with PHP 7.3?
There is nothing to "release", just CI matrix changes.

PR #2582, opened 2020-02-13
Test php 7.4

phil-davis:
see what happens:

- phpunit 4 reports deprecations with PHP 7.4, likely that is not going to work happily
- phpunit 9 reports `Fatal error: Uncaught Error: Class 'PHPUnit_Framework_TestCase' not found in /home/travis/build/guzzle/guzzle/tests/BatchResultsTest.php on line 9` - because that is now `PHPUnit\Framework\TestCase` - so that would be painful to sort out test code that will run with both phpunit 4 and 9
- phpunit 8 reports similar problems with `PHPUnit_Framework_TestCase`


phil-davis:
Hmmm - there are lots of unit test changes needed to support the newer versions of phpunit.
That will be time-consuming to sort out.

The initial try with PHP 7.4 and phpunit4 was https://travis-ci.org/guzzle/guzzle/jobs/649778451?utm_medium=notification&utm_source=github_status

That gave only "harmless" deprecation warnings. The tests themselves passed, so probably guzzle5 will work OK on PHP 7.4

Not much point putting a lot of effort into this. Even though it can be painful, people should move to Guzzzle6 anyway :)

PR #2586, opened 2020-02-15
Update license year

mvdnbrk:


GrahamCampbell:
Hmm, maybe just change it to 2011.

mvdnbrk:
`date('Y')`
😄

GrahamCampbell:
Well, no actually. It needs to be the earliest year, not the current year.

mvdnbrk:
https://www.gnu.org/licenses/gpl-howto.html

The copyright notice should include the year in which you finished preparing the release (so if you finished it in 1998 but didn't post it until 1999, use 1998). You should add the proper year for each past release; for example, “Copyright 1998, 1999 Terry Jones” if some releases were finished in 1998 and some were finished in 1999. If several people helped write the code, use all their names.

For software with several releases over multiple years, it's okay to use a range (“2008-2010”) instead of listing individual years (“2008, 2009, 2010”) if and only if every year in the range, inclusive, really is a “copyrightable” year that would be listed individually; and you make an explicit statement in your documentation about this usage.

GrahamCampbell:
This is not a GNU license.

GrahamCampbell:
Though, actually in the US, for a long time, a year has not actually been needed for a copyright to be valid. ;)

mvdnbrk:
Hm probably depends on where your are.
I am not a legal expert.

The year can also be removed completely. Pick an option ;)

GrahamCampbell:
The reason for my suggest was just to eliminate the need to update it each year.

LICENSE
```diff
@@ -1,4 +1,4 @@
-Copyright (c) 2011-2018 Michael Dowling, https://github.com/mtdowling <mtdowling@gmail.com>
+Copyright (c) 2011-2020 Michael Dowling, https://github.com/mtdowling <mtdowling@gmail.com>
```
Nyholm, line 1:
```suggestion
Copyright (c) 2011Michael Dowling, https://github.com/mtdowling <mtdowling@gmail.com>
```

gmponos: @Nyholm you forgot a space here?

Nyholm: Thank you

PR #2589, opened 2020-02-16
Ensure RequestException always has response

davidgrayston:
> Attempts to address https://github.com/guzzle/guzzle/pull/2541#issuecomment-586730619 by @Tobion 

### Changed
- `GuzzleHttp\Exception\RequestException` now always has a response

#### I think there are a few possible problems with these changes:
- `RequestException::wrapException()` now requires a response, so exceptions cannot be wrapped in `GuzzleHttp\Handler\StreamHandler::__invoke()`
- `CurlFactory::createRejection()` assumes `ConnectException` when `$easy->response` is `null`

### DIFF
--- a/src/Exception/RequestException.php
+++ b/src/Exception/RequestException.php
@@ -34,7 +34,6 @@ class RequestException extends TransferException implements RequestExceptionInte
         \Throwable $previous = null,
         array $handlerContext = []
     ) {
-        // Set the code of the exception if the response is set and not future.
         parent::__construct($message, $response->getStatusCode(), $previous);
         $this->request = $request;
         $this->response = $response;
--- a/src/Exception/RequestException.php
+++ b/src/Exception/RequestException.php
@@ -18,7 +18,7 @@ class RequestException extends TransferException implements RequestExceptionInte
     private $request;
 
     /**
-     * @var ResponseInterface|null
+     * @var ResponseInterface
      */
     private $response;
 
@@ -30,13 +30,12 @@ class RequestException extends TransferException implements RequestExceptionInte
     public function __construct(
         string $message,
         RequestInterface $request,
-        ResponseInterface $response = null,
+        ResponseInterface $response,
         \Throwable $previous = null,
         array $handlerContext = []
     ) {
         // Set the code of the exception if the response is set and not future.
-        $code = $response ? $response->getStatusCode() : 0;
-        parent::__construct($message, $code, $previous);
+        parent::__construct($message, $response->getStatusCode(), $previous);
         $this->request = $request;
         $this->response = $response;
         $this->handlerContext = $handlerContext;
@@ -45,11 +44,14 @@ class RequestException extends TransferException implements RequestExceptionInte
     /**
      * Wrap non-RequestExceptions with a RequestException
      */
-    public static function wrapException(RequestInterface $request, \Throwable $e): RequestException
-    {
+    public static function wrapException(
+        RequestInterface $request,
+        ResponseInterface $response,
+        \Throwable $e
+    ): RequestException {
         return $e instanceof RequestException
             ? $e
-            : new RequestException($e->getMessage(), $request, null, $e);
+            : new RequestException($e->getMessage(), $request, $response, $e);
     }
 
     /**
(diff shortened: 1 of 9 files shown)


GrahamCampbell:
I think this change might have been planned for Guzzle 8, but intentionally not included in Guzzle 7.x, to minimize BC breakage?

src/Handler/StreamHandler.php
```diff
@@ -63,8 +63,6 @@ public function __invoke(RequestInterface $request, array $options): PromiseInte
                 || \strpos($message, "connection attempt failed")
             ) {
                 $e = new ConnectException($e->getMessage(), $request, $e);
-            } else {
-                $e = RequestException::wrapException($request, $e);
```
Tobion: Not sure what kind of exception can actually occur at this point.

GrahamCampbell: Why do we have the if statement here then? Surely it can be entire removed, with lines 59 to 68 (original line numbers), simply replaced with the single line `$e = new ConnectException($e->getMessage(), $request, $e);`.

davidgrayston: The exceptions being tested in `testVerifiesVerifyIsValidIfPath()` and `testVerifiesCertIfValidPath()` occur here. 

Perhaps they should be wrapped in a `TransferException`?

davidgrayston: Or perhaps they should be invalid argument exceptions like this: https://github.com/guzzle/guzzle/blob/master/tests/Handler/CurlFactoryTest.php#L115

davidgrayston: I've made these changes in dc1591c - I can revert if you think it's incorrect.

src/Handler/CurlFactory.php
```diff
@@ -359,8 +359,14 @@ private function applyHandlerOptions(EasyHandle $easy, array &$conf): void
...
-                    if (\is_dir($options['verify']) ||
-                        (\is_link($options['verify']) && \is_dir(\readlink($options['verify'])))) {
+                    if (
+                        \is_dir($options['verify']) ||
+                        (
+                            \is_link($options['verify']) === true &&
+                            \readlink($options['verify']) !== false &&
+                            \is_dir(\readlink($options['verify']))
+                        )
+                    ) {
```
davidgrayston, line 369: This was flagged by phpstan:
```
 ------ ----------------------------------------------------------- 
  Line   Handler/CurlFactory.php                                    
 ------ ----------------------------------------------------------- 
  363    Parameter #1 $filename of function is_dir expects string,  
         string|false given.                                        
 ------ ----------------------------------------------------------- 
```
I can alternatively ignore this?

GrahamCampbell: Don't call readlink over and over. Assign it to a variable, then you will avoid both slow code and warnings.

GrahamCampbell (changes requested):

src/Handler/CurlFactory.php
davidgrayston: I've now squashed that into 1 commit 0c2f920

davidgrayston: This fix has been applied on master, so no longer on this PR

src/Handler/StreamHandler.php
davidgrayston: Please unresolve if this isn't correct

davidgrayston:
I realised most of the changes in this PR aren't necessary in order to have the response on the `TooManyRedirectsException` - I can open a separate PR with just the changes in `src/RedirectMiddleware.php` if you would like to have that as a fix for 7?

src/Handler/CurlFactory.php
```diff
@@ -169,7 +169,7 @@ private static function createRejection(EasyHandle $easy, array $ctx): PromiseIn
 
         // If an exception was encountered during the onHeaders event, then
         // return a rejected promise that wraps that exception.
-        if ($easy->onHeadersException) {
+        if ($easy->response && $easy->onHeadersException) {
```
davidgrayston, line 172: I'm not sure if this is the correct fix for the PHPStan error, as there will always be a `response` when `onHeadersException` is set.

We could do something like this instead https://github.com/davidgrayston/guzzle/commit/76d370b96d3e289bf910234fa002ba0d2e2293ea, which makes a few assumptions:
- No response = `ConnectException`
- There's a future possibility that `onHeadersException` can be present without a response (seems unlikely)

src/Exception/RequestException.php
```diff
@@ -30,13 +30,11 @@ class RequestException extends TransferException implements RequestExceptionInte
     public function __construct(
         string $message,
         RequestInterface $request,
-        ResponseInterface $response = null,
+        ResponseInterface $response,
```
Nyholm, line 33: This is a BC break

Nyholm:
CI is failing. 

I’m also not sure they it is possible to always provide a response to this exception.

GrahamCampbell:
👎 here from me. Certainly not for Guzzle 7.

davidgrayston:
This was originally intended for Guzzle 7 before it was released, but would have to be for Guzzle 8 now.

I don't think this change is needed, shall we close this PR?

Nyholm:
Thank you

Issue #2534, opened 2020-01-02
Guzzle Pool promise finishes before all requests are complete


williamjulianvicary:
**Guzzle version(s) affected**: 6.5.2
**PHP version**: 7.3.9-1+ubuntu18.04.1+deb.sury.org+1 
**cURL version**: 7.58.0

**Description**  
I'm seeing an intermittent issue (frequent enough - the below may need to be ran a handful of times to see the problem) where, it appears that the Pool is finishing prematurely and the `->wait()` method call is unwrapping the most recent exception - even though there may be many other requests waiting to be processed.

I've pasted a sample simple Laravel command below, which requests 100 URLs that all return a 999 response code (something that the assertions fail in the Response generation). This sometimes runs with no issue, but often (say 1 in 4 times) the call to `->wait()` bubbles an exception before all other requests are finished.

**How to reproduce**  
```<?php

namespace App\Console\Commands;

use GuzzleHttp\Client;
use GuzzleHttp\Pool;
use GuzzleHttp\Psr7\Request;
use GuzzleHttp\Psr7\Response;
use GuzzleHttp\TransferStats;
use Illuminate\Console\Command;
use Log;

class TestCommand extends Command
{
    /**
     * The name and signature of the console command.
     *
     * @var string
     */
    protected $signature = 'test';

    /**
     * The console command description.
     *
     * @var string
     */
    protected $description = 'Command description';
    /**
     * @var \Closure
     */
    private $successCallback;
    /**
     * @var \Closure
     */
    private $failureCallback;

    /**
     * Create a new command instance.
     *
     * @return void
     */
    public function __construct()
    {
        parent::__construct();
        $this->failureCallback = function($reason, $index) {
            var_dump($reason->getMessage());
        };

        $this->successCallback = function ($response, $index) {
        };
    }

    public function getRequests()
    {
        for ($i=0; $i < 100; $i++) {
            yield new Request('get', 'https://httpstat.us/999');
        }
    }

    /**
     * Execute the console command.
     *
     * @return mixed
     */
    public function handle()
    {
        $client = new Client(['timeout' => 10, 'verify' => false]);

        $pool = new Pool($client, $this->getRequests(), [
            'concurrency' => '5',
            'fulfilled' => function(Response $response, $index) {
                call_user_func($this->successCallback, $response, $index);
            },
            'rejected' => function($reason, $index) {
                call_user_func($this->failureCallback, $reason, $index);
            }
        ]);

        $promise = $pool->promise();

[...]

PR #2591, opened 2020-02-18
Handle exceptions during response creation

williamjulianvicary:
Catches exceptions thrown during the createResponse method of the easy handler to avoid crashes during the headers function used to parse the headers. 

Potential additional changes:
- Add validation to EasyHandle.php to validate the status code and pre-empt the exception thrown from the PSR7 Response
- Pass a more sensible message from the exception into the RequestException raised and passed into the rejection.

Resolves issues discussed here: https://github.com/guzzle/guzzle/issues/2534


graid2030:
Can somebody check it?

tam5:
bumping this up, any chance to get it merged?

strarsis:
Please merge!

williamjulianvicary:
I’ve opened a fresh issue here https://github.com/guzzle/guzzle/issues/2762 as it has been some time since this pull request and issue was opened (and a new major version has been released, which still contains this bug).

strarsis:
@williamjulianvicary: In my case I have an OpenAPI generated PHP client that uses guzzle.
The API server responds with an unusual HTTP status code in some cases, guzzle will then throw an `InvalidArgumentException`.
I am unable to retrieve the response body then, as the `InvalidArgumentException` exception is not an extended Exception with guzzle request/response instances, etc as with other errors.

Nyholm:
Im missing a test for this. Could you add one please?

GrahamCampbell:
This should target master I think?

GrahamCampbell:
To be included in the 7.1.0 release.

Nyholm:
> To be included in the 7.1.0 release.

Or 7.1.1. 

Either way, we need a test before merging.

GrahamCampbell:
So the PR target branch needs changing from 6.5 to master?

strarsis:
@GrahamCampbell, @Nyholm: Would this PR also make it possible to get the response body/response object of responses that cause an `InvalidArgumentException` because of a HTTP status code that is outside the RFC range?
The thrown `InvalidArgumentException` doesn't contain a reference to the originating response object, or has means to retrieve the response body.

Nyholm:
That seams like a very different PR. 

If we keep the PRs small and only fixes one thing, then they will be easier to review and merge.

williamjulianvicary:
I’ll get a test added to the branch and target merge master later. With regard to a test, I found the underlying issue to be intermittent but I’ll see what I can come up with and update the PR.

strarsis:
@Nyholm: This would be related to this issue: https://github.com/guzzle/psr7/issues/349
So this PR fixes a similar issue, but not the linked one?

Nyholm:
@strarsis Feel free to open a new PR for that issue.

strarsis:
@Nyholm: Not sure about the approach. Should the `InvalidArgumentException` be extended by a Guzzle exception, 
so the original response / response body can be passed with that new exception?

williamjulianvicary:
Test added to check that a rejection is sent. I've needed to build an additional static helper on the server.php helper. This is because the root cause of this problem stems from PSR7\Response - so I'm unable to use that to mock the invalid response. The helper I've added allows passing of raw data to enqueue on to the mock response based on the enqueue() method of this class. Hopefully that's suitable, otherwise if you have any bright ideas how I'd go about otherwise testing this I'm happy to adjust the approach.

Nyholm:
I marked this as 7.2. But ideally we will merge this PR next and tag a 7.1.1 release.

strarsis:
@Nyholm: Not wanting to sound like a broken record, but would this PR also allow to retrieve the response/body in case of an InvalidArgumentException from an invalid HTTP status code?

Nyholm:
> @Nyholm: Not wanting to sound like a broken record, but would this PR also allow to retrieve the response/body in case of an InvalidArgumentException from an invalid HTTP status code?

If you are looking at the [PR contents](https://github.com/guzzle/guzzle/pull/2591/files): No, it does not look like it does. 

Feel free to create a new PR that solves https://github.com/guzzle/psr7/issues/349.

williamjulianvicary:
@strarsis @Nyholm This produces a rejection that does pass in the originating exception as the native previous exception. If you see the test I added, you can do this:

```        try {
            $promise->wait();
        } catch  (\GuzzleHttp\Exception\RequestException $e) {
            var_dump($e->getPrevious()->getMessage()); // "Status code must be an integer value between 1xx and 5xx."
        }
```

Nyholm:
@williamjulianvicary Could I ask you to rebase this PR so the tests will pass? 

I would also ask everybody else for a review.

src/Handler/CurlFactory.php
```diff
@@ -178,6 +178,18 @@ private static function createRejection(EasyHandle $easy, array $ctx): PromiseIn
             \CURLE_GOT_NOTHING          => true,
         ];
 
+        if ($easy->createResponseException) {
+            return \GuzzleHttp\Promise\rejection_for(
```
GrahamCampbell, line 182: This function is deprecated. You need to use the static class method.

tests/Server.php
```diff
@@ -75,6 +75,32 @@ public static function enqueue($responses)
...
+            [
+                'status'  => (string) $statusCode,
+                'reason'  => $reasonPhrase,
+                'headers' => $headers,
+                'body'    => \base64_encode((string) $body)
+            ]
+        ];
+
+        self::getClient()->request('PUT', 'guzzle-server/responses', [
+            'json' => $data
```
GrahamCampbell, line 100: Missing trailing comma.

src/Handler/CurlFactory.php
GrahamCampbell: @Nyholm It's a shame phpstan is not loaded using composer so that we can enable checking for deprecated calls.

GrahamCampbell (changes requested):

src/Handler/CurlFactory.php
williamjulianvicary: Updated - it looks like master had changed since I updated this PR, latest commit should be good.

tests/Server.php
williamjulianvicary: This doesn't seem to follow the standard set with the method above this (enqueue) which doesn't use trailing commas on arrays for a very similar request - just wanted to confirm that?

src/Handler/CurlFactory.php
Nyholm: > It's a shame phpstan is not loaded using composer so that we can enable checking for deprecated calls.

Are you sure about that? 

Im not sure that function was deprecated when PHPStan was running on this PR.

tests/Server.php
Nyholm: As long as the CI passes, then Im fine. 

We should consider update our CS rules if the CI is green with the current code (But that is a separate PR)

Nyholm (approved):
Thank you

strarsis:
So this PR also now allows to read the response body/response instance in case of an `InvalidArgumentException`? 😸 t

williamjulianvicary:
@strarsis As has been mentioned multiple times in this thread. No.

strarsis:
@williamjulianvicary: OK, the confusion arose as in the related issue this PR has been mentioned as closing.

Issue #2270, opened 2019-03-01
unable to set `CURLOPT_ENCODING` to empty string to send all supported encoding by curl


deepika-maj:
**Guzzle version(s) affected**: 5.3.3

**Description**  
How can I set CURLOPT_ENCODING to empty string as following config is not working:

```
$httpClient = new Client([
    'curl' => [
        CURLOPT_ENCODING => ''
    ]
]);
```

gmponos:
> Guzzle version(s) affected: 5.3.3

Are you using guzzle v5?

if not.. and you are using v6.. how did you validated that the curl option is not working?

We will need a reproducing example that clearly displays a bug.

PR #2595, opened 2020-02-23
Fix CURLOPT_ENCODING not to be overwritten

YoakeSakura:
This PR fixes an issue where CURLOPT_ENCODING is overwritten when decode_content is true (default is true).

Before this fix, setting only CURLOPT_ENCODING removes the Accept-Encoding header and the server response is not compressed.
After this fix, the Accept-Encoding header is sent and the server response is compressed.

Maybe this will solve that issue #2270

### DIFF
--- a/src/Handler/CurlFactory.php
+++ b/src/Handler/CurlFactory.php
@@ -394,7 +394,7 @@ class CurlFactory implements CurlFactoryInterface
             }
         }
 
-        if (!empty($options['decode_content'])) {
+        if (!isset($options['curl'][\CURLOPT_ENCODING]) && !empty($options['decode_content'])) {
             $accept = $easy->request->getHeaderLine('Accept-Encoding');
             if ($accept) {
                 $conf[\CURLOPT_ENCODING] = $accept;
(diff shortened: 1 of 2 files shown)


Nyholm:
This needs a rebase and a test. 

Also make sure CI is green

Nyholm:
Friendly ping

Nyholm:
Ping ping =)

Nyholm:
@YoakeSakura will you finish this PR?

YoakeSakura:
Thanks for the ping, sorry for the delay.
I added the test code.
MockHandler could not test the request headers that cURL adds.
So I'm testing with Server and MockHandler.

Incidentally, I think it's a good idea to add an option to automatically set the compression that cURL can support.
Now the only way to do that is to set it as a cURL option.

Nyholm (approved):
Fair. So if you specify `$options['curl'][\CURLOPT_ENCODING]` you dont want Guzzle to override it with data from the `Accept-Encoding` heading.

GrahamCampbell (approved):

Nyholm:
Thank you

PR #2610, opened 2020-04-06
Update quickstart.rst

georaldc:
Update "Concurrent requests" examples when using Promise unwrap and settle functions


georaldc:
I guess I could, but I see that @juan-morales already addressed it in a recent commit. My change last month was just slightly more verbose.

georaldc:
Updated my branch

PR #2621, opened 2020-04-20
Fix use and phpDoc for `@throws` GuzzleException/InvalidArgumentException

guilliamxavier:


guilliamxavier:
The remaining PHPStan error is unrelated to this changeset... it says that `readlink` can return `false`, on this line: https://github.com/guzzle/guzzle/blob/aab4ebd862aa7d04f01a4b51849d657db56d882e/src/Handler/CurlFactory.php#L359

PR #2624, opened 2020-04-22
Small docblock fix.

shehi:


shehi:
Apparently PHPStan configuration doesn't fit this v6 of this package anymore?

GrahamCampbell:
> Apparently PHPStan configuration doesn't fit this v6 of this package anymore?

The first error was actually directly caused by your PR. The second was caused by the upgrade of phpstan from .18 to .19. Please just re-generate the baseline file with your PR.

shehi:
> The first error was actually directly caused by your PR. 
Not really, coz that specific exception is dynamic one (please check it's source). It is Throwable wherever Throwable is supported:
```
<?php
namespace GuzzleHttp\Exception;

use Throwable;

if (interface_exists(Throwable::class)) {
    interface GuzzleException extends Throwable
    {
    }
} else {
    /**
     * @method string getMessage()
     * @method \Throwable|null getPrevious()
     * @method mixed getCode()
     * @method string getFile()
     * @method int getLine()
     * @method array getTrace()
     * @method string getTraceAsString()
     */
    interface GuzzleException
    {
    }
}```

I guess for this reason you guys got rid of this check in master branch?

GrahamCampbell:
> Not really, coz that specific exception is dynamic one (please check it's source).

If you revert your changes, you will see the warning go away...

GrahamCampbell:
Evidence that I am correct is within the CI run of https://github.com/guzzle/guzzle/pull/2626.

shehi:
Thank you, that's enough of me :)

GrahamCampbell:
This PR is still valid though, just the baseline file needs correcting. The current phpdoc references a class/interface that doesn't exist.

GrahamCampbell:
The fact that it doesn't extend throwable on all systems goes away on later versions of Guzzle.

PR #2626, opened 2020-04-24
[6.5] Fix various intl icu issues

### LINKED ISSUE #2620: INTL_IDNA_VARIANT_2003 is deprecated
**Guzzle version(s) affected**: 6.5.3
**PHP version**: 7.3.16 (hint: `php --version`)
**cURL version**: 7.69.1 (hint: `php -i | grep cURL`)

**Description**  
My Telegram Bot is giving me this error when upgrade from 6.4.0 to 6.5.3:

idn_to_ascii(): INTL_IDNA_VARIANT_2003 is deprecated
--
in Utils.php line 35
at HandleExceptions->handleError(8192, 'idn_to_ascii(): INTL_IDNA_VARIANT_2003 is deprecated', '/home/erich33/vendor/guzzlehttp/guzzle/src/Utils.php', 35, array('uri' => object(Uri), 'options' => 0, 'idnaVariant' => 0))

**How to reproduce**  
When somebody message to my Telegram Bot, I have a script which send to my email the Exceptions, and turns now a hell with this error.

**Possible Solution**  
Downgrade to 6.4?
Something with symfony idn?

**Additional context**  
My actual composer.lock:

{
    "name": "laravel/laravel",
    "type": "project",
    "description": "The Laravel Framework.",
    "keywords": [
        "framework",
        "laravel"
    ],
    "license": "MIT",
    "require": {
        "php": "^7.3",
        "artesaos/seotools": "^0.17.1",
        "binarytorch/larecipe": "^2.2",
        "biscolab/laravel-recaptcha": "^4.0",
        "facebook/graph-sdk": "^5.7",
        "fideloper/proxy": "^4.0",
        "guzzlehttp/guzzle": "^6.5",
        "intervention/image": "^2.5",
        "irazasyed/telegram-bot-sdk": "3.1.0",
        "laravel-notification-channels/facebook": "^0.1.0",
        "laravel-notification-channels/facebook-poster": "^2.0",
        "laravel-notification-channels/telegram": "^0.3.0",
        "laravel-notification-channels/twitter": "^3.0",
        "laravel/framework": "^6.2",
        "laravel/passport": "^8.4",
        "laravel/socialite": "^4.1",
        "laravel/tinker": "^2.2",
        "overtrue/laravel-like": "^2.0",
        "predis/predis": "^1.1",
        "renatomarinho/laravel-page-speed": "^1.8",
        "rennokki/befriended": "^3.0",
        "rennokki/rating": "^2.0",
        "spatie/laravel-permission": "^3.11",
        "spatie/laravel-responsecache": "^6.3",
        "spatie/laravel-sitemap": "^5.3",
        "spatie/schema-org": "^2.2",
        "tedivm/fetch": "^0.7.1"
    },
    "require-dev": {
        "barryvdh/laravel-debugbar": "^3.2",
        "beyondcode/laravel-dump-server": "^1.0",
        "filp/whoops": "^2.0",
        "fzaninotto/faker": "^1.4",
        "mockery/mockery": "^1.0",
        "nunomaduro/collision": "^3.0",
        "phpunit/phpunit": "^9.0"
    },
    "config": {
        "optimize-autoloader": true,
        "preferred-install": "dist",
        "sort-packages": true
    },
    "extra": {
        "laravel": {
            "dont-discover": []
        }
    },
    "autoload": {
        "files": [
            "app/Helpers/Helper.php"
        ],
        "psr-4": {
            "App\\": "app/"
        },
        "classmap": [
            "database/seeds",
            "database/factories"
        ]
    },
    "autoload-dev": {
        "psr-4": {
            "Tests\\": "tests/"
        }
    },
    "minimum-stability": "dev",
    "prefer-stable": true,
    "scripts": {
        "post-autoload-dump": [
            "Illuminate\\Foundation\\ComposerScripts::postAutoloadDump",
            "@php artisan package:discover --ansi",
            "@php artisan optimize",
            "@php artisan view:clear",
            "@php artisan cache:clear",
            "@php artisan responsecache:clear"
        ],
        "post-root-package-install": [
            "@php -r \"file_exists('.env') || copy('.env.example', '.env');\""
        ],
        "post-create-project-cmd": [
            "@php artisan key:generate --ansi"
        ]
    }
}

### LINKED ISSUE #2640: IDN conversion fails with domains containing multiple hyphens
**Guzzle version(s) affected**: 6.5.3
**PHP version**: 7.3.14
**cURL version**: 7.64.0

**Description**  
With some domains containing multiple hyphens, IDN conversion fails with this error:
```
PHP Fatal error:  Uncaught GuzzleHttp\Exception\InvalidArgumentException: IDN conversion failed (errors: IDNA_ERROR_HYPHEN_3_4) in /home/pierre/www/alltube/vendor/guzzlehttp/guzzle/src/Utils.php:56
```

**How to reproduce**  
```php
<?php

use GuzzleHttp\Client;

require_once __DIR__ . '/vendor/autoload.php';

$client = new Client();
$client->get('https://r3---sn-25glene6.example.com/');
```

**Possible Solution**  
The workaround I am using:
```php
$client = new Client(['idn_conversion' => false]);
```

GrahamCampbell:
This fixes the following issues:

1. Whenever PHP has been compiled against a too old version of intl icu, we should use the polyfill implementation (Fixes #2620).
2. A minimum version of the polyfill of 1.17.0 should be imposed to fix issues with constants being defined in error (Fixes https://github.com/guzzle/guzzle/pull/2626#issuecomment-627452837).
3. When a domain is only ASCII, we should skip conversion entirely (Fixes #2640).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,7 @@
     "require": {
         "php": ">=5.5",
         "ext-json": "*",
-        "symfony/polyfill-intl-idn": "^1.11",
+        "symfony/polyfill-intl-idn": "^1.17",
         "guzzlehttp/promises": "^1.0",
         "guzzlehttp/psr7": "^1.6.1"
     },
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -496,6 +496,11 @@ parameters:
 			path: src/Handler/CurlFactory.php
 
 		-
+			message: "#^Parameter \\#1 \\$filename of function is_dir expects string, string\\|false given\\.$#"
+			count: 1
+			path: src/Handler/CurlFactory.php
+
+		-
 			message: "#^Method GuzzleHttp\\\\Handler\\\\CurlFactory\\:\\:retryFailedRewind\\(\\) has no return typehint specified\\.$#"
 			count: 1
 			path: src/Handler/CurlFactory.php
@@ -1309,6 +1314,11 @@ parameters:
 			message: "#^Method GuzzleHttp\\\\UriTemplate\\:\\:isAssoc\\(\\) has parameter \\$array with no value type specified in iterable type array\\.$#"
 			count: 1
 			path: src/UriTemplate.php
+
+		-
+			message: "#^Method GuzzleHttp\\\\Utils\\:\\:idnToAsci\\(\\) has parameter \\$info with no value type specified in iterable type array\\.$#"
+			count: 1
+			path: src/Utils.php
 
 		-
 			message: "#^Function GuzzleHttp\\\\uri_template\\(\\) has parameter \\$variables with no value type specified in iterable type array\\.$#"
--- a/src/Utils.php
+++ b/src/Utils.php
@@ -72,6 +72,10 @@ final class Utils
      */
     private static function idnToAsci($domain, $options, &$info = [])
     {
+        if (\preg_match('%^[ -~]+$%', $domain) === 1) {
+            return $domain;
+        }
+
         if (\extension_loaded('intl') && defined('INTL_IDNA_VARIANT_UTS46')) {
             return \idn_to_ascii($domain, $options, INTL_IDNA_VARIANT_UTS46, $info);
         }
--- a/src/Utils.php
+++ b/src/Utils.php
@@ -3,7 +3,7 @@ namespace GuzzleHttp;
 
 use GuzzleHttp\Exception\InvalidArgumentException;
 use Psr\Http\Message\UriInterface;
-use Symfony\Polyfill\Intl\Idn;
+use Symfony\Polyfill\Intl\Idn\Idn;
 
 final class Utils
 {
@@ -63,6 +63,13 @@ final class Utils
         return $uri;
     }
 
+    /**
+     * @param string $domain
+     * @param int    $options
+     * @param array  $info
+     *
+     * @return string|false
+     */
     private static function idnToAsci($domain, $options, &$info = [])
     {
         if (\extension_loaded('intl') && defined('INTL_IDNA_VARIANT_UTS46')) {
--- a/src/Utils.php
+++ b/src/Utils.php
@@ -3,6 +3,7 @@ namespace GuzzleHttp;
 
 use GuzzleHttp\Exception\InvalidArgumentException;
 use Psr\Http\Message\UriInterface;
+use Symfony\Polyfill\Intl\Idn;
 
 final class Utils
 {
@@ -30,10 +31,7 @@ final class Utils
     public static function idnUriConvert(UriInterface $uri, $options = 0)
     {
         if ($uri->getHost()) {
-            $idnaVariant = defined('INTL_IDNA_VARIANT_UTS46') ? INTL_IDNA_VARIANT_UTS46 : 0;
-            $asciiHost = $idnaVariant === 0
-                ? idn_to_ascii($uri->getHost(), $options)
-                : idn_to_ascii($uri->getHost(), $options, $idnaVariant, $info);
+            $asciiHost = self::idnToAsci($uri->getHost(), $options, $info);
             if ($asciiHost === false) {
                 $errorBitSet = isset($info['errors']) ? $info['errors'] : 0;
 
@@ -63,5 +61,14 @@ final class Utils
         }
 
         return $uri;
+    }
+
+    private static function idnToAsci($domain, $options, &$info = [])
+    {
+        if (\extension_loaded('intl') && defined('INTL_IDNA_VARIANT_UTS46')) {
+            return \idn_to_ascii($domain, $options, INTL_IDNA_VARIANT_UTS46, $info);
+        }
+
+        return Idn::idn_to_ascii($domain, $options, Idn::INTL_IDNA_VARIANT_UTS46, $info);
     }
 }


mbabker:
Rolled this to production, not getting the deprecation notice anymore.

jonnott:
+1 for getting this merged and released!

jonnott:
Is https://github.com/symfony/polyfill/issues/262 related?

GrahamCampbell:
Hmmm, not really. There were issues with the older version of the polyfill too.

jonnott:
@GrahamCampbell  @fabpot

Still getting 'IDN conversion failed' on all Guzzle Client requests, unless I specifically set 'idn_conversion=>false'

Now using 1.17.0 of symfony/polyfill-intl-idn, and latest Guzzle 6.5.3

Guzzle 6.5.1 fixed this issue so I was able to drop 'idn_conversion=>false', but now the problem is back and I don't know why. Upping to 6.5.3 didn't cause this regression I don't think, but the problem seems to be there whether I'm using 1.15, 1.16 or 1.17 of symfony/polyfill-intl-idn :(

GrahamCampbell:
You haven't tried applying this PR though...

GrahamCampbell:
This will fix the issue for you.

jonnott:
@GrahamCampbell Waiting for it to be in a release ... I tend to be happier to leave work-arounds in my code (like ['idn_conversion' => false]) than have forked/bodged-in-vendor-dir versions of dependencies.

GrahamCampbell:
The only way for this PR to make it into a release is if people (like yourself) test it so we know if it works.

phpstan-baseline.neon
```diff
@@ -495,6 +495,11 @@ parameters:
 			count: 1
 			path: src/Handler/CurlFactory.php
 
+		-
+			message: "#^Parameter \\#1 \\$filename of function is_dir expects string, string\\|false given\\.$#"
+			count: 1
+			path: src/Handler/CurlFactory.php
+
```
guilliamxavier, line 502: extract this unrelated change to another PR so that it can be merged ASAP?

jonnott:
> The only way for this PR to make it into a release is if people (like yourself) to test it so we know if it works.

@GrahamCampbell I would love to be able to help by directly testing the change .. but unfortunately the only environment in which I'm experiencing the issue is in production (which is CentOS 6 / ICU 4.2.1 / cPanel WHM / PHP 7.3) .. whereas in my local dev environment (MAMP 5.7, PHP 7.3, macOS Catalina, ICU 56.1) there's no problem at all.

So the only thing I can do (so production isn't broken), after researching the issue, is to set `idn_conversion=>false` for every `new GuzzleHttp\Client()`.

I would love to be able to help more, but my production environment is very busy running a constantly-used API + app, and is using Guzzle to handle all communication with external services and APIs.

I understand that I would somehow perhaps deploy the fixed branch into production to test it, but that's also a bit out of my league re: composer skills. Anyone got a link to a simple guide on how to do this?

What would help (as much of the detail of this IDN/ICU stuff is way over my head) .. would be to have a clear idea of how somewhere between guzzle 6.5.1 and 6.5.3, and/or symfony/polyfill-intl-idn 1.15.0 to 1.16.0 .. the 'IDN conversion failed' error (which had gone away) .. has returned ?

phpstan-baseline.neon
GrahamCampbell: It is not an urgent change. Would be a waste of everyone's time.

jonnott:
@GrahamCampbell Can you advise on how I could temporarily put your `icu-fixes` branch into my production environment (using composer) in order to test this?

jonnott:
Is the first answer here 

> @GrahamCampbell Can you advise on how I could temporarily put your `icu-fixes` branch into my production environment (using composer) in order to test this?

Like this? https://stackoverflow.com/questions/13498519/how-to-require-a-fork-with-composer

jonnott:
@GrahamCampbell  I can confirm this FIXES the problem for me, after testing in production..

`(new GuzzleHttp\Client())->get('https://some.url/here');` now works without needing `'idn_conversion'=>false` option setting for the Client.

WHM/cPanel (latest on 'release' channel)
CentOS 6
PHP 7.3.17

```
Internationalization support | enabled (intl)
-- | --
ICU version | 4.2.1
ICU TZData version | 2009j
ICU Unicode version | 5.1
```

composer.json file:

```
{
    "repositories": [
        {
            "type": "vcs",
            "url": "https://github.com/GrahamCampbell/guzzle"
        }
    ],
    "require": {
        "stripe/stripe-php": "^7.0",
        "drewm/mailchimp-api": "^2.4",
        "tijsverkoyen/akismet": "^1.1",
        "facebook/graph-sdk": "^5.4",
        "guzzlehttp/guzzle": "dev-icu-fixes",
        "calcinai/xero-php": "^1.7",
        "vlucas/phpdotenv": "^4.0",
        "edamov/pushok": "^0",
        "mpdf/mpdf": "^8.0",
        "plasticbrain/php-flash-messages": "^1.0",
        "symfony/cache": "^5.0"
    }
}
```

`composer show`

```
calcinai/xero-php                       v1.9.1                A client implementation of the Xero API, with a cleaner OAuth interface and ORM-like abstraction.
drewm/mailchimp-api                     v2.5.4                Super-simple, minimum abstraction MailChimp API v3 wrapper
edamov/pushok                           0.11.1                PHP client for Apple Push Notification Service (APNs) - Send push notifications to iOS using the new APNs HTTP/2 protocol with token-based (JWT with p8 private key) or ...
facebook/graph-sdk                      5.7.0                 Facebook SDK for PHP
fgrosse/phpasn1                         v2.1.1                A PHP Framework that allows you to encode and decode arbitrary ASN.1 structures using the ITU-T X.690 Encoding Rules.
guzzlehttp/guzzle                       dev-icu-fixes 6606165 Guzzle is a PHP HTTP client library
guzzlehttp/promises                     v1.3.1                Guzzle promises library
guzzlehttp/psr7                         1.6.1                 PSR-7 message implementation that also provides common utility methods
mpdf/mpdf                               v8.0.5                PHP library generating PDF files from UTF-8 encoded HTML
myclabs/deep-copy                       1.9.5                 Create deep copies (clones) of your objects
paragonie/random_compat                 v9.99.99              PHP 5.x polyfill for random_bytes() and random_int() from PHP 7
phpoption/phpoption                     1.7.3                 Option Type for PHP
plasticbrain/php-flash-messages         v1.0.1                A modern take on PHP session-based flash messages
psr/cache                               1.0.1                 Common interface for caching libraries
psr/container                           1.0.0                 Common Container Interface (PHP FIG PSR-11)
psr/http-client                         1.0.0                 Common interface for HTTP clients
psr/http-factory                        1.0.1                 Common interfaces for PSR-7 HTTP message factories
psr/http-message                        1.0.1                 Common interface for HTTP messages
psr/log                                 1.1.3                 Common interface for logging libraries
ralouphie/getallheaders                 3.0.3                 A polyfill for getallheaders.
setasign/fpdi                           v2.3.3                FPDI is a collection of PHP classes facilitating developers to read pages from existing PDF documents and use them as templates in FPDF. Because it is also possible to ...
spomky-labs/base64url                   v2.0.1                Base 64 URL Safe Encoding/Decoding PHP Library
stripe/stripe-php                       v7.32.0               Stripe PHP Library
symfony/cache                           v5.0.8                Symfony Cache component with PSR-6, PSR-16, and tags
symfony/cache-contracts                 v2.0.1                Generic abstractions related to caching
symfony/polyfill-ctype                  v1.17.0               Symfony polyfill for ctype functions
symfony/polyfill-intl-idn               v1.17.0               Symfony polyfill for intl's idn_to_ascii and idn_to_utf8 functions
symfony/polyfill-mbstring               v1.17.0               Symfony polyfill for the Mbstring extension
symfony/polyfill-php72                  v1.17.0               Symfony polyfill backporting some PHP 7.2+ features to lower PHP versions
symfony/service-contracts               v2.0.1                Generic abstractions related to writing services
symfony/var-exporter                    v5.0.8                A blend of var_export() + serialize() to turn any serializable data structure to plain PHP code
tijsverkoyen/akismet                    1.1.1                 Akismet is a wrapper-class to communicate with the Akismet API.
vlucas/phpdotenv                        v4.1.5                Loads environment variables from `.env` to `getenv()`, `$_ENV` and `$_SERVER` automagically.
web-token/jwt-core                      v2.1.6                Core component of the JWT Framework.
web-token/jwt-key-mgmt                  v2.1.6                Key Management component of the JWT Framework.
web-token/jwt-signature                 v2.1.6                Signature component of the JWT Framework.
web-token/jwt-signature-algorithm-ecdsa v2.1.6                ECDSA Based Signature Algorithms the JWT Framework.
```

jonnott:
> @GrahamCampbell I can confirm this FIXES the problem for me, after testing in production

NB: this includes your latest `Skip conversion for ASCII domains` (6606165) commit, that's what my `dev-icu-fixes` is checked-out at.

GrahamCampbell:
Yeh. Great. Thanks for confirming. I think this PR is good to be merged now. I'll contact the Guzzle core team.

src/Utils.php
```diff
@@ -64,4 +62,24 @@ public static function idnUriConvert(UriInterface $uri, $options = 0)
...
+    {
+        if (\preg_match('%^[ -~]+$%', $domain) === 1) {
+            return $domain;
+        }
+
+        if (\extension_loaded('intl') && defined('INTL_IDNA_VARIANT_UTS46')) {
+            return \idn_to_ascii($domain, $options, INTL_IDNA_VARIANT_UTS46, $info);
+        }
+
+        return Idn::idn_to_ascii($domain, $options, Idn::INTL_IDNA_VARIANT_UTS46, $info);
```
nicolas-grekas, line 83: Can this skip using the polyfill class directly? It's marked internal and it shouldn't be used.
Of not, why did you do it this way?

GrahamCampbell: Because we need to be able to call the polyfill even if the original extension is installed to work around the extension using a too old version of the C library. Symfony doesn't have a way for us to do this, other to call code that is formally internal.

jonnott: And this commit (included in polyfill-intl-idn 1.17 release) totally stops Guzzle doing anything other than calling the class method directly now..
https://github.com/symfony/polyfill-intl-idn/commit/3bff59ea7047e925be6b7f2059d60af31bb46d6a

jonnott: ...which to me suggests it should in fact be polyfill-intl-idn which should be verifying whether the ICU version available on the host machine is usable, or if the polyfill should be used instead, and it not be left to Guzzle to be look into this and decide for itself (therefore 'breaking' Symfony's rules on using code marked internal directly) .. ?

src/Utils.php
```diff
@@ -64,4 +62,24 @@ public static function idnUriConvert(UriInterface $uri, $options = 0)
...
+     *
+     * @return string|false
+     */
+    private static function idnToAsci($domain, $options, &$info = [])
+    {
+        if (\preg_match('%^[ -~]+$%', $domain) === 1) {
+            return $domain;
+        }
+
+        if (\extension_loaded('intl') && defined('INTL_IDNA_VARIANT_UTS46')) {
```
stof, line 79: you should remove the `extension_loaded` check. That way, if you have only the polyfill installed, you will still call `idn_to_ascii` (from the polyfill) rather than the internal API.

src/Utils.php
stof: @jonnott not possible. Userland implementations cannot replace functions defined by PHP.

jonnott: Oh well, looks like this PR is the best solution we have then. Merge away @mtdowling & co. ;)

stof: @jonnott this commit in the polyfill fixes detecting wrongly that `INTL_IDNA_VARIANT_UTS46` is supported by `idn_to_ascii`. It does not change the fact that `idn_to_ascii` is defined by the extension when defined (and so the polyfill could not define it).

src/Utils.php
nicolas-grekas: There will always be a case where using the polyfill directly will be needed, so anyway...

BUT it's very possible that ppl forced the polyfill to be NOT installed, by using a "replace" for "intl" in their composer.json.

This "if" should add a `|| !class_exists(Idn::class)` to safeguard against this.

nicolas-grekas:
Check https://github.com/guzzle/guzzle/pull/2626#discussion_r425253546 (I commented on a resolved comment, and gh didn't "unresolve it")

src/Utils.php
GrahamCampbell: Hmm, if anyone has done this, we'd expect a fatal error. I think that would be fine, since it indicates the class doesn't exist, and anyone who has forced non-installation will know that that means.

nicolas-grekas: require ext-intl + replace symfony/polyfill-intl is going to be very common.
We recommend it when possible for other polyfilled extensions.
A fatal error would be totally WTF I think...

GrahamCampbell: Another option would be to copy the polyfill source code into guzzle I guess, and not require the package.

GrahamCampbell: Unless symfony would be interested in creating a reuseable package both the polyfill and guzzle could depend on that would be at risk of being replaced at composer install.

nicolas-grekas: To support something that dates back from 2011? I would think Guzzle maintainers had better not do this, it would be pure legacy.

I would suggest to not use the polyfill directly.

That would fix everything for people that don't run outdated software. And for those running outdated software, variant 2003 will be used. That's fine. They can upgrade if they have an issue.

nicolas-grekas: > interested in creating a reuseable package both the polyfill and guzzle could depend on

we're not :

stof: > There will always be a case where using the polyfill directly will be needed, so anyway...

but that case would be the pure legacy one (having ext-intl installed with a legacy ICU version). Currently, polyfill internals are also used for the non-legacy case of not having ext-intl.

And for that legacy case, I would also vote for using the variant 2003 rather than using the polyfill internals.

GrahamCampbell: > To support something that dates back to 2011?

Not actually 2011. Some distros from 2019 ship with old versions of the icu library that don't have this constant. This is one of the points of this PR and the older PR this fixes.

jonnott: > > To support something that dates back to 2011?
> 
> Not actually 2011. Some distros from 2019 ship with old versions of the icu library that don't have this constant. This is one of the points of this PR and the older PR this fixes.

For example CentOS 6, which is still supported until 2020-11-30, has ICU 4.2.1.

src/Utils.php
gmponos: > Because we need to be able to call the polyfill even if the original extension is installed to work around the extension using a too old version of the C library. Symfony doesn't have a way for us to do this, other to call code that is formally internal.

I wonder if it would be best to remove the polyfill. Library users can that see a deprecation can set `idn_conversion` to false and use their own custom solutions with a middleware or something like that.

mbabker: That's not always possible.  For me, the trigger on the deprecation notice came from a third party library that uses yet another library which is instantiating its own `Guzzle\Client` instance without any sane way for me to reach it to add middleware or set the `idn_conversion` configuration.

src/Utils.php
stof: But do recent PHP versions trigger a deprecation for the variant 2003 even when it was compiled without support for the uts46 variant ? that looks weird to me.

phpstan-baseline.neon
guilliamxavier: Well that ended up being included in #2638

ahilles107:
Hey, i confirm that with this fix i cannot reproduce issue anymore. It will be merged and released? Or there is another approach for this bug?

Nyholm:
Thank you for all reviews and confirming that this works. 

Im happy with this PR after some small changes. See: https://github.com/GrahamCampbell/guzzle/pull/1

composer.json
```diff
@@ -23,7 +23,7 @@
     "require": {
         "php": ">=5.5",
         "ext-json": "*",
-        "symfony/polyfill-intl-idn": "^1.11",
+        "symfony/polyfill-intl-idn": "1.17.0",
```
nicolas-grekas, line 26: Sorry to chime in after merge: locking to a specific version of the polyfill effectively prevents installing any new(/old) versions ppl might want to use in their app. This is taking control away from users.

I'd recommend using `^1.17`.

I suppose this was done as a measure against using an `@internal` class. But the "risk" this guards against is not strong enough to justify locking all users.

The bigger risk I mentioned in my review is someone using `"replace": "symfony/polyfill-intl"`.
In order to guard from this risk, I'd recommend adding a `class_exists(Idn::class)` before using the class instead.

jonnott: Have to agree with @nicolas-grekas here, locking to a specific version is disastrous in terms of the future, and other libs which might have this same dependency..

GrahamCampbell: We will allow more versions after they are released and we can confirm comparability. Just like someone would write 5.x interested of >=5, we have to bound our versions.

GrahamCampbell: > In order to guard from this risk, I'd recommend adding a class_exists(Idn::class)

That is insufficient. What if the method is renamed, etc. No point in guessing everything possible. If someone has "replaced" the polyfill then this won't matter anyway.

PR #2630, opened 2020-05-01
Idea to replace >>>>>> because also used by git for conflicts

seansan:
Idea to replace >>>>>> because also used by git for conflicts


Nyholm:
Thank you. 

Could you help me understand why this is an issue? Your app will never have conflicts in this file.

Nyholm:
Closing due inactivity.

PR #2635, opened 2020-05-12
Add guzzle version to user agent

jenschude:
With Composer2 it's possible to retrieve the version information of an installed package with no overhead. For older composer versions there is a composer plugin which was originally created by ocramius.

The defaultAgent method now reads the the package version, throws away the git hash and adds the version number to the user agent string.

Maybe the catch should be wider if guzzle hasn't been installed with composer as the class may be not there.

This PR has been inspired by the discussion here https://github.com/guzzle/guzzle/pull/2421/files/1965ac6bdd75ae90ad1b95fdfb8b7ecb9a4c95b7#diff-84b928a9905378810b602c8a740b4cba and to avoid parsing the installed.json file

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,11 +28,11 @@
     "require": {
         "php": "^7.2.5",
         "ext-json": "*",
+        "composer/package-versions-deprecated": "^1.8",
         "guzzlehttp/promises": "^1.0",
         "guzzlehttp/psr7": "^1.6.1",
         "psr/http-client": "^1.0",
-        "symfony/polyfill-intl-idn": "^1.11",
-        "composer/package-versions-deprecated": "^1.8"
+        "symfony/polyfill-intl-idn": "^1.11"
     },
     "provide": {
         "psr/http-client-implementation": "1.0"
--- a/composer.json
+++ b/composer.json
@@ -31,7 +31,8 @@
         "guzzlehttp/promises": "^1.0",
         "guzzlehttp/psr7": "^1.6.1",
         "psr/http-client": "^1.0",
-        "symfony/polyfill-intl-idn": "^1.11"
+        "symfony/polyfill-intl-idn": "^1.11",
+        "composer/package-versions-deprecated": "^1.8"
     },
     "provide": {
         "psr/http-client-implementation": "1.0"
--- a/src/Utils.php
+++ b/src/Utils.php
@@ -6,6 +6,8 @@ use GuzzleHttp\Handler\CurlHandler;
 use GuzzleHttp\Handler\CurlMultiHandler;
 use GuzzleHttp\Handler\Proxy;
 use GuzzleHttp\Handler\StreamHandler;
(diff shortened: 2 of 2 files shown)


GrahamCampbell:
👎 This was already discussed, and it was decided this wasn't wanted, in particular because Guzzle need not be installed using composer.

GrahamCampbell:
> and to avoid parsing the installed.json file

You haven't avoided this. That package you install does exactly that...

jenschude:
> > and to avoid parsing the installed.json file
> 
> You haven't avoided this. That package you install does exactly that...

Not at runtime. Only at installation. But I see the point with the guzzle not being installed using composer.

gmponos:
> Anyhow. A followup PR may fix this.

I do understand that this PR was driven by this statement and thank you for your effort here.

IMHO and this might be a different approach from other maintainers I believe that the less info you expose the better you are... Why would you like to serve to the other side which guzzle version you use.. which PHP version you use or which curl version you use?

Personally when I consume APIs that are external/3rd parties (even if I trust them) I am overriding this header with something like `HttpClient`

Yes.. they might still be able to understand these things from something else.. but I don't think there is any point of sharing those so openly.

with that being said I don't see the need of adding this feature thus discussing one way or the other :)

will keep this open for other maintainers to check and share their thoughts. Thank you again :)

jenschude:
Btw i don‘t mind rejecting the PR.

But i just looked up the user agent format in the respective RFC and after the slash in the product identifier should be a version which is optional. So instead of using GuzzleHttp/Guzzle it would be better to just send GuzzleHttp. See also https://tools.ietf.org/html/rfc7231#section-5.5.3

gmponos:
>But i just looked up the user agent format in the respective RFC and after the slash in the product identifier should be a version which is optional. So instead of using GuzzleHttp/Guzzle it would be better to just send GuzzleHttp. See also https://tools.ietf.org/html/rfc7231#section-5.5.3

That's a valid point.. and also checking this particular code we must also re-discuss the fact that the guzzle is making the assumption that curl handler is used.

jenschude:
This is okay as all other handlers are builtin PHP functions and curl is only added if it‘s available.

gmponos:
My point is that the http client could be initialized with a custom or another handler

```
$stack = new \GuzzleHttp\HandlerStack(new \GuzzleHttp\Handler\StreamHandler());
$client = new GuzzleHttp\Client(
    [
        'handler' => $stack,
    ]
);
```

This will still add that the user agent is CURL although it is wrong.. IMHO the code should not make these assumptions.

jenschude:
Right. And as it‘s a static variable it‘s even harder to debug. Or as said plain wrong if someone instantiates multiple clients with different handlers each.

gmponos:
Closing as @Nyholm provided an alternative and i don't think this will be merged.. thank you for your efforts though

PR #2641, opened 2020-05-14
Change link due repository deprecation

dxops:
- https://github.com/bagder/ca-bundle#deprecated

Do you consider original URL also? https://tls-observatory.services.mozilla.com/api/v1/truststore?store=mozilla&format=pem


gmponos:
Thank you.

PR #2644, opened 2020-05-17
Option to force passing Authorization header on redirect (#1272)

### LINKED ISSUE #1272: 'forward_headers' request option to forward headers when following redirects
I think it would be useful to have a 'forward_headers' request option that would cause Guzzle to forward the request headers specified with the 'headers' option when following redirects.

The use case I have for this at the moment is making a request to a service that requires a secret header to allow me through, but then redirects guzzle to another url, doesn't see the header and then throws a 401 unauthorized error.

For now, i'll use a workaround by disabling redirects and handling the redirect myself manually.

ondrej-bouda:
Resolves #1272 by adding an option to force passing Authorization header on redirects.

### DIFF
--- a/src/RedirectMiddleware.php
+++ b/src/RedirectMiddleware.php
@@ -25,11 +25,12 @@ class RedirectMiddleware
      * @var array
      */
     public static $defaultSettings = [
-        'max'             => 5,
-        'protocols'       => ['http', 'https'],
-        'strict'          => false,
-        'referer'         => false,
-        'track_redirects' => false,
+        'max'                      => 5,
+        'protocols'                => ['http', 'https'],
+        'strict'                   => false,
+        'referer'                  => false,
+        'track_redirects'          => false,
+        'force_pass_authorization' => false,
     ];
 
     /**
@@ -194,7 +195,9 @@ class RedirectMiddleware
         }
 
         // Remove Authorization header if host is different.
-        if ($request->getUri()->getHost() !== $modify['uri']->getHost()) {
+        if (!$options['allow_redirects']['force_pass_authorization'] &&
+            $request->getUri()->getHost() !== $modify['uri']->getHost()
+        ) {
             $modify['remove_headers'][] = 'Authorization';
         }
 
--- a/tests/RedirectMiddlewareTest.php
+++ b/tests/RedirectMiddlewareTest.php
@@ -277,4 +277,26 @@ class RedirectMiddlewareTest extends TestCase
         $client = new Client(['handler' => $handler]);
         $client->get('http://example.com?a=b', ['auth' => ['testuser', 'testpass']]);
     }
+
+    public function testNotRemoveAuthorizationHeaderOnRedirectWhenForced()
+    {
+        $mock = new MockHandler([
+            new Response(302, ['Location' => 'http://test.com']),
+            static function (RequestInterface $request) {
+                self::assertTrue($request->hasHeader('Authorization'));
+                return new Response(200);
+            }
+        ]);
+        $handler = HandlerStack::create($mock);
+        $client = new Client(['handler' => $handler]);
+        $client->get(
+            'http://example.com?a=b',
+            [
+                'auth' => ['testuser', 'testpass'],
+                'allow_redirects' => [
+                    'force_pass_authorization' => true,
+                ],
+            ]
+        );
+    }
 }


gmponos:
Not sure about this feature.. we can discuss this when we come back to 7.1

ondrej-bouda:
No problem, we have to work around the problem anyway in our project as we cannot wait for this feature to be released. Just wanted to help others with the same problems in the future.

As for the reasoning whether to allow such feature, I can see two points in favor of accepting it:
1) Having a secure default and an option to bypass it is quite common (e.g., see Slim enforcing HTTPS with basic auth, [unless explicitly permitted](https://github.com/tuupola/slim-basic-auth#ignore)). We are implementing a system of internal services, all protected from the public by passwords, and we consider no problem letting our client app be redirected from one service to another including credentials. We see no security problems there and the fact some decision logic (same host vs. different host) is hard-wired simply trapped us.
2) I had problems even being aware of the current behaviour and just thought there is a bug in my application. There is no mention anything about it in the documentation and only got to it by debugging. From my point of view, the problem could not be foreseen.

Nyholm:

ondrej-bouda:
No problem, here you go.

src/RedirectMiddleware.php
```diff
@@ -194,7 +195,9 @@ public function modifyRequest(
         }
 
         // Remove Authorization header if host is different.
```
Nyholm, line 197: This comment needs to be updated.

gmponos:
Hi.. 

> @gmponos you were hesitant before. How do you feel now?

I was hesitant because we were about to release the first version of 7. No other reason. Checking the implementation it seems fine.. except that I feel weird for the option name.. @Nyholm do you like that?

src/RedirectMiddleware.php
```diff
@@ -25,11 +25,12 @@ class RedirectMiddleware
...
-        'protocols'       => ['http', 'https'],
-        'strict'          => false,
-        'referer'         => false,
-        'track_redirects' => false,
+        'max'                      => 5,
+        'protocols'                => ['http', 'https'],
+        'strict'                   => false,
+        'referer'                  => false,
+        'track_redirects'          => false,
+        'force_pass_authorization' => false,
```
GrahamCampbell, line 33: I don't like this name. How about `forward_authorization` or even `forward_authorization_on_redirect`. I suppose there's no real reason to have a short string over clarity?

GrahamCampbell (changes requested):

src/RedirectMiddleware.php
Nyholm: 👍  for `forward_authorization_on_redirect`

Nyholm:
@ondrej-bouda Do you mind finishing this PR?

src/RedirectMiddleware.php
GrahamCampbell: I also think this should have 3 possible values. `false`, `null` or `true`. Where `null` passes where the Host is the same, and `false` never passes, and `true` always passes.

Nyholm:
I want to release 7.2.0 in October. I would either like to include this PR in that release or to close it.

sagikazarmark:
I remember turning this down once, mainly for security reasons. It's too easy to send credentials to an unwanted party. It's not exactly the same as using HTTP instead of HTTPS.

Wouldn't simply adding a middleware after the redirect middleware solve this? It would require you to create a new client with that specific middleware, but:

- it's explicit, so you have to know what you are doing
- it keeps the RedirectMiddleware clean

I'm planning on (probably for years now) creating a contrib repo or a place for snippets that helps people find solutions for these edge cases instead of shoving everything into the core (keeps maintaining the core easier).

Nyholm:
Thank you @sagikazarmark for sharing your thoughts. Im closing this PR, I think you are correct.

ondrej-bouda:
I can see your point regarding security. I find it obscure, however, that there is a limitation hard-wired in the core deciding based on the host (even with no mention in the documentation).
In my opinion, the library should not preferably do any unexpected magic. And when it does for a good reason, there should be a way to override it easily. Strictly speaking, it is heuristics, because even same host does not necessarily mean things are more secure. I can understand that the heuristics catches most of the problems correctly, however, as with any heuristics, in my opinion, there should be a way to override it in an easy way.
Of course, contrib repo containing a solution, mentioned in the documentation, would also to the job. Still, I see no point in complicating things for the library users.

sagikazarmark:
> even with no mention in the documentation

That's fair, we should mention it in the docs (PRs welcome 🙂 )

> the library should not preferably do any unexpected magic

I don't think this is unexpected magic. Most HTTP clients strip the authorization header when the host changes. (At least the default Go HTTP client, HTTP clients in Python and C# do that I know of) Those also don't provide a way to pass authorization header when automatic redirection is enabled (and for very good reason).

The solution in most of the cases is to manually follow the redirection. These are edge cases that warrant a custom solution.
You can also just copy the redirect middleware (it's open source and only 200 LOC) and maintain a copied version. You could also get rid of most of the logic if you don't need things like redirect loop protection.

In general, I believe software should be open for extension (and this is actually the case), so you can add your custom logic, but the core does not necessarily have to provide everything by default. I understand that simply setting an option would be more comfortable for you, but from a maintainer perspective a smaller, more stable set of feature set is better than supporting all the edge cases OOTB.

Last, but not least: this is a security question after all, so I'm perfectly happy to take on all the frustration from everyone affected by this issue, if that keeps at least 1 user safe. 🙂

ondrej-bouda:
Fair enough. Thanks for your explanation.

Stelian:
@sagikazarmark I'd like to kindly ask that you reconsider this as it's sadly not an edge case but a quite often occurence, especially when you deal with anything that has open dns (cisco now) behind and they bind sessions to it. In those setups the host always changes for instance for host.com to host.com.hash.opendns.com.

Ideally this would not happen but it does quite a lot on enterprise grade and especially on oauth providers.

sagikazarmark:
@Stelian My first instinct would be just copying the middleware. It's ~200 LOC.

Alternatively, we can probably provider a separate `insecure-redirect-middleware` package.

Stelian:
@sagikazarmark from a package perspective, there are ways to extend it as pointed above but for me the inconvenience comes from the fact that it forces me to create multiple clients when I could reuse the same one (in my case).

A clear example of service dealing with such practices is discord, for example

```
> GET /api/users/@me HTTP/1.1 Host: discord.com User-Agent: GuzzleHttp/7 Authorization: Bearer xxx 
* Mark bundle as not supporting multiuse 
< HTTP/1.1 302 Moved Temporarily 
< Server: Cisco Umbrella 
< Date: Tue, 05 Jan 2021 12:17:13 GMT 
< Content-Type: text/html 
< Content-Length: 192 
< Connection: keep-alive 
< Set-Cookie: X-OpenDNS-Session=xxx; Path=/; Expires=Tue, 05-Jan-21 12:22:13 GMT 
< Location: https://discord.com.x.0fbcd188047c2040f00a1020370751b7e706.9270fd56.id.opendns.com/s/discord.com/api/users/@me?X-OpenDNS-Session=_0fbcd188047c2040f00a1020370751b7e7069270fd56_c7cjzSWy_ 
< Via: HTTP/1.1 a_proxy_fra 
```

Nyholm:
I’m sorry it is “inconvenient” for you to write insecure code. This thread explains why we will not merge this feature and how to work around. 

I’m happy to help you and your application further but then I need to start invoicing you. Feel free to reach out by email.

Stelian:
@Nyholm seems my message came across differently than expected, wasn't related to being incovenient, I was bringing some feedback along. Now having said that, I can help you and your fragile ego on receiving feedback but I need to start invoicing you. Feel free to reach out by email.

sagikazarmark:
Let's keep the conversation civilized.

@Stelian 

> but for me the inconvenience comes from the fact that it forces me to create multiple clients

Why is that? Or rather: how does having this feature in the core make any difference compared to an out-of-tree middleware?

You could just replace the core middleware with your own in your single client. Or am I missing something?

> Now I do understand the security concern as well

Security is not the only argument here: I'm not a huge fan of packing every feature into the core. It makes maintenance of the core harder. Keeping things simple is a good long-term strategy.

It seems that people are usually against copying and maintaining their own versions of middleware/plugins, but I don't know why. That's what I usually do and what I suggest as a first solution. Guzzle is extensible through middleware, so there is no reason to have every imaginable feature in the core.

I understand that you are not the first person who asked for this feature which is why I suggested to host an out of tree version of the middleware under the Guzzle organization.

You asked me to reconsider and I did. I still think this is an edge case (sorry) and that the above choices are the best solution for the project.

Stelian:
@Nyholm apologies for the earlier comment, found yours a bit patronizing and reacted inappropriately myself.

@sagikazarmark as previously mentioned everything you said makes perfect sense and the reasoning for wanting to not have this in core is strong and valid. I was merely highlighting that it's a rather wider spread practice that could take it outside of the realm of the edge-cases at which point an argument can be made for having it in core.

Creating a middleware is straight forward and that is the option I went for, that is not the issue. Clearly it's easier to pass an argument than to build the middleware but if you guys feel like it opens up too much of a vulnerability, I completely understand. My train of thought on the matter was that both the middleware and the config option would lead to the same level of vulnerability in the end so if the consumer is not paying attention it would still be at risk.

Hope this clarifies it.

sagikazarmark:
I'm pretty sure that if we ever opt for maintaining a version that offers this feature, it'll still be opt-in, disabled by default. Given it has security implications, that's what I would suggest to anyone deciding to maintain a copy as well.

GrahamCampbell:
> anyway since discord is unlikely to change their infrastructure

You could report this to them as a security vulnerability. Even if they say it isn't (in the sense that they are not vulnerable), it may make them fix the issue.

Stelian:
@GrahamCampbell I have done that already.

PR #2646, opened 2020-05-18
Set better callable documentation for phpstan

gmponos:


src/Cookie/FileCookieJar.php
```diff
@@ -85,7 +85,7 @@ public function load(string $filename): void
 
         $data = Utils::jsonDecode($json, true);
         if (\is_array($data)) {
-            foreach (Utils::jsonDecode($json, true) as $cookie) {
+            foreach ($data as $cookie) {
```
gmponos, line 88: @GrahamCampbell was there any reason decoding twice the data?

GrahamCampbell: Don't remember. Did I write this???

gmponos: Haven't checked history.. I guessed that you did because you worked on the `utils`.

gmponos: Well you did and you didn't... https://github.com/guzzle/guzzle/pull/2546/files#diff-369479c0c2c035cfd85197bb3a8bfe6fR88

What seems to have happened is that you have only changed the functions.. the problem was there before too..

Issue #2256, opened 2019-01-30
Should the connectionException Message add info with the host?


sandacn:
**Description**  
Now with curl to send a http request, when there is ConnectionErrors(), including (            CURLE_OPERATION_TIMEOUTED ,CURLE_COULDNT_RESOLVE_HOST, CURLE_COULDNT_CONNECT, CURLE_SSL_CONNECT_ERROR, CURLE_GOT_NOTHING)

but the exception with error does not has host info, when we see the default exception we can not clearly know which host is wrong.

**Example**  
for example, the error 7's error message:
```
error: cURL error 7: couldn't connect to host (see http:\/\/curl.haxx.se\/libcurl\/c\/libcurl-errors.html)
```

**Additional context**
now the format is 
```php
$message = sprintf(
            'cURL error %s: %s (%s)',
            $ctx['errno'],
            $ctx['error'],
            'see http://curl.haxx.se/libcurl/c/libcurl-errors.html'
        );
```
I suggest whether we can add info in the (), like this:
```php
$message = sprintf(
            'cURL error %s: %s (%s)',
            $ctx['errno'],
            $ctx['error'],
            'see http://curl.haxx.se/libcurl/c/libcurl-errors.html, while request host is xxxxx'
        );
```
when we the error, we know the host and we can quickly deal with it.

If the RFC is OK, I can implement this feature. 

Thanks .

phansys:
You should take care that some error messages returned by `curl_error()` (which is used to fill `$ctx['error']`) already include the host name, by instance:
>Could not resolve host: nonexistent.tld

gmponos:
@sandacn most probably your example will throw a `ConnectException` which contains the request object.

Why don't you simply get the info from there?

sandacn:
> You should take care that some error messages returned by `curl_error()` (which is used to fill `$ctx['error']`) already include the host name, by instance:
> 
> > Could not resolve host: nonexistent.tld

In my test case, the host does not on the error arrray

sandacn:
> @sandacn most probably your example will throw a `ConnectException` which contains the request object.
> 
> Why don't you simply get the info from there?

I just use error, not exception to catch like this.

gmponos:
> I just use error, not exception to catch like this.

I am not sure what this means...

phansys:
> > You should take care that some error messages returned by `curl_error()` (which is used to fill `$ctx['error']`) already include the host name, by instance:
[...]

PR #2648, opened 2020-05-20
[6.5] Fix "curl error msg for low version"

guilliamxavier:
My try at https://github.com/guzzle/guzzle/pull/2108#discussion_r405509826: revert the incorrect `version_compare` test and append the URI if not already contained in the error message

### DIFF
--- a/src/Handler/CurlFactory.php
+++ b/src/Handler/CurlFactory.php
@@ -15,7 +15,6 @@ use Psr\Http\Message\RequestInterface;
 class CurlFactory implements CurlFactoryInterface
 {
     const CURL_VERSION_STR = 'curl_version';
-    const LOW_CURL_VERSION_NUMBER = '7.21.2';
 
     /** @var array */
     private $handles = [];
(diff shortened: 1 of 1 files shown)


gmponos:
Is this related with this also? https://github.com/guzzle/guzzle/issues/2256

guilliamxavier:
@gmponos: It seems this would fulfill #2256 indeed, as the URI (which includes the host) would be added to the message if not already there <sup>(but maybe better would be to append it unconditionally?)</sup>, whereas currently it is added only if the curl version is *equal to* 7.21.2 (which is a bug, #2108's intention was to add it if the curl version is *less than* 7.21.2 – but I still haven't seen any justification of how "7.21.2" was chosen).

Now you are also right that the request object (from which you can get the URI and/or the host) is passed alongside the message to the exception constructor (either a specific `ConnectException` or a  generic `RequestException`) and you can `try { /* your code */ } catch (RequestException $e) { /* use $e->getMessage() and $e->getRequest() */ }`, but at least two people seemed to wish the URI/host be included in the (logged) error message...

To recap:

1. the current logic is wrong (and to be fixed);
2. open to discussion is when (and if at all) to add the URI <sup>(or just the host?)</sup> to the error message?

gmponos:
> Now you are also right that the request object (from which you can get the URI and/or the host) is passed alongside the message to the exception constructor

Well I was wrong about that.. ErrorHandlers do not have the logic in order to get the URI from the request. So if you let an exception bubble up then when it reaches your ErrorHandler you can not get these info and you need to start patching your error handler.

I guess that this is what the author of the linked issue wanted to [describe here](https://github.com/guzzle/guzzle/issues/2256#issuecomment-571115421) but somehow we got lost in translation :)

guilliamxavier:
@gmponos: Okay :) so what's your opinion on when to add the URI (or just the host?) to the error message?

gmponos:
Haven't reviewed this thoroughly.. reason is this that I am not quite sure which branch it should target...

IIRC You have been into some other issues about targeting v6 or v7 so let me explain you my thinking... 

if the change is very very small.. (doc, changelog, codestyle, phpstan-baseline fix) I am not going to bother asking ppl to retarget to v6 or v7. it seems like a headache to me.. opening two PR. closing the other etc etc...

Every other change I believe it should target v7 except the fixes for the INTL/ICU.. (Although I sould mention this to another issue I am trying to gather all the info and issues we had with intl the last months and get a full summary with solutions).

Nevertheless this is only MY thinking.. so let's wait for other maintainers as well..

gmponos:
Hi. Can you target the v7 and we I will move this to 7.1 milestone

guilliamxavier:
See #2661

PR #2654, opened 2020-05-23
Dont show curl version in user agent

### LINKED ISSUE #2642: Request to an external API using Guzzle randomly returns Server error: 504 but it works using cURL
**Guzzle version(s) affected**: 6.0.2
**PHP version**: 7.0.0
**cURL version**: 7.35.0

**Description**  
I'm giving some maintenance to an old system built on top of Laravel 5.1 (currently providing only bug fixes).

This system needs to make a request to an external API and uses GuzzleHttp (6.0.2) in order to achieve that. This has worked for several years but last week it started to randomly return Server **Error 504** in some cases.

I've contacted the company that developed and maintains the external API and they say on their logs they can see my requests as being **Error 408** (so a client error).

I've made some tests and tried to make the exact same requests using cURL and all worked properly.

**Code using Guzzle**  
`$client = new GuzzleHttp\Client();`
`$response = $client->get('https://example.com/user/1')->getBody();`

With this code, sometimes it manages to make a few requests before it fails (with 504), sometimes it fails immediately on the first attempt.

**Code using cURL**  
`$curl = curl_init();`
`curl_setopt_array($curl, [`
`    CURLOPT_RETURNTRANSFER => 1,`
`    CURLOPT_URL => 'https://example.com/user/1'`
`]);`
`$result = curl_exec($curl);`

With this code, works 100% of the times.

As stated, nothing changed recently. The only maintenance given to this system is related to bug fixes (latest update was almost 2 years ago). Also, the guys who developed the external API also said that nothing changed on their side.

Any hint of what could have caused this problem suddenly?

Thanks in advance

Nyholm:
Don't use any fancy user agent. There is no reason to tell the world about what CURL version you have installed. If a user finds this important, they can add it themselves. 

This will fix #2635, fix #2642

### DIFF
--- a/src/Utils.php
+++ b/src/Utils.php
@@ -103,24 +103,6 @@ final class Utils
     }
 
     /**
-     * Get the default User-Agent string to use with Guzzle.
-     */
-    public static function defaultUserAgent(): string
-    {
-        static $defaultAgent = '';
-
-        if (!$defaultAgent) {
-            $defaultAgent = 'GuzzleHttp/Guzzle';
-            if (\extension_loaded('curl') && \function_exists('curl_version')) {
-                $defaultAgent .= ' curl/' . \curl_version()['version'];
-            }
-            $defaultAgent .= ' PHP/' . PHP_VERSION;
-        }
-
-        return $defaultAgent;
-    }
-
-    /**
      * Returns the default cacert bundle for the current system.
      *
      * First, the openssl.cafile and curl.cainfo php.ini settings are checked.
--- a/src/Utils.php
+++ b/src/Utils.php
@@ -107,7 +107,7 @@ final class Utils
      */
     public static function defaultUserAgent(): string
     {
-        return 'GuzzleHttp/Guzzle';
+        return 'GuzzleHttp';
     }
 
     /**
--- a/src/Utils.php
+++ b/src/Utils.php
@@ -107,7 +107,7 @@ final class Utils
      */
     public static function defaultUserAgent(): string
     {
-        return 'GuzzleHttp';
+        return sprintf('GuzzleHttp/%d', ClientInterface::MAJOR_VERSION);
     }
 
     /**
--- a/src/Utils.php
+++ b/src/Utils.php
@@ -103,6 +103,14 @@ final class Utils
     }
 
     /**
+     * Get the default User-Agent string to use with Guzzle.
+     */
+    public static function defaultUserAgent(): string
+    {
+        return 'GuzzleHttp/Guzzle';
+    }
+
+    /**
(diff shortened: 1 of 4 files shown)


src/Utils.php
```diff
@@ -107,17 +107,7 @@ public static function chooseHandler(): callable
...
-        if (!$defaultAgent) {
-            $defaultAgent = 'GuzzleHttp/Guzzle';
-            if (\extension_loaded('curl') && \function_exists('curl_version')) {
-                $defaultAgent .= ' curl/' . \curl_version()['version'];
-            }
-            $defaultAgent .= ' PHP/' . PHP_VERSION;
-        }
-
-        return $defaultAgent;
+        return 'GuzzleHttp/Guzzle';
```
jenschude, line 110: Should be GuzzleHttp only as the part behind the slash is specified as a version in the corresponding RFC.

Nyholm: It doesn't matter too much, does it?

I named the guzzle 7 to "Guzzle", and now there is a version. =)

jenschude:
Depends as always 😁. The RFC defines the version as a Token. And the agent is a token too. So it may be conform but unexpected as the example show numeric version strings.

But what the RFC clearly states is that the version part itself is optional.

src/Utils.php
jenschude: Depends as always 😁. The RFC defines the version as a Token. And the agent is a token too. So it is may be conform but unexpected as the examples show numeric version strings.

But what the RFC clearly states is that the version part itself is optional.

Nyholm: >But what the RFC clearly states is that the version part itself is optional.

Okey, I've not read the RFC. Im happy to drop this. 

Thank you

GrahamCampbell:
Maybe include the major version from the client interface as the version?

src/Utils.php
```diff
@@ -107,17 +107,7 @@ public static function chooseHandler(): callable
...
-        if (!$defaultAgent) {
-            $defaultAgent = 'GuzzleHttp/Guzzle';
-            if (\extension_loaded('curl') && \function_exists('curl_version')) {
-                $defaultAgent .= ' curl/' . \curl_version()['version'];
-            }
-            $defaultAgent .= ' PHP/' . PHP_VERSION;
-        }
-
-        return $defaultAgent;
+        return 'GuzzleHttp';
```
GrahamCampbell, line 110: `return sprintf('GuzzleHttp/%d', ClientInterface::MAJOR_VERSION);`?

Nyholm: Thank you

PR #2660, opened 2020-05-24
Pass response to TooManyRedirectsException

davidgrayston:
### Fixed
- `TooManyRedirectsException` now has response 

> Note: https://github.com/guzzle/guzzle/pull/2589 includes additional breaking changes to ensure `RequestException` always has a response, which will be considered for version `8.0.0`
>
> If you'd prefer this fix to remain as part of https://github.com/guzzle/guzzle/pull/2589 I'll close this PR

### DIFF
--- a/src/RedirectMiddleware.php
+++ b/src/RedirectMiddleware.php
@@ -88,7 +88,7 @@ class RedirectMiddleware
             return $response;
         }
 
-        $this->guardMax($request, $options);
+        $this->guardMax($request, $response, $options);
         $nextRequest = $this->modifyRequest($request, $options, $response);
 
         if (isset($options['allow_redirects']['on_redirect'])) {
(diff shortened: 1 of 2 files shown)


gmponos:
Hi! I saw the comment from @Tobion .. 

> Looks like TooManyRedirectsException does not have a response but should have.

but I wonder.. when you have too-many-redirects.. it most probably mean you also have many-responses.. so which response out of these should be included to the exception? Why only the last one?

davidgrayston:
The main reason I used the last response is because that's the one available when the exception is thrown. I suppose it gives some context to what's happening, but perhaps it isn't providing much value?

If this causes more confusion that it is helpful, and there's a decision not to make the response mandatory, perhaps this idea should be abandoned? (happy to close both PRs if that's the case?)

gmponos:
I am open to the discussion as well.. let's leave this open.. It's just the reason why I am not 100% sure about this.

Nyholm:
What is the status of this PR?

GrahamCampbell:
I think it just needs to be rebased on master?

Nyholm (approved):
I think this looks good. Thank you

GrahamCampbell (approved):

PR #2662, opened 2020-05-26
Allow running static Actions via Make

rvanlaak:
Having the image version numbers in an environment file somewhere would allow both the Makefile and the Github action to read them.

As a follow-up on this PR I will start working on reducing the size of `phpstan-baseline.neon`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -62,4 +62,12 @@ tag:
 package:
 	php build/packager.php
 
+static: static-phpstan static-cs-fixer
+
+static-phpstan:
+	docker run --rm -it -e REQUIRE_DEV=true -v ${PWD}:/app -w /app oskarstark/phpstan-ga:0.12.23 analyze
+
+static-cs-fixer:
+	docker run --rm -it -v ${PWD}:/app -w /app oskarstark/php-cs-fixer-ga:2.16.3.1 --dry-run --diff-format udiff
+
 .PHONY: docs burgomaster coverage-show view-coverage
--- a/Makefile
+++ b/Makefile
@@ -62,12 +62,18 @@ tag:
 package:
 	php build/packager.php
 
-static: static-phpstan static-cs-fixer
+static: static-phpstan static-codestyle-check
 
 static-phpstan:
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -62,4 +62,12 @@ tag:
 package:
 	php build/packager.php
 
+static: static-phpstan static-cs-fixer
+
+static-phpstan:
+	docker run --rm -it -e REQUIRE_DEV=true -v ${PWD}:/app -w /app oskarstark/phpstan-ga:0.12.23 analyze
```
GrahamCampbell, line 68: can we use 0.12.25?

GrahamCampbell: We should probably add a command to regenerate the baseline file too

Makefile
```diff
@@ -62,4 +62,12 @@ tag:
...
+
+static-cs-fixer:
+	docker run --rm -it -v ${PWD}:/app -w /app oskarstark/php-cs-fixer-ga:2.16.3.1 --dry-run --diff-format udiff
```
GrahamCampbell, line 71: maybe there should be a non-dry-run variant of this too. I know someone can pipe the output to git, but still, the point of a Makefile is that you have a simple command which does the most common use case, which would not be a dry run.

Makefile
rvanlaak: What do you think about somehow centralizing the image version numbers for both the makefile and the Github Action?

rvanlaak: What about delivering the PR first that reduces the baseline errors. Thereafter we can think of incrementing the PHPStan version.

rvanlaak: Did update it to `.27` here as well

Makefile
rvanlaak: Did remove the `--dry-run` parameter, as error code output also will be non-0 when changes were made.

rvanlaak:
@GrahamCampbell added `static-phpstan-update-baseline` make target, and updated PHPStan to the recently released 0.12.28

gmponos:
thank you.

PR #2664, opened 2020-05-26
[RFC] Follow experiment friendly PHP constraint 

rvanlaak:
... as suggested by @nicolas-grekas

After reading all comments on https://github.com/symfony/symfony/pull/36876 I found the arguments listed there sufficient enough to open this PR to also attempt discussing this for Guzzle.

Feel free to close this PR once the discussion is inconclusive.

## Why?

see https://twitter.com/nicolasgrekas/status/1263023258938548225

also see https://github.com/symfony/symfony/pull/36876

> 1. it's either planned obsolescence xor a strong promise to maintain in the long run. None is sustainable.
> 2. if you actually end up maintaining in the long run (not by promise but by fact), your latest versions will work with PHP 8 by definition.
> 3. meanwhile, "^7.x" prevented all your ecosystem from experimenting with PHP 8, which means they increased the workload on you the core maintainer.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,7 @@
         }
     ],
     "require": {
-        "php": "^7.2.5",
+        "php": ">=7.2.5",
         "ext-json": "*",
         "guzzlehttp/promises": "^1.0",
         "guzzlehttp/psr7": "^1.6.1",


GrahamCampbell:
Big 👎 here.

GrahamCampbell:
In Guzzle 7, maybe PHP 8.x will be supported, (in a later release like 7.2.0 or something) but I don't think this will ever be in scope for Guzzle 6. The idea is that everyone should upgrade to Guzzle 7 as soon as possible.

GrahamCampbell:
In any case, Guzzle 7 would implement this by `^7.2 || ^8.0` or something like that. Unbounded version constraints are a terrible idea!

rvanlaak:
As I understand the point of view from the Symfony PR correctly, it is mostly a community-friendly decision that makes experimenting with upcoming releases of PHP easier for other package developers?

Yes, there is a risk that support for the upcoming PHP major is broken before the actual PHP major gets released. IMHO the risk for this to happen with the clear [release schedule of PHP](https://wiki.php.net/todo/php80) (read: no PR for Guzzle that makes CI green between PHP's six `RCs` starting in september, and `GA` in december) is smaller than the upsides of having the flexibility.

Maybe this PR only is mergeable when CI testing on PHP nightly gets added as well?

gmponos:
From here: https://github.com/symfony/symfony/pull/36876#issuecomment-631927489

> We respect all opinions. We listen to all comments. But at some point, we need to make decisions. If we look at the thumbs up and down, we can see that there is no clear consensus on what to do.

> I won't re-iterate the reasons for this change as Nicolas listed them quite well. I could add more reasons, but I will refrain from doing so to not start a controversial conversation.

These words from fabpot fit perfectly here.. I do agree with every word in these lines.. (although votes changed and IMHO they changed clearly to negative).

So yes.. there is no reason to re-iterate if this is needed or not.. they are also written on the PR above and on twitter as well. In the end we need to make a decision. So my vote goes 👎 for this.

Fabpot continues:

> Now, in a few months, if we realize that it was a mistake, we will re-evaluate. Nothing is settled in stone. As a matter of fact, this very strategy was used before Symfony 4.

Again fits perfectly here.. I believe this is not the time for guzzle to take a decision like this and we can come back to that later. I am saying this because I do understand some points of Nicolas but for now I want to live on the safe side.

gmponos:
Also this thread of tweets is very interesting as well.
https://twitter.com/seldaek/status/1263190321992806400?s=20

sagikazarmark:
I think this idea resonates with Minimal Version Selection in Go. The idea is to tell the world which versions you tested your library with, then following semantic versioning you are free to use whatever version you want.

Personally I agree with @nicolas-grekas that in open source there are no guarantees, compatibility must be ensured on the application level. In this case, the alternative is to reject installing a library, which is arguably worse than installing a library that might break.

I understand that this is somewhat "scary" for the PHP community, because it's quite unusual, so I don't think the push back is a surprise. I'm not saying that one or the other is ultimately better, because this feels like an edge case and there are numerous hacks one can use to force installing a library on PHP 8, despite the version constraint.

As a last thought: This might or might not make sense in the Guzzle context. While I like the idea, I'd much rather focus on Guzzle 7 and get back to questions like this later.

AegirLeet:
I don't see how `>=7.2.5` would ever be better than something like `^7.2.5 || ^8.0`. The latter has all the benefits but none of the downsides.

`>=7.2.5` is saying "this will work with any future version of PHP". That kind of statement is impossible to verify and there's a good chance it will turn out to be false at some point in the future. So why not avoid that scenario by being a bit more specific?

pfrenssen:
Here is an alternative approach: https://github.com/guzzle/guzzle/pull/2669

Let's set the version to `^7.2.5||^8` *and actually test on PHP 8 that the code works*.

If we can get green test results on PHP 8 then we can support it moving forward. If not we should not even declare compatibility with it.

creocoder:
Great, lets multiply Symfony dev mistakes!

rvanlaak:
Enough valuable and constructive feedback is collected, @Nyholm @GrahamCampbell @gmponos feel free to close this PR 👍

GrahamCampbell:
This should be closed. Guzzle fundamentally won't work on PHP 8 without changes to types and docs that could be argued as breaking. For example, many things that were resources in PHP 7 are now genuine objects. Things will need to be reviewed properly, possibly only to support PHP 8 in Guzzle 8, and not 7!

Nyholm:
I am generally positive to merging PRs like this. But since the resource issue requires quite a lot of changes, I think it is best to let someone fix that first and then change our php requirements. 

Thank you @rvanlaak

PR #2665, opened 2020-05-26
Change to ArgumentCountError for magic methods

gmponos:
This PR changes the behavior of magic methods trying to simulate in a better form PHP 7.

Before in cases the arguments were missing we used to throw an `Exception`. Now it will throw an `Error`

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -72,7 +72,9 @@ class Client implements ClientInterface, \Psr\Http\Client\ClientInterface
     public function __call($method, $args)
     {
         if (\count($args) < 1) {
-            throw new InvalidArgumentException('Magic request methods require a URI and optional options array');
+            throw new \ArgumentCountError(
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
Hmm, not sure we need this. We only kept the call method for compatibility with 6.x. It will be removed in Guzzle 8. Changing the exception it throws would only be an unnecessary BC break for code we plan to remove and only kept for BC with Guzzle 6.

gmponos:
> Changing the exception it throws would only be an unnecessary BC break for code we plan to remove and only kept for BC with Guzzle 6.

That's a debate I am also having.. and I am willing to hear the thoughs of others as well.. the notion behind this PR is that these kind of "problems" should not be caught easily with a `try... catch(\Exception)`

Nyholm:
Hm, Na, I dont see any benefits with this. 

I dont think we need to give it any love since we want to deprecated it as soon as 7.0 is released.

PR #2666, opened 2020-05-26
Updated editorconfig file

gmponos:


sagikazarmark:
Personally I prefer having a default rule: https://github.com/emperror/errors/blob/master/.editorconfig#L3-L9

gmponos:
I prefer it too.. but it was PHP before.. changed now :)

.editorconfig
```diff
@@ -5,7 +5,16 @@ root = true
...
+[*]
+charset = utf-8
 end_of_line = lf
 indent_style = space
 indent_size = 4
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[*.md]
+trim_trailing_whitespace = false
```
Nyholm, line 17: Why do we need this?

gmponos: So when you first commented I thought "why not to trim on `md` files".. but then I saw that this was false. My bad.. so I updated this..

PR #2667, opened 2020-05-26
Update changelog and upgrading guide

gmponos:
Minor fixes


gmponos:
🤦  need to add also https://github.com/guzzle/guzzle/pull/2629

forgot to add on that PR.

PR #2669, opened 2020-05-27
Declare compatibility with PHP 8

pfrenssen:
Let's see if we have any failures on PHP 8.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,7 +33,7 @@ jobs:
         uses: actions/checkout@v2
 
       - name: Download dependencies
-        run: composer update --no-interaction --prefer-stable --prefer-lowest --prefer-dist
+        run: composer update --no-interaction --prefer-stable --prefer-lowest --prefer-dist --ignore-platform-reqs
 
       - name: Run tests
         run: make test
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
It is much too early to do this. We should review this later in the Guzzle 7.x series.

pfrenssen:
I agree that it is too early, Guzzle 7 would be a good target and we also need to wait until our dependencies are PHP 8 compatible.

I'm posting this as a counterproposal for #2664 - this is an exercise to discover a good practice for supporting PHP 8 early on. I think the proposal in #2664 to set a version constraint like `>=7` is risky and unneeded.

For now I am curious how Guzzle fares on PHP 8. I will close this when I'm done experimenting.

pfrenssen:
Awesome, it seems we are mostly compatible with PHP 8 already. There are a few failures but they are all related to a single line in `guzzlehttp/promises` which throws the following:

```
TypeError: method_exists(): Argument #1 ($object_or_class) must be of type object|string, null given

##[error]/home/runner/work/guzzle/guzzle/vendor/guzzlehttp/promises/src/Promise.php:151
```

So it appears we should wait here until promises is PHP 8 ready :)

GrahamCampbell:
Yeh, there are some known issues with both our promises library and react's one. The will both be fixed in good time. :)

PR #2670, opened 2020-05-28
Fix more issues about phpstan

gmponos:


src/Handler/StreamHandler.php
```diff
@@ -77,7 +80,7 @@ private function invokeStats(
         RequestInterface $request,
         $startTime,
```
GrahamCampbell, line 81: missing type?

gmponos: I believe we can't be sure what this line returns

https://github.com/guzzle/guzzle/blob/master/src/Handler/StreamHandler.php#L36

So I left startTime as `mixed`.. other than that it is done

src/Handler/StreamHandler.php
```diff
@@ -94,6 +94,10 @@ private function invokeStats(
         }
     }
 
+    /**
+     * @param resource $stream
+     * @param mixed $startTime
```
GrahamCampbell: what types can this be?

gmponos: I am resolving this.. 
Please check my previous answer and let's continue the discussion there in order to avoid having the same discussion to multiple places :)

src/Handler/StreamHandler.php
gmponos: I just noticed that on my previous commit I typehinted a different function from the one you commented. Will typehint this one too..

gmponos: OK.. added typehints now to the correct function :)

There are two functions with an argument `$startTime`... both added a typehint `mixed` since the line I mention above returns `float|mixed`

gmponos:
This needs to go to 7.0-RC otherwise lot of ppl will have erros with their PHPStan analysis

PR #2671, opened 2020-05-29
Correct default protocol version

lucasRolff:
Guzzle Client doesn't use HTTP/1.1 by default, but rather 1.0:

https://github.com/guzzle/guzzle/blob/master/src/Handler/CurlFactory.php#L226-L233

### DIFF
--- a/docs/request-options.rst
+++ b/docs/request-options.rst
@@ -1050,10 +1050,10 @@ version
 
 :Summary: Protocol version to use with the request.
 :Types: string, float
-:Default: ``1.1``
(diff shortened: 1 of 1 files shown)


gmponos:
🤔 hi.. this is the part that sets the default version: https://github.com/guzzle/guzzle/blob/master/src/Client.php#L149

lucasRolff:
Sorry 😅I just saw 1.0 being passed on, then it turns out it's a proxy sitting in front and translates 1.1 to 1.0 :(

PR #2672, opened 2020-05-30
Restore ^ to symfony/polyfill-intl-idn version

reedy:


gmponos:
Hi.. there is also this: https://github.com/guzzle/guzzle/pull/2663

Let's keep the discussion there.. and let us know your feedback. 😄

GrahamCampbell:
👎 to this solution, because we are using internal code, and semver breaking changes could be made in patch releases, so fixing to 1.17.0 is the widest version constraint possible.

Nyholm:
I suggested to pin the version. It was not accidental. However, it was a misstake and we should fix it. But as pointed out, just this one line change will not fix the problem.

PR #2673, opened 2020-05-30
Restore ^ to symfony/polyfill-intl-idn version

reedy:


gmponos:
Hi.. there is also this: https://github.com/guzzle/guzzle/pull/2663

Let's keep the discussion there.. and let us know your feedback. 😄

reedy:
Breaking change for me was the removal of the `^`, meant the `replace` directive in composer.json didn't work (using a version of 1.99). We don't need this polyfill as we install php-intl/ext-intl. Sure it's easily fixed by just changing the replace from 1.99 to 1.17.0, but it's not very future proof, and if another library has a different require constraint... It gets awkward :)

GrahamCampbell:
👎 to this solution, because we are using internal code, and semver breaking changes could be made in patch releases, so fixing to 1.17.0 is the widest version constraint possible.

GrahamCampbell:
> We don't need this polyfill as we install php-intl/ext-intl. 

You actually may need the polyfill still, which is the reason for us doing this... even if you have the extension loaded!

Nyholm:
Thank you for this PR. 

https://github.com/guzzle/guzzle/pull/2672#issuecomment-640024468

PR #2674, opened 2020-05-30
Add Slack to the readme

sagikazarmark:


gmponos:
Shall we keep having `gitter` ?

sagikazarmark:
Some people still use it

alexeyshockov:
IMO it's better to remove the Gitter link, because it's used mostly by new people to ask questions, that are usually don't get answered at all, or get answered with a very big delay.

Most of the maintainers are in Slack AFAIK, so this channel should be forced as the only method for interactive communication. For other things there is StackOverflow and GH Issues.

PR #2675, opened 2020-05-31
[7.0] IDN off by default and removed polyfill

GrahamCampbell:


src/Utils.php
```diff
@@ -374,10 +374,10 @@ private static function idnToAsci(string $domain, int $options, ?array &$info =
             return $domain;
         }
```
gmponos, line 375: Do we need the regexe? Since we convert by default to `false` it's the users problem to deal with...

src/Client.php
```diff
@@ -221,7 +221,7 @@ private function configureDefaults(array $config): void
             'decode_content'  => true,
             'verify'          => true,
             'cookies'         => false,
-            'idn_conversion'  => true,
+            'idn_conversion'  => false,
```
gmponos, line 224: Do we need to revert the suggest in composer? 
```
"ext-intl": "Required for Internationalized Domain Name (IDN) support",
```

gmponos:
@alexeyshockov do you mind having a look here?

gmponos:
Thank you.

Nyholm:
Thank you

PR #2677, opened 2020-06-04
Reduce PHPStan baseline errors

rvanlaak:
related to #2662

This PR reduces almost all errors from the PHPStan baseline file, and adds coverage on testing an exception.

No functionality has been changed, and I have been running phpunit locally every time a file was fixed.

@GrahamCampbell I've also addressed your review in #2670 so this PR could replace #2670


src/Handler/EasyHandle.php
```diff
@@ -35,7 +35,7 @@ final class EasyHandle
     /** @var int cURL error number (if any) */
     public $errno = 0;
 
-    /** @var \Throwable Exception during on_headers (if any) */
+    /** @var ?\Throwable Exception during on_headers (if any) */
```
GrahamCampbell, line 38: `Throwable|null`

src/Handler/EasyHandle.php
```diff
@@ -23,7 +23,7 @@ final class EasyHandle
     /** @var array Received HTTP headers so far */
     public $headers = [];
 
-    /** @var ResponseInterface Received response (if any) */
+    /** @var ?ResponseInterface Received response (if any) */
```
GrahamCampbell, line 26: `ResponseInterface|null`

src/Handler/StreamHandler.php
```diff
@@ -75,9 +76,9 @@ public function __invoke(RequestInterface $request, array $options): PromiseInte
     private function invokeStats(
         array $options,
         RequestInterface $request,
-        $startTime,
+        ?float $startTime,
         ResponseInterface $response = null,
-        $error = null
+        ?\Throwable $error = null
```
GrahamCampbell, line 81: Don't need the quesionmark here.

src/Cookie/FileCookieJar.php
```diff
@@ -79,7 +79,8 @@ public function load(string $filename): void
         $json = \file_get_contents($filename);
         if (false === $json) {
             throw new \RuntimeException("Unable to load file {$filename}");
-        } elseif ($json === '') {
+        }
+        if ($json === '') {
```
gmponos, line 83: Let's avoid these kind of changes.. as 6.5 is still active and it will make merging back from 6.5 difficult

src/Exception/BadResponseException.php
```diff
@@ -33,6 +33,9 @@ public function hasResponse(): bool
      */
     public function getResponse(): ResponseInterface
     {
-        return parent::getResponse();
+        /** @var ResponseInterface $response */
+        $response = parent::getResponse();
+
+        return $response;
```
gmponos, line 39: I might say something that might not be accurate in the slightest.. and I am not that familiar with the deep internals of PHP... personally if this has even the slightest performance impact I prefer having a quick return and a baseline error rather than having a clean baseline.

GrahamCampbell: You can just write:

```
/** @var ResponseInterface */
return parent::getResponse();
```

and phpstan will work out what you meant. No need to assign the variable.

src/Cookie/FileCookieJar.php
GrahamCampbell: 6.5 branch is barely active now. I'd be happy to handling merging 6.5 -> 7.x going forward, in any case.

phpstan.neon.dist
```diff
@@ -4,6 +4,8 @@ includes:
 parameters:
     level: max
     checkMissingIterableValueType: false
+    checkGenericClassInNonGenericObjectType: false
+    treatPhpDocTypesAsCertain: false
```
gmponos, line 8: I am not sure.. if this is something we want...

src/Cookie/FileCookieJar.php
```diff
@@ -88,7 +89,7 @@ public function load(string $filename): void
             foreach ($data as $cookie) {
                 $this->setCookie(new SetCookie($cookie));
             }
-        } elseif (\strlen($data)) {
+        } elseif (is_string($data) && \strlen($data)) {
```
gmponos, line 92: I never updated this part for various reasons... what if the json decode above returns an `int` or a `bool` we don't have strict types enabled so this was working so far.

src/Cookie/FileCookieJar.php
gmponos: Fair enough.. and doing a quick pass this might be the only place... we need to get back to the php-cs rules thing.. #2498

src/Cookie/FileCookieJar.php
gmponos: Please if you have other changes like this.. review them again.

src/Handler/MockHandler.php
```diff
@@ -61,8 +61,9 @@ public static function createWithMiddleware(
      * {@see \Psr\Http\Message\ResponseInterface} objects, Exceptions,
      * callables, or Promises.
      *
-     * @param callable|null $onFulfilled Callback to invoke when the return value is fulfilled.
-     * @param callable|null $onRejected  Callback to invoke when the return value is rejected.
+     * @param array<int, mixed> $queue       The parameters to be passed to the append function, as an indexed array.
```
gmponos, line 64: What's the benefit of the typehint of queue here? even if the typehint here stays it needs to be `|null`

src/functions.php
```diff
@@ -4,10 +4,13 @@
...
  * Debug function used to describe the provided value type and class.
  *
+ * @param mixed $input Any type of variable to describe the type of. This
+ *                     parameter misses a typehint because of that.
+ *
  * @return string Returns a string containing the type of the variable and
  *                if a class is provided, the class name.
  */
-function describe_type($input): string
+function describe_type(/* mixed */ $input): string
```
gmponos, line 13: what is this offering here?

src/Handler/MockHandler.php
```diff
@@ -83,8 +84,10 @@ public function __invoke(RequestInterface $request, array $options): PromiseInte
             throw new \OutOfBoundsException('Mock queue is empty');
         }
 
-        if (isset($options['delay']) && \is_numeric($options['delay'])) {
-            \usleep($options['delay'] * 1000);
+        /** @var ?int $delay */
+        $delay = $options['delay'] ?? null;
+        if (\is_numeric($delay)) {
+            \usleep($delay * 1000);
```
gmponos, line 90: well again here.. I prefer having a baseline error rather just have some phpstan optimization for the sake of it...

src/Utils.php
```diff
@@ -288,7 +300,7 @@ public static function jsonDecode(string $json, bool $assoc = false, int $depth
     public static function jsonEncode($value, int $options = 0, int $depth = 512): string
     {
         $json = \json_encode($value, $options, $depth);
-        if (JSON_ERROR_NONE !== \json_last_error()) {
+        if (JSON_ERROR_NONE !== \json_last_error() || false === $json) {
```
gmponos, line 303: IMHO opinion this is a problem of PHPStan

https://github.com/phpstan/phpstan/issues/3317

If I already checked a function that can tell me if `$json`  is false.. why do I need to check `$json` as well..

src/Handler/EasyHandle.php
gmponos: @rvanlaak  there are other places as well.. so pls check them out.

src/Handler/StreamHandler.php
```diff
@@ -304,25 +311,23 @@ private function createStream(RequestInterface $request, array $options)
...
+            static function () use ($context, $params) {
                 return \stream_context_create($context, $params);
             }
         );
 
         return $this->createResource(
             function () use ($uri, &$http_response_header, $context, $options) {
-                $resource = \fopen((string) $uri, 'r', null, $context);
+                /** @var resource $resource */
+                $resource = \fopen((string) $uri, 'r', false, $context);
```
gmponos, line 330: fopen may return false.. this is hiding a possible error here that it would've be better to keep it inside the baseline.

Note this.. moving towards to 7.x we could have addressed these. For instance this specific thing we could introducing a function under `utils` that could either return a resource of throw an exception.. we wanted to reduce as much as possible the changes from 6.x to 7.x (at least we tried our best)

src/Handler/MockHandler.php
GrahamCampbell: Yeh, this is unnecessary.

src/Utils.php
GrahamCampbell: Yep.

rvanlaak: Any idea on how to prevent the PHPStan error from happening? Would it help to write unit tests to cover that `false === $json` assertion?

src/Handler/StreamHandler.php
rvanlaak: Would throwing an exception if `$resource === false` work?

rvanlaak: I think your note about introducing a function in `utils` for `7.x` should get addressed in another PR?

src/Utils.php
gmponos: Do you have a test case that the `\json_last_error()` will return `JSON_ERROR_NONE`  but the json would still be false?

if yes pls post on 3v4l.org/ first...

src/Handler/StreamHandler.php
gmponos: To be fair... on [this PR ](https://github.com/guzzle/guzzle/pull/2670/files#diff-8196fda03bd99e941e07599792a645a4R70) I also introduced a similar change.. so maybe a runtime exception might not be so bad.. either way if it returns false it will fail down the road..

src/functions.php
rvanlaak: fixed in 129a4d86acec990ba533ac6906fd3eb934d95b15

phpstan.neon.dist
rvanlaak: The below errors then would get triggered. But, managed to get those fixed in afef0f3e14e80379b6dea9f0b0560252c4c0bd68 to allow PHPStan to do these type of checks as well.

image

src/Cookie/FileCookieJar.php
rvanlaak: Added unit tests to cover these values in 68f1314d52befd4f6304d368dfef2c9d17b2e244

In other words, you could say that the PHPStan error that was reported here exposed a possible bug.

src/Exception/BadResponseException.php
rvanlaak: Thank you for mentioning this @GrahamCampbell 👏 Agree with not having to define a variable.

I was still living under assumption that PHPStan did not properly support this yet, but they might have changed / improved this in the meanwhile.

Fixed in 734e50e667101f58d560ce93207814a7cb09fe8a

src/Handler/EasyHandle.php
rvanlaak: fixed in 583f8291db35249b94ca8893068002a654caf4a1

src/Handler/EasyHandle.php
rvanlaak: fixed in 583f8291db35249b94ca8893068002a654caf4a1

src/Handler/MockHandler.php
rvanlaak: This defines that `$queue` is an array with `mixed` content, which is an improvement over no typehint.

Added `|null` in ce607dbd0c5df30f85cea13072772b133f7b78dc

In addition I've added descriptive phpdoc to what it's used for, which is copied from the `call_user_func_array` docs.

> The parameters to be passed to the append function, as an indexed array.

src/Handler/MockHandler.php
rvanlaak: reverted in e05fc9539d28e1b7d145079c1e7a7e572fb3e94e

src/Handler/StreamHandler.php
rvanlaak: fixed in 583f8291db35249b94ca8893068002a654caf4a1

phpstan.neon.dist
gmponos: I was talking about `treatPhpDocTypesAsCertain`

src/Utils.php
rvanlaak: Agreed, as encoding `false` will result in `string(5) "false"`

reverted/fixed in 6b3389102af334bec698edeff60513a6dd142444

src/Handler/StreamHandler.php
rvanlaak: Fixed in 0e8f47a44d10535cedaaee819b9c5c1b575aa291

phpstan.neon.dist
gmponos: all the reports above are for `checkGenericClassInNonGenericObjectType` which I am fine setting it to `false`

rvanlaak: Managed to get all errors resolved, so there is no need to change the PHPStan config anymore 👍

gmponos:
Hi.. please don't resolve the discussions as I need to review the changes...

rvanlaak:
My apologies, I'm still working on this PR and am resolving them to have a clear overview myself.

Will unresolve them and re-request for review once I'm done 👍

rvanlaak:
@gmponos @GrahamCampbell thanks for the extensive reviews. I've tried addressing all comments, can you review again?

gmponos:
> Will unresolve them and re-request for review once I'm done

For now don't unresolve al.. just leave open the ones I left...

> I'm still working on this PR

Please also note that a batch of so many change request (**at least for me**) it's not helpful sometimes.. or at least if there are so many changes they need to be cohesive.

For instance this line https://github.com/guzzle/guzzle/pull/2677/commits/e9b9a646d3482c6f08a184ede5e1fecfcebea54c#diff-1ec0793e916ece24b0b0c18df5ebb223R64

or changing to quick returns here https://github.com/guzzle/guzzle/pull/2677/files#diff-392a8065a86ef320fdf3763730ce850eR81

Reviewers will need more time to review your changes and validate that they are correct.

src/RedirectMiddleware.php
```diff
@@ -77,7 +77,7 @@ public function checkRedirect(
         array $options,
         ResponseInterface $response
     ) {
-        if (\substr($response->getStatusCode(), 0, 1) != '3'
+        if (\strpos((string) $response->getStatusCode(), '3') !== 0
```
gmponos, line 80: why is this change needed? 

substr vs strpos

rvanlaak: Micro optimization (`strpos` is roughly 20% faster and consuming less memory) in combination with fixing a PHPStan error.

Feel free to let me know if you want me to revert this change.

src/Cookie/CookieJar.php
```diff
@@ -256,7 +259,7 @@ private function getCookiePathFromRequest(RequestInterface $request): string
         if ('/' === $uriPath) {
             return '/';
         }
-        if (0 === $lastSlashPos = \strrpos($uriPath, '/')) {
+        if (0 === ($lastSlashPos = \strrpos($uriPath, '/')) || false === $lastSlashPos) {
```
Nyholm, line 262:
```suggestion
        $lastSlashPos = \strrpos($uriPath, '/');
        if (0 === $lastSlashPos || false === $lastSlashPos) {
```

src/Cookie/FileCookieJar.php
```diff
@@ -88,7 +89,7 @@ public function load(string $filename): void
             foreach ($data as $cookie) {
                 $this->setCookie(new SetCookie($cookie));
             }
-        } elseif (\strlen($data)) {
+        } elseif (is_scalar($data) && $data != '') {
```
Nyholm, line 92: Is this better?
```suggestion
        } elseif (is_scalar($data) && $data !== '') {
```

src/Utils.php
```diff
@@ -63,10 +68,12 @@ public static function debugResource($value = null)
     {
         if (\is_resource($value)) {
             return $value;
-        } elseif (\defined('STDOUT')) {
+        }
+        if (\defined('STDOUT')) {
             return STDOUT;
         }
 
+        /** @var resource */
```
gmponos, line 76: Can we leave this back to PHPStan as here we are hiding a possible error.

src/Cookie/FileCookieJar.php
gmponos: IMHO I don't see the point of `is_scalar`.. IIRC the `jsoDecode` will return a scalar value. This is a PHPStan issue that should understand that `all` scalar values are returned by `jsonDecode`

src/Cookie/CookieJar.php
gmponos: but isn't this changing the behavior here? 

https://3v4l.org/rZZ4s

src/Handler/CurlMultiHandler.php
```diff
@@ -20,11 +20,30 @@ class CurlMultiHandler
...
+
     /** @var int */
     private $selectTimeout;
-    private $active;
+
+    /** @var resource|null the currently executing resource in `curl_multi_exec`. */
+    private $active = null;
+
+    /**
+     * @var array Request entry handles, indexed by handle id in `addRequest`.
```
gmponos, line 31: Maybe this should be an array of resources ? no?

src/Handler/CurlMultiHandler.php
```diff
@@ -214,6 +242,8 @@ private function timeToNext(): int
             }
         }
 
-        return \max(0, $nextTime - $currentTime) * 1000000;
+        $timeToNext = (int) \max(0, $nextTime - $currentTime);
+
+        return $timeToNext * 1000000;
```
gmponos, line 247: Shall we move this to a one line with a paranthesis?

src/Handler/StreamHandler.php
```diff
@@ -304,27 +311,28 @@ private function createStream(RequestInterface $request, array $options)
         }
 
         // Microsoft NTLM authentication only supported with curl handler
-        if (isset($options['auth'])
-            && \is_array($options['auth'])
-            && isset($options['auth'][2])
-            && 'ntlm' == $options['auth'][2]
+        if (isset($options['auth'][2]) && 'ntlm' === $options['auth'][2]
         ) {
```
gmponos, line 315: This parenthesis should move up.

src/MessageFormatter.php
```diff
@@ -70,7 +70,8 @@ public function format(
     ): string {
         $cache = [];
 
-        return \preg_replace_callback(
+        /** @var string $message */
+        $message = \preg_replace_callback(
```
gmponos, line 74: Can we use the suggestion of @GrahamCampbell with quick return annotation here?

src/Handler/CurlMultiHandler.php
rvanlaak: When looking at `addRequest` it's values aren't resources. `$handles` is an array of arrays.

```
# line 121
$this->addRequest(['easy' => $easy, 'deferred' => $promise]);
```

src/Cookie/FileCookieJar.php
rvanlaak: `jsonDecode` returns an array from perspective of the `FileCookieJar`, several other values are under test coverage.

What is asserted here, is whether any other non-empty value throws the `RuntimeException`. 

I've changed the loose `$data != ''` to `!empty($data)` to clarify that in a826a3af692aef8ebe680584f9a434efdada896d.

src/Utils.php
rvanlaak: Fixed in 8a5a2db52a9e56f5df09d4975ee89817ab7c4ed3

src/Handler/CurlMultiHandler.php
rvanlaak: Fixed in 44148c15357f5601487877f3d8bf8a97a8c4b1af

src/Handler/StreamHandler.php
rvanlaak: Interesting that cs-fixer did not catch this one. Fixed in 44148c15357f5601487877f3d8bf8a97a8c4b1af

src/MessageFormatter.php
rvanlaak: Thanks, missed this one! Fixed in 44148c15357f5601487877f3d8bf8a97a8c4b1af

rvanlaak:
@gmponos @Nyholm thanks for the additional reviews. Branch is rebased on `master` now ✅

src/Cookie/CookieJar.php
```diff
@@ -256,7 +259,8 @@ private function getCookiePathFromRequest(RequestInterface $request): string
         if ('/' === $uriPath) {
             return '/';
         }
-        if (0 === $lastSlashPos = \strrpos($uriPath, '/')) {
+        $lastSlashPos = \strrpos($uriPath, '/');
+        if (0 === $lastSlashPos || false === $lastSlashPos) {
```
gmponos, line 263: but isn't this changing the behavior here?

https://3v4l.org/rZZ4s


related comment: https://github.com/guzzle/guzzle/pull/2677#discussion_r436259529

src/Handler/CurlMultiHandler.php
gmponos: true.. sorry..

src/Handler/StreamHandler.php
```diff
@@ -61,9 +61,9 @@ public function __invoke(RequestInterface $request, array $options): PromiseInte
             $message = $e->getMessage();
             // This list can probably get more comprehensive.
             if (\strpos($message, 'getaddrinfo') // DNS lookup failed
```
gmponos, line 63: Shouldn't we add the `!== false` to the line here?

src/Middleware.php
```diff
@@ -72,7 +72,7 @@ function (ResponseInterface $response) use ($request) {
     /**
      * Middleware that pushes history data to an ArrayAccess container.
      *
-     * @param array|\ArrayAccess $container Container to hold the history (by reference).
+     * @param null|array|\ArrayAccess<int, array> $container Container to hold the history (by reference).
```
gmponos, line 75: Why is this allowed null?

src/Utils.php
```diff
@@ -300,13 +315,13 @@ public static function jsonEncode($value, int $options = 0, int $depth = 512): s
      * Wrapper for the hrtime() or microtime() functions
      * (depending on the PHP version, one of the two is used)
      *
-     * @return float|mixed UNIX timestamp
+     * @return float UNIX timestamp
      *
      * @internal
      */
     public static function currentTime()
```
gmponos, line 322: if we are sure that this returns a float.. we could remove the docblock and add a return type.. 

or leave the docblock but I don't see much value to `@return float UNIX timestamp` instead I would prefer if it could be moved to description.

rvanlaak: Adding the return type on a public method would be a BC break

gmponos: yes.. but we are moving from 6 towards 7.. and v7 is still as beta.. so we are allowed to do so.

gmponos: also this is a function marked with `@internal`.. which means that it is not under our BC policy.

src/Cookie/SetCookie.php
```diff
@@ -207,6 +208,8 @@ public function setMaxAge($maxAge): void
 
     /**
      * The UNIX timestamp when the cookie Expires
+     *
+     * @return null|string|int
```
gmponos, line 212: According to [this](https://github.com/guzzle/guzzle/pull/2677/files#diff-6456b4cf27d0bc8bc517e1e5d36e0618R226)

```
 $this->data['Expires'] = \is_numeric($timestamp)
            ? (int) $timestamp
            : \strtotime($timestamp);
```

then most probably the `expires` key will always be an int right?

src/Utils.php
```diff
@@ -220,7 +231,10 @@ public static function isHostInNoProxy(string $host, array $noProxyArray): bool
 
         // Strip port if present.
         if (\strpos($host, ':')) {
-            $host = \explode($host, ':', 2)[0];
+            $hostParts = \explode($host, ':', 2);
+            if (false !== $hostParts) {
+                $host = $hostParts[0];
+            }
```
gmponos, line 237: So I was trying to have a real good look on this.. because I was really confused in the way that the code (not yours) is written.. 

So after examining this I believe that `explode` function will never throw false.. as we have already checked that the string contains the `:` and it is able to explode. 

I would be happy to live here with a baseline error or a docblock...

src/Cookie/CookieJar.php
rvanlaak: Good one. I've spun up XDebug to verify your assumption. `CookieJarTest::testCookiePathWithEmptySetCookiePath` seems to properly cover this.

Line 253 lets the method return early in case there isn't any `/` in `$uriPath`, so `$lastSlashPos` will never be `false`:

```
if (0 !== \strpos($uriPath, '/')) {
```

src/Cookie/SetCookie.php
rvanlaak: Theoretically, no. Practically (after construction), yet.

* `__construct` does use both the getter and setter 
* `$defaults` defines `Expires` as `null`.

I eventually removed `int` as return typehint. 

Not calling the getter / setter in the constructor would allow removing `null`. Let me know if you want me to do that, I'll then create another PR for that.

src/Utils.php
rvanlaak: You are right, and in addition to that from the `explode` documentation:

> If delimiter is an empty string (""), explode will return false.

An `InvalidArgumentException` is thrown on top of the explode, so no this won't happen. 

Fixed in c5687a9d8f67e113e92dd1490c3f51ccf65d5bc6

src/Utils.php
rvanlaak: Great! Fixed in 7f77c9ec0be33951e751f0a3394f908c6aafd880

Planning on adding more typehints? This PR focuses on fixing the PHPStan errors, so I'd be happy to help on adding more typehints / code safety in another PR.

src/Middleware.php
rvanlaak: You are right, `null` is not a allowed value as that will cause an `InvalidArgumentException`.

Fixed in 969a11de1411491f0a02c5f1e722fa0d0fce7736

src/Handler/StreamHandler.php
rvanlaak: fixed in 5326569762e076fc8e5580277fcda355ee13a68b

rvanlaak:
Thank for your extensive review again 👍 Did analyze all your assumptions and made the required changes accordingly.

src/Middleware.php
gmponos: but why did we move the docblock inline below?

gmponos:
> Planning on adding more typehints?

There are places that for X, Y reasons we were not sure about what typehint to add. If you want to help there with a follow up PR that would be great.

src/Cookie/FileCookieJar.php
```diff
@@ -88,7 +89,7 @@ public function load(string $filename): void
             foreach ($data as $cookie) {
                 $this->setCookie(new SetCookie($cookie));
             }
-        } elseif (\strlen($data)) {
+        } elseif (is_scalar($data) && !empty($data)) {
```
GrahamCampbell, line 92: Missing leading slash.

src/Handler/CurlMultiHandler.php
```diff
@@ -20,11 +20,30 @@ class CurlMultiHandler
 {
     /** @var CurlFactoryInterface */
     private $factory;
+
     /** @var int */
     private $selectTimeout;
-    private $active;
+
+    /** @var resource|null the currently executing resource in `curl_multi_exec`. */
+    private $active = null;
```
GrahamCampbell, line 28: Don't need to initialise this. Already initialises to null?

src/Handler/CurlMultiHandler.php
```diff
@@ -49,25 +68,34 @@ public function __construct(array $options = [])
...
 
-        $this->options = isset($options['options']) ? $options['options'] : [];
+        $this->options = $options['options'] ?? [];
     }
 
+    /**
+     * @param string $name
+     *
+     * @return resource
+     */
```
GrahamCampbell, line 78: Missing throws doc? The other `__get` lower down the diff has throws doc.

src/Handler/CurlMultiHandler.php
```diff
@@ -49,25 +68,34 @@ public function __construct(array $options = [])
...
+        }
+
+        $this->_mh = $multiHandle;
 
-            // Further calls to _mh will return the value directly, without entering the
-            // __get() method at all.
-            return $this->_mh;
+        foreach ($this->options as $option => $value) {
+            // A warning is raised in case of a wrong option.
+            curl_multi_setopt($this->_mh, $option, $value);
```
GrahamCampbell, line 95: Does this intentionally not have a leading slash for mocking?

src/Handler/StreamHandler.php
```diff
@@ -317,27 +313,27 @@ private function createStream(RequestInterface $request, array $options)
...
 
         return $this->createResource(
-            function () use ($uri, &$http_response_header, $context, $options) {
-                $resource = \fopen((string) $uri, 'r', null, $context);
+            function () use ($uri, &$http_response_header, $contextResource, $context, $options, $request) {
+                $resource = \fopen((string) $uri, 'r', false, $contextResource);
                 $this->lastHeaders = $http_response_header;
 
+                if (false === $resource) {
+                    throw new ConnectException('Connection refused for URI ' . $uri, $request, null, $context);
```
GrahamCampbell, line 334: sprintf should be used here?

src/HandlerStack.php
```diff
@@ -252,6 +252,6 @@ private function debugCallable($fn): string
                 : "callable(['" . \get_class($fn[0]) . "', '{$fn[1]}'])";
         }
 
-        return 'callable(' . \spl_object_hash($fn) . ')';
+        return 'callable(' . \spl_object_hash((object) $fn) . ')';
```
GrahamCampbell, line 255: What codepath causes this cast to be needed?

src/Utils.php
```diff
@@ -286,6 +299,7 @@ public static function jsonDecode(string $json, bool $assoc = false, int $depth
      */
     public static function jsonEncode($value, int $options = 0, int $depth = 512): string
     {
+        /** @var string $json */
```
GrahamCampbell, line 302: This could actually be false. We only know it is a string after the guard has been checked on the lower line. This doc should be moved to just above the return statement.

src/Utils.php
```diff
@@ -300,13 +314,13 @@ public static function jsonEncode($value, int $options = 0, int $depth = 512): s
...
-     * @return float|mixed UNIX timestamp
+     * @return float UNIX timestamp
      *
      * @internal
      */
-    public static function currentTime()
+    public static function currentTime(): float
     {
-        return \function_exists('hrtime') ? \hrtime(true) / 1e9 : \microtime(true);
+        return (float) \function_exists('hrtime') ? \hrtime(true) / 1e9 : \microtime(true);
```
GrahamCampbell, line 323: When is this not a float?

src/Middleware.php
rvanlaak: This change fixes a PHPStan error stating that `!\is_array($container) && !$container instanceof \ArrayAccess` can never be `false`.

As there is no typehint, the parameter's value *can* be null. The reason for putting the docblock inline is to not leak that information to outside of the named `history` constructor's method signature.

Let me know how to otherwise address this 👍

gmponos: > As there is no typehint, the parameter's value can be null.

Well it can be anything else as well.. not only null.. I believe leaving the phpstan error back is fine..

src/HandlerStack.php
gmponos: 👍 I had this in mind.. forgot to comment about it..

src/Handler/CurlMultiHandler.php
rvanlaak: I remember making a change in a PHP7.4 environment, where "uninitialized exceptions" were thrown. See: https://stackoverflow.com/questions/60112627/how-to-let-phpunit-test-property-initialization-in-php7-4

Nikic recommends initializing as `null` when the constructor doesn't write a field's value.

GrahamCampbell: That only is relevant with typed properties.

src/Handler/CurlMultiHandler.php
rvanlaak: `curl_multi_setopt` gets overwritten in `tests/bootstrap.php`. Adding the leading slash makes `\GuzzleHttp\Tests\Handler\CurlMultiHandlerTest::testCanAddCustomCurlOptions` fail.

Maybe make a note somewhere to change this test mock?

src/HandlerStack.php
rvanlaak: `spl_object_hash` expects an object to create a hash out of it, so my assumption given that this is related to a debug method was that the safest bet here would be to try reaching that with a typecast.

I've analyzed all `Middleware::$stack` writes, and we can be safe its items are covered in `\GuzzleHttp\Tests\HandlerStackTest::testCanPrintMiddleware`. The only usages of `debugMessage` are in `__toString`.

gmponos: @rvanlaak not sure what's your conclusion here? I am still in favor of removing the typecast (even after your comment) since I consider it safe.

https://3v4l.org/EL7ZV

https://3v4l.org/EedLd

check the OP codes as well.

src/Utils.php
rvanlaak: [`hrtime` documentation:](https://www.php.net/manual/en/function.hrtime.php) 

> Returns an array of integers in the form [seconds, nanoseconds], if the parameter get_as_number is false. Otherwise the nanoseconds are returned as integer (64bit platforms) or float (32bit platforms).

So, returns non-float when hrtime exists and on a 64 bit platform

src/Utils.php
rvanlaak: Good point. Fixed in 1cb4840a6dda807ea7a86d1785950fc10f2a7a36

src/Handler/StreamHandler.php
rvanlaak: Fixed in 9eb2f5ab6d17224c9b83c311a81a79311f847a10

src/Handler/CurlMultiHandler.php
rvanlaak: Fixed in 9ef1da2e03cbbe60b29e23b4114b08f12cd8b5b0

src/Cookie/FileCookieJar.php
rvanlaak: Wouldn't this be a matter of configuring CS-Fixer to prevent future review efforts? Or what about importing these methods?

Fixed in 3f37ada3b1af0d7d027b165cc494c5269c3da9e3

src/HandlerStack.php
rvanlaak: Ok As in, replace it?

```
/** @var object $fn */
```

Fixed in 2a5e4d5de1df05c96cb4b4b80ef2fe1be59b27db

rvanlaak:
@GrahamCampbell thank you for your review 👍 Did make the related changes

src/Utils.php
GrahamCampbell: Cool

GrahamCampbell:
I've sent a PR. Let's see what happens: https://github.com/rvanlaak/guzzle/pull/1.

src/HandlerStack.php
gmponos: Hi.. I am afraid this is not sufficient.. since it can be a callback... 

just a note here.. and as a feedback.. note that PHPStan is just a tool.. there is no need to address EVERY baseline error even in the wrong way just in order not to have baseline errors.

same goes for this as well https://github.com/guzzle/guzzle/pull/2677#discussion_r436955245

GrahamCampbell: Can array-syntax callbacks actually be cast to an object? I think both variants of the code are broken? Do we need to make a closure from the callable first?

rvanlaak:
@GrahamCampbell merged your PR, thanks! 

Also had checking that on my list when I got the message 0.12.27 was released Yesterday 🎉 https://github.com/phpstan/phpstan/releases/tag/0.12.27

gmponos:
if possible to address this as well.. https://github.com/guzzle/guzzle/pull/2677#discussion_r436955245

rvanlaak:
@GrahamCampbell the object cast was fixed in 2a5e4d5

@gmponos did fix your suggestion in 1903a61ff86dd195b5c762426da161e8be57ca0f

src/HandlerStack.php
rvanlaak: array-syntax callbacks get covered in the lines above:

```php
        if (\is_array($fn)) {
            return \is_string($fn[0])
                ? "callable({$fn[0]}::{$fn[1]})"
                : "callable(['" . \get_class($fn[0]) . "', '{$fn[1]}'])";
        }
```

rvanlaak:
@GrahamCampbell thanks! Squashed and merged ✅

gmponos:
thank you so much for this and all the follow ups and patience you had :)

GrahamCampbell:
🎉

rvanlaak:
Great! Thanks for the extensive reviews!

Nyholm:
Great work. Thank you for the reviews.

PR #2682, opened 2020-06-09
Finalize php-cs-fixer changes

gmponos:
All changes are made with using php-cs-fixer tool


src/Handler/EasyHandle.php
```diff
@@ -88,11 +105,9 @@ public function createResponse(): void
     /**
      * @param string $name
      *
-     * @return void
-     *
      * @throws \BadMethodCallException
      */
-    public function __get($name)
+    public function __get($name): void
```
GrahamCampbell, line 110: do we actually want this?

.php_cs.dist
```diff
@@ -5,28 +5,74 @@ $config = PhpCsFixer\Config::create()
...
         'array_syntax' => ['syntax' => 'short'],
+        'blank_line_after_opening_tag' => true,
+        'compact_nullable_typehint' => true,
+        'concat_space' => ['spacing' => 'one'],
+        'declare_equal_normalize' => ['space' => 'none'],
         'declare_strict_types' => false,
+        'dir_constant' => true,
+        'final_static_access' => true,
+        'fully_qualified_strict_types' => true,
         'function_to_constant' => ['functions' => ['php_sapi_name']],
```
GrahamCampbell, line 16: do we want to add things like `php_version` in here?

src/Handler/EasyHandle.php
gmponos: Hm.. this will indeed be problematic for ppl that extend this class.. and it has less benefit indeed. Reverted

.php_cs.dist
gmponos: used true instead.

src/Handler/CurlFactory.php
```diff
@@ -453,34 +458,34 @@ private function applyHandlerOptions(EasyHandle $easy, array &$conf): void
...
-                    list($sslKey, $conf[CURLOPT_SSLKEYPASSWD]) = $options['ssl_key'];
+                    [$sslKey, $conf[\CURLOPT_SSLKEYPASSWD]] = $options['ssl_key'];
                 } else {
-                    list($sslKey) = $options['ssl_key'];
+                    [$sslKey] = $options['ssl_key'];
                 }
             }
 
-            $sslKey = isset($sslKey) ? $sslKey: $options['ssl_key'];
+            $sslKey = $sslKey?? $options['ssl_key'];
```
guilliamxavier, line 481: curious, why didn't the fixer ensure a space between `$sslKey` and `??` here?

guilliamxavier: oh, the `binary_operator_spaces` rule is not enabled by the `@PSR2` ruleset...

gmponos: Do you mind submitting a followup PR?

guilliamxavier: @gmponos: done: #2688

PR #2683, opened 2020-06-12
replace allow_redirects to \GuzzleHttp\RequestOptions::ALLOW_REDIRECTS

dypa:
\GuzzleHttp\RequestOptions class contains 30 constants, but in project code still use string literals instead of constants. I decided to try replace allow_redirects with a constant to see reaction of project maintainers. If PR will be merged i want to continue replace other strings to constants.

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -25,7 +25,7 @@ class Client implements ClientInterface, \Psr\Http\Client\ClientInterface
      *     $client = new Client([
      *         'base_uri'        => 'http://www.foo.com/1.0/',
      *         'timeout'         => 0,
-     *         'allow_redirects' => false,
+     *         \GuzzleHttp\RequestOptions::ALLOW_REDIRECTS => false,
      *         'proxy'           => '192.168.16.1:10'
      *     ]);
      *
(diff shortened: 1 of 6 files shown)


Nyholm:
Thank you for the PR. 

I fail to see the benefits of this.

mfn:
I'm not sure there benefits for users of the library, but one benefit of the constants is for IDEs which allow to "find usage" and za-BANG you immediately know where it's used; not accurate/possibility for ambiguity with simple text searches.

If I were to got into this project to contribute and, having this constants, it would be easier for me to "follow them" in the code.

Just my 2c, I see the value.

dypa:
@Nyholm i see such advantages from this PR
* constant \GuzzleHttp\RequestOptions::{HTTP_ERRORS, SYNCHRONOUS, STREAM, READ_TIMEOUT} is already used in the code
* constant \GuzzleHttp\RequestOptions::TIMEOUT allows me to distinguish stream context timeout from guzzle timeout for example \GuzzleHttp\Handler\StreamHandler::add_timeout
* simplification of code navigation in the IDE
* PR does not create changes for library users, no BC changes

Nyholm:
We are currently using some magic strings and some constants. 
This PR does not make us consistent nor does it help readability. Changes like this fix 1 our of 4 magic strings. 

Screenshot 2020-06-27 at 14 07 42

Im open for a discussion about removing all magic strings, but a PR like this do not make sense to me. 

Thank you for your time and effort.

PR #2685, opened 2020-06-13
Fail on warning/risky

gmponos:


gmponos:
https://github.com/guzzle/guzzle/pull/2685/files#diff-98c12a693941ae3c4f50872403416471L19

Shall we keep this? or maybe add coverage options there?

gmponos:
> Shall we keep this? or maybe add coverage options there?

Yea after all we should.. I just saw that we do have a make command to generate coverage.

PR #2690, opened 2020-06-14
Remove call_user_func

gmponos:
Based on 

- https://github.com/fab2s/call_user_func (@fab2s)
- https://github.com/symfony/symfony/pull/29309


GrahamCampbell:
Note that this will break if the array contains string keys.

gmponos:
@GrahamCampbell I guess your comment was about the `call_user_func_array` 

I reverted those.. nice catch.. 👍

GrahamCampbell:
In 8.x, we could add extra typing requirements to insist on arrays are are lists.

PR #2691, opened 2020-06-14
Fix an issue where cookies with an empty string were not stored.

peckrob:
This is a re-creation of PR #1991, with the requested fix in it and based off master.

Cookies with empty string values are technically a violation of RFC 6265, but this issue is known to occur in the wild (I personally encountered it in dealing with a SAP application server). Browsers handle it property.

Test was also fixed an an additional test for empty string was added.

### DIFF
--- a/src/Cookie/SetCookie.php
+++ b/src/Cookie/SetCookie.php
@@ -392,9 +392,10 @@ class SetCookie
                 . 'following characters: ()<>@,;:\"/?={}';
         }
 
-        // Value must not be empty, but can be 0
+        // Value must not be null. 0 and empty string are valid. Empty strings
+        // are technically against RFC 6265, but known to happen in the wild.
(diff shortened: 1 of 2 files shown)


src/Cookie/SetCookie.php
```diff
@@ -392,9 +392,10 @@ public function validate()
                 . 'following characters: ()<>@,;:\"/?={}';
         }
 
-        // Value must not be empty, but can be 0
+        // Value must not be null. 0 and empty string are valid. Empty strings
```
GrahamCampbell, line 395: These comments about zero are kinda stupid (not your fault). They must have been added after someone used empty by mistake instead of comparing against the empty string...

Nyholm: I agree. But Im fine with keeping them.

PR #2699, opened 2020-06-27
[7.0] Fix multiply defined functions fatal error

GrahamCampbell:
I think we need to cut a 7.0 branch for this, and release as 7.0.1 ASAP.


GrahamCampbell:
The reason nobody reported this error is because it will only happen if guzzle 7 is installed globally at the same time... but likely everyone testing guzzle 7 betas only installed it locally, and had guzzle 6 globally if at all.

PR #2562, opened 2020-01-18
[7.1] Deprecate Utils::defaultCaBundle

GrahamCampbell:
Requires https://github.com/guzzle/guzzle/pull/2561 to be merged first. Is a follow up to https://github.com/guzzle/guzzle/pull/2560.

PR #2701, opened 2020-06-27
[8.0] Remove Utils::defaultCaBundle

GrahamCampbell:
Depends on #2700. Follow-up to #2562.


GrahamCampbell:

PR #2704, opened 2020-06-27
Minor CI cleanup

Nyholm:
This will speed up the test a little

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,7 +18,7 @@ jobs:
           php-version: '7.2'
           coverage: none
           extensions: mbstring, intl
-          tools: flex
+          tools: composer:v2
 
       - name: Set up Node
         uses: actions/setup-node@v1
@@ -52,7 +52,7 @@ jobs:
           php-version: ${{ matrix.php }}
           coverage: none
           extensions: mbstring, intl
-          tools: flex
+          tools: composer:v2
 
       - name: Set up Node
         uses: actions/setup-node@v1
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Can we use composer 2 instead of flex?

GrahamCampbell:
Just like https://github.com/laravel/framework/pull/33158.

Nyholm:
Yes, of course. I was not aware of that setting.

GrahamCampbell:
> I was not aware of that setting.

It's quite new. I asked them to add it. :)

.github/workflows/ci.yml
```diff
@@ -13,13 +13,12 @@ jobs:
 
     steps:
       - name: Set up PHP
-        uses: shivammathur/setup-php@1.7.0
+        uses: shivammathur/setup-php@2.3.1
```
GrahamCampbell, line 16: Could we make this just `@v2` like we did for checkout and node? I don't think we need to pin?

gmponos:
Thank you

Nyholm:
Great. Thank you for the reviews

.github/workflows/ci.yml
Nyholm: Actually, that is not the same.. 

Github actions maintainers are suggested to update the tag. 

So `shivammathur/setup-php@2.3.1` means "get version 2.3.1. 
And `shivammathur/setup-php@v2` means "get version v2. 

It is nothing like composer. If maintainers of `shivammathur/setup-php` does not update (move) the tag, then we are stuck with an old version... Just like now.. 

With this we know what version we are using. If a future version includes stuff that we like, then we just change the version here.

GrahamCampbell: v2 does actually seem to get 2.3.1, otherwise the composer:v2 thing wouldn't work.

PR #2705, opened 2020-06-27
[7.1] Tweaked CI

GrahamCampbell:


.github/workflows/ci.yml
```diff
@@ -13,7 +13,7 @@ jobs:
 
     steps:
       - name: Set up PHP
-        uses: shivammathur/setup-php@2.3.1
```
Nyholm, line 16: This is the only change Im not sure about.

GrahamCampbell: They seem to have been reliably moving the v2 tag, just like the checkout code action.

image

Nyholm:
Thank you

Issue #2603, opened 2020-03-22
Add psalm


vladyslavstartsev:
**Description**
Adding similar issue to http client repo as this https://github.com/guzzle/psr7/issues/324
tl;dr
add psalm as static analysis tool to the project

gmponos:
I am in the same page as @Tobion on this [comment](https://github.com/guzzle/psr7/issues/324#issuecomment-628355418).

Moreover I believe that there is no actual benefit and it will cause more headaches to developers that want to contribute. Let's focus on resolving the current PHPStan issues and the we can re-discuss about psalm.

vladyslavstartsev:
:+1:

PR #2706, opened 2020-06-27
Adding Psalm

Nyholm:
I added a baseline with the current errors. The errorLevel 3 was suggested by `psalm --init`

### DIFF
--- a/.github/workflows/static.yml
+++ b/.github/workflows/static.yml
@@ -51,7 +51,7 @@ jobs:
 
       - name: Download dependencies
         run: |
-          composer require --dev --no-updatepsalm/phar:3.12.1
+          composer require --dev --no-update psalm/phar:3.12.1
           composer update --no-interaction --prefer-dist
 
       - name: Psalm
--- a/.github/workflows/static.yml
+++ b/.github/workflows/static.yml
@@ -50,9 +50,11 @@ jobs:
         uses: actions/checkout@v2
 
       - name: Download dependencies
-        run: composer update --no-interaction --prefer-dist
+        run: |
+          composer require --dev --no-updatepsalm/phar:3.12.1
+          composer update --no-interaction --prefer-dist
 
       - name: Psalm
-        uses: psalm/psalm-github-actions@master
-        with:
-          args: --no-progress --show-info=false --stats
+        run: |
+          ./vendor/bin/psalm --version
+          ./vendor/bin/psalm --output-format=github --no-progress --show-info=false --stats
--- a/.github/workflows/static.yml
+++ b/.github/workflows/static.yml
@@ -39,8 +39,18 @@ jobs:
     name: Psalm
     runs-on: ubuntu-latest
     steps:
+      - name: Set up PHP
+        uses: shivammathur/setup-php@2.3.1
+        with:
+          php-version: '7.4'
+          extensions: mbstring, intl
+          tools: composer:v2
+
       - name: Checkout code
         uses: actions/checkout@v2
+
+      - name: Download dependencies
+        run: composer update --no-interaction --prefer-dist
 
       - name: Psalm
         uses: psalm/psalm-github-actions@master
--- a/.github/workflows/static.yml
+++ b/.github/workflows/static.yml
@@ -56,5 +56,5 @@ jobs:
 
       - name: Psalm
         run: |
-          ./vendor/bin/psalm --version
-          ./vendor/bin/psalm --output-format=github --no-progress --show-info=false --stats
+          ./vendor/bin/psalm.phar --version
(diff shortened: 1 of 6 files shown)


.github/workflows/static.yml
```diff
@@ -34,3 +34,15 @@ jobs:
...
+
+  psalm:
+    name: Psalm
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Psalm
+        uses: psalm/psalm-github-actions@master
```
GrahamCampbell, line 46: This is unfortunate that we can't pin. I'd actually rather we just loaded PHP then downloaded a phar, so we can pin the tool version, if upstream don't want to do tags.

gmponos:
FYI https://github.com/guzzle/guzzle/issues/2603#issuecomment-629826442

.github/workflows/static.yml
```diff
@@ -50,9 +50,11 @@ jobs:
         uses: actions/checkout@v2
 
       - name: Download dependencies
-        run: composer update --no-interaction --prefer-dist
+        run: |
+          composer require --dev --no-updatepsalm/phar:3.12.1
```
GrahamCampbell: I don't think we need this line if it's going to be in the committed composer.json. Also, if it is, why not put phpstan in there too?

Nyholm:
I disagree with @Tobion here. 

Psalm and PHPStan are doing the same thing. PHPStan is a bit more friendly and easy to use locally. But Psalm has way more features and discovers way more things (Just look at our psalm baseline file). Psalm also integrates with github reviews better, it actually makes a review comment on the line with issues. 

So, yes, Two static analysers may be too much. I dont mind using two, but if we should remove one, then PHPStan is the one to go.

.github/workflows/static.yml
Nyholm: Oh, sorry. I was just testing locally. It should not be in composer.json. 

Thank you

Nyholm:
The PR is ready for a review

Makefile
```diff
@@ -63,7 +63,10 @@ tag:
 	git commit -m '$(TAG) release'
 	chag tag
 
-static: static-phpstan static-codestyle-check
+static: static-phpstan static-psalm static-codestyle-check
+
+static-psalm:
+	docker run --rm -it -v ${PWD}:/app -w /app vimeo/psalm-github-actions
```
GrahamCampbell, line 69: will produce different results to using the phar. can we have the makefile download the phar to /tmp or something?

Nyholm: Yes, it might. But in this case it is a good thing, right? 

If the output is different then we need to be more defensive in our programming and that is never bad.

GrahamCampbell: From what I can see, this is 3 months old, so running a buggy local version would not be a good idea?

GrahamCampbell: Better to have no local version than not the same as the CI I think?

GrahamCampbell: Another option is to do this for both static analysis tools: https://github.com/vlucas/phpdotenv/blob/4.1/composer.json#L26, https://github.com/vlucas/phpdotenv/blob/4.1/vendor-bin/phpstan/composer.json.

Nyholm: The vimeo psalm docker image is updated every 4 hours. 

https://hub.docker.com/r/vimeo/psalm-github-actions
https://github.com/psalm/psalm-github-actions/blob/master/.github/workflows/build.yml

I dont mind removing this from the Makefile all together. I dont feel like it is a good idea spending time on this.. I mean.. Power users have quick commands to run psalm anyways (like installing it globally) and others will have a Github review comment on the exact line with the issue.

psalm.baseline.xml
```diff
@@ -0,0 +1,87 @@
...
+  </file>
+  <file src="src/RedirectMiddleware.php">
+    <ImplicitToStringCast occurrences="1">
+      <code>$location</code>
+    </ImplicitToStringCast>
+  </file>
+  <file src="src/RetryMiddleware.php">
+    <TooManyArguments occurrences="1">
+      <code>($this-&gt;delay)(++$options['retries'], $response)</code>
+    </TooManyArguments>
```
gmponos, line 77: Can we add this `TooManyArguments` to ignore list?

Nyholm: Thank you for the review. 

Why should we ignore all instances of TooManyArguments? I rather manually confirm false positives.

gmponos: Scrap that.. I though that the error here says that the function has too many arguments.. something like a bad practice.. what it is meant to say is that you are using extra arguments to call a function...

Nyholm:
I've been using both in my daily work. They are reporting different things and I do thing they complement each other.

PR #2708, opened 2020-06-27
[7.1] Test coverage of Utils

GrahamCampbell:
We are removing the functions in Guzzle 8, but the new Utils class doesn't have direct coverage. This PR provides that. In Guzzle 8, we remove the parts from this new test class that call the functions, leaving the Utils class tested.


GrahamCampbell:
🎉

PR #2710, opened 2020-06-27
Removed deprecated code

GrahamCampbell:
This PR removes all the deprecated code from Guzzle 7.1+.


sagikazarmark:
End of an era 👍

Issue #2702, opened 2020-06-27
Support PHP 8


Nyholm:
We currently do not support PHP 8. We need to update CI config, update composer.json and make sure that we handle the new resource object in PHP 8.

GrahamCampbell:
I imagine phpstan might start to work against us here, lol. We'll have to be careful our types and phpdoc are accurate.

Nyholm:
This is fixed in master.

loading84:
Hello, 
![chybaroundcube](https://user-images.githubusercontent.com/15268574/113609507-2a856c00-964c-11eb-8e51-6ea45353d878.jpg)
I am installing a roundcube and it reports an error, see a screenshot, can you please advise me? I use php 8 in my linux server. Thanks.

PR #2712, opened 2020-06-27
Avoid "functions" from dependencies

GrahamCampbell:
The aim of this PR is to avoid the use of functions provided by our dependencies. This means using their new Utils classes, and bumping their minimum versions (`guzzlehttp/promises:^1.4`; `guzzlehttp/psr7:^1.7`). Note that similar changes were already implemented for Guzzle 7.0 itself, but not its dependencies, which we are going to address now, for Guzzle 7.1.

### TODO:

- [x] Implement upstream `guzzlehttp/promises` changes (https://github.com/guzzle/promises/pull/113)
- [x] Implement upstream `guzzlehttp/psr7` changes (https://github.com/guzzle/psr7/pull/345)
- [x] Merge https://github.com/guzzle/guzzle/pull/2720 and perform 6.5 -> 7.1 merge
- [x] Update `composer.json` with `^1.4` and `^1.7`
- [x] Replace all uses of functions with their static method replacements
- [x] Update documentation and examples
- [x] Tag `guzzlehttp/promises` 1.4.0 and remove `@dev`
- [x] Tag `guzzlehttp/psr7` 1.7.0 and remove `@dev`
- [x] Update change log

### Mappings:

| Original Function | Replacement Method |
|----------------|----------------|
| `GuzzleHttp\Promise\queue` | `GuzzleHttp\Promise\Utils::queue` |
| `GuzzleHttp\Promise\task` | `GuzzleHttp\Promise\Utils::task` |
| `GuzzleHttp\Promise\promise_for` | `GuzzleHttp\Promise\Create::promiseFor` |
| `GuzzleHttp\Promise\rejection_for` | `GuzzleHttp\Promise\Create::rejectionFor` |
| `GuzzleHttp\Promise\exception_for` | `GuzzleHttp\Promise\Create::exceptionFor` |
| `GuzzleHttp\Promise\iter_for` | `GuzzleHttp\Promise\Create::iterFor` |
| `GuzzleHttp\Promise\inspect` | `GuzzleHttp\Promise\Utils::inspect` |
| `GuzzleHttp\Promise\inspect_all` | `GuzzleHttp\Promise\Utils::inspectAll` |
| `GuzzleHttp\Promise\unwrap` | `GuzzleHttp\Promise\Utils::unwrap` |
| `GuzzleHttp\Promise\all` | `GuzzleHttp\Promise\Utils::all` |
| `GuzzleHttp\Promise\some` | `GuzzleHttp\Promise\Utils::some` |
[...]

PR #2714, opened 2020-06-28
Add GitHub Actions builds for PHP 8 in experimental mode

Ayesh:
Related: #2702

Guzzle currently fails to run tests on PHP 8 due to [`CurlHandle` migration](https://php.watch/versions/8.0/resource-CurlHandle) and `method_exists()` function throwing a `\TypeError` exception due to [internal functions enforcing types](https://php.watch/versions/8.0/internal-function-exceptions). 

This PR:

- Update composer.json PHP version requirements to allow PHP 8
- Update GitHub Actions to allow experimental PHP 8 builds
  - Update the PHP version matrix to include PHP 8 builds
  - Add a PHP8-only task that runs `composer install` with [`--ignore-platform-req=php`](https://php.watch/articles/composer-ignore-platform-req) flag.
  - Add a new experimental true/false matrix to allow overall build status to pass even if PHP 8 builds fail.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,10 +40,15 @@ jobs:
   build:
     name: Build
     runs-on: ubuntu-latest
+    continue-on-error: ${{ matrix.experimental }}
     strategy:
       max-parallel: 10
       matrix:
         php: ['7.2', '7.3', '7.4']
+        experimental: [false]
+        include:
+          - php: 8.0
+            experimental: true
 
     steps:
       - name: Set up PHP
@@ -66,7 +71,12 @@ jobs:
         uses: actions/checkout@v2
 
       - name: Download dependencies
+        if: matrix.php != '8.0'
         run: composer update --no-interaction --prefer-dist
+
+      - name: Download dependencies (ignore-platform-req)
+        if: matrix.php == '8.0'
+        run: composer update --no-interaction --prefer-dist --ignore-platform-req=php
 
       - name: Run tests
         run: make test
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -71,12 +71,7 @@ jobs:
         uses: actions/checkout@v2
 
       - name: Download dependencies
-        if: matrix.php != '8.0'
         run: composer update --no-interaction --prefer-dist
-
-      - name: Download dependencies (ignore-platform-req)
-        if: matrix.php == '8.0'
-        run: composer update --no-interaction --prefer-dist --ignore-platform-req=php
 
       - name: Run tests
         run: make test


Ayesh:
I will send separate PRs to PHP 8 test failures which are not that tricky to fix. Hopefully this PR can be merged first, so the new ones can have automatic CI tests on them.

GrahamCampbell:
It is much too early for this, and there are lots of changes needed for PHP 8, including changes to the Guzzle PSR7 library. This has already been discussed, and we must wait more until PHP 8 stops making changes in order to decide what to do.

GrahamCampbell:
In particular, our use of resources in phpdoc and types needed to change.

Ayesh:
> It is much too early for this, and there are lots of changes needed for PHP 8

I agree @GrahamCampbell 
It wouldn't hurt to allow PHP 8 in `composer.json`, and enable tests for PHP 8 (which does not fail entire build even if PHP 8 build fails) right?

As long as we don't tag a release with the `composer.json` change, users will not be able to install `guzzle/guzzle` due to current taggged version constraint (`^7.2`).

GrahamCampbell:
> It wouldn't hurt to allow PHP 8 in composer.json

GrahamCampbell:
> It wouldn't hurt to allow PHP 8 in composer.json

👎 for that, actually. It would hurt. However, work can start on PHP 8 support without putting it into the main codebase. In fact, even testing on PHP 8 can happen in master if we wanted, without allowing PHP 8 in the composer.json file.

Ayesh:
Thank you @GrahamCampbell for the clarification.
I rebased and force-pushed without the commit that updates `composer.json`. 

This is now purely a GitHub Actions update change.

GrahamCampbell:
I think this change should be PRed together with the other change, and PHP 8 not be allowed to fail in the CI. I don't think it helps keep things too atomic, in this instance.

Thanks for starting to work on this btw. :+1:

Nyholm:
Should we merge this PR with #2714

That would make sure the master ci is green at all time.

driesvints:
> Should we merge this PR with #2714

@Nyholm this is PR #2714? Were you referring to a different one?

Nyholm:
Oh, Im sorry. Im a bit confused =)

I'll let @GrahamCampbell finish #2712 and then I'll get back to this PR.

.github/workflows/ci.yml
```diff
@@ -66,7 +71,12 @@ jobs:
         uses: actions/checkout@v2
 
       - name: Download dependencies
+        if: matrix.php != '8.0'
         run: composer update --no-interaction --prefer-dist
 
+      - name: Download dependencies (ignore-platform-req)
+        if: matrix.php == '8.0'
+        run: composer update --no-interaction --prefer-dist --ignore-platform-req=php
```
GrahamCampbell, line 79: Do we still need to ignore platform reqs now?

GrahamCampbell (changes requested):

.github/workflows/ci.yml
Nyholm: No, our dependencies supports PHP8

Nyholm:
Could you rebase the PR and remove `ignore-platform`?

GrahamCampbell:
Please close this in favour of https://github.com/guzzle/guzzle/pull/2789.

PR #2715, opened 2020-06-28
Update Curl is_resource() to work with PHP 8 CurlHandle and CurlMultiHandle objects

Ayesh:
Updates `is_resource()` calls for curl handlers to properly work with PHP 8's `\CurlHandle` and `\CurlMultiHandle` objects.
See https://php.watch/versions/8.0/resource-CurlHandle

PHPDoc comments are updated to indicate that the variable/property can be either a resource or a Curl object.
In PHP 8, `curl_close()` calls are no-op, and and explicit `unset()` call is added to ensure the `\CurlHandle` object is destroyed.

---

This PR depends on #2714 (to enable tests). 
Related: #2702

---

On manual/forced PHP 8 [tests](https://github.com/Ayesh/guzzle/runs/816178711#step:8:336), there are two failures related to this repo (and 10 errors related to `promises`).

```
There were 2 failures:

1) GuzzleHttp\Test\Handler\CurlFactoryTest::testCreatesCurlHandle
Failed asserting that CurlHandle Object &00000000184dcf9b000000006cc08298 () is of type "resource".

##[error]/home/runner/work/guzzle/guzzle/tests/Handler/CurlFactoryTest.php:49

2) GuzzleHttp\Test\Handler\CurlFactoryTest::testEmitsProgressToFunction
Failed asserting that actual size 5 matches expected size 4.

##[error]/home/runner/work/guzzle/guzzle/tests/Handler/CurlFactoryTest.php:289

ERRORS!
Tests: 486, Assertions: 2092, Errors: 10, Failures: 2, Skipped: 5.
```

This PR fixes the two failures related to this repo:

```
Tests: 486, Assertions: 2120, Errors: 10, Skipped: 5.
```

### DIFF
--- a/src/Handler/CurlFactory.php
+++ b/src/Handler/CurlFactory.php
@@ -27,7 +27,7 @@ class CurlFactory implements CurlFactoryInterface
     public const LOW_CURL_VERSION_NUMBER = '7.21.2';
 
     /**
-     * @var resource[]
+     * @var resource[]|\CurlHandle[]
      */
     private $handles = [];
 
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -6,9 +6,99 @@ parameters:
 			path: src/Client.php
 
 		-
+			message: "#^Property GuzzleHttp\\\\Handler\\\\CurlFactory\\:\\:\\$handles has unknown class CurlHandle as its type\\.$#"
+			count: 1
+			path: src/Handler/CurlFactory.php
+
+		-
+			message: "#^Parameter \\#1 \\$ch of function curl_setopt_array expects resource, CurlHandle\\|resource given\\.$#"
+			count: 1
+			path: src/Handler/CurlFactory.php
+
+		-
+			message: "#^Parameter \\#1 \\$ch of function curl_close expects resource, CurlHandle\\|resource given\\.$#"
+			count: 1
+			path: src/Handler/CurlFactory.php
+
+		-
+			message: "#^Parameter \\#1 \\$ch of function curl_setopt expects resource, CurlHandle\\|resource given\\.$#"
+			count: 4
+			path: src/Handler/CurlFactory.php
+
+		-
+			message: "#^Parameter \\#1 \\$ch of function curl_reset expects resource, CurlHandle\\|resource given\\.$#"
+			count: 1
+			path: src/Handler/CurlFactory.php
+
+		-
+			message: "#^Parameter \\#1 \\$ch of function curl_getinfo expects resource, CurlHandle\\|resource given\\.$#"
+			count: 4
+			path: src/Handler/CurlFactory.php
+
+		-
+			message: "#^Parameter \\#1 \\$ch of function curl_error expects resource, CurlHandle\\|resource given\\.$#"
+			count: 1
+			path: src/Handler/CurlFactory.php
+
+		-
 			message: "#^Parameter \\#1 \\$str1 of function strcasecmp expects string, int\\|string given\\.$#"
 			count: 1
 			path: src/Handler/CurlFactory.php
+
+		-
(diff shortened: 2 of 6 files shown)


src/Handler/CurlFactory.php
```diff
@@ -75,6 +75,7 @@ public function release(EasyHandle $easy): void
 
         if (\count($this->handles) >= $this->maxHandles) {
             \curl_close($resource);
+            unset($resource);
```
nikic, line 78: This is not needed, the `$resource` variable is automatically destroyed when the function return (which happens directly after this).

src/Handler/CurlFactory.php
```diff
@@ -21,7 +21,7 @@ class CurlFactory implements CurlFactoryInterface
     public const LOW_CURL_VERSION_NUMBER = '7.21.2';
 
     /**
-     * @var resource[]
+     * @var resource|\CurlHandle[]
```
nikic, line 24:
```suggestion
     * @var (resource|\CurlHandle)[]
```
I think this will be interpreted as "single resource or multiple CurlHandles" otherwise.

Ayesh: You are right, thank you @nikic. I'm reading at [PHPDoc page](https://docs.phpdoc.org/latest/guides/types.html#multiple-types-combined), and it also has a similar example. 

I tested the brace syntax with PHPStorm, which it didn't seem to understand. 

`resource[]|\CurlHandle[]` worked:

image

src/Handler/CurlFactory.php
Ayesh: Thank you. I will update this.

Nyholm:
Could you make sure we test this on php8 too?

GrahamCampbell:
PHP 8 is pending https://github.com/guzzle/guzzle/pull/2712 first, to avoid conflicts and to ensure the dependencies work on PHP 8.

GrahamCampbell:
This needs rebasing. :)

Nyholm (approved):
I'll be happy to merge this once you rebase it and the CI is green

Ayesh:
Thank you @Nyholm and @GrahamCampbell - I just force-pushed after rebasing from current `master` branch.

Nyholm:
Could you also make sure the CI is green?

Ayesh:
Both test failures are due to the PHP 8's changes in parameter types. I suppose both static analyzers infer the parameter types via Reflection.

It wouldn't be ideal if we were to make psalm/phpstan suppress these errors, but that's the only approach I can think of for now. Even if were to run the static analyzers on PHP 8, it would now complain about the other type of parameters too.

Ayesh:
I will make changes in psalm/phpstan config files so it ignores the type mismatch false positives.

Nyholm:
I agree with you. Let's update the baselines for now.

Ayesh:
Thanks for updating the baselines, it wasn't something I had a lot of experience.

Nyholm:
I ran the following commands: 

```
phpstan analyze  --generate-baseline
psalm --set-baseline=psalm.baseline.xml
```

Nyholm:
Thank you for your works

mfn:
> I ran the following commands:
> 
> ```
> phpstan analyze  --generate-baseline
> psalm --set-baseline=psalm.baseline.xml
> ```

I was about to suggest to add this to `composer.json` `scripts` but then realized neither phpstan/psalm is actually a dev dependency, so that would be odd.

This could be unnecessary friction for newcomers, can't expect everyone to know the ins/outs of all the tools I guess, at least when it comes to handling things like updating the baseline.

Any ideas how to make this easier? 🤔

Nyholm:
Running these commands are rare. Newcomers will never run this.

PR #2721, opened 2020-07-01
Removed PHP 5 work arounds and fix PHP 8 call_user_* issues

GrahamCampbell:
// cc @gmponos this safely removes `call_user_func_array` too

### DIFF
--- a/src/Handler/MockHandler.php
+++ b/src/Handler/MockHandler.php
@@ -77,7 +77,8 @@ class MockHandler implements \Countable
         $this->onRejected = $onRejected;
 
         if ($queue) {
-            \call_user_func_array([$this, 'append'], $queue);
+            // array_values included for BC
+            $this->append(...array_values($queue));
         }
     }
 
--- a/src/Handler/StreamHandler.php
+++ b/src/Handler/StreamHandler.php
@@ -503,7 +503,7 @@ class StreamHandler
      */
     private function add_progress(RequestInterface $request, array &$options, $value, array &$params): void
     {
-        $this->addNotification(
+        self::addNotification(
             $params,
             static function ($code, $a, $b, $c, $transferred, $total) use ($value) {
                 if ($code == \STREAM_NOTIFY_PROGRESS) {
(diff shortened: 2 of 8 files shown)


GrahamCampbell:
The phpstan warning seems to be a bug in phpstan. I'll re-gen the baseline and report the bug to them.

GrahamCampbell:
https://github.com/phpstan/phpstan/issues/3559

src/Handler/StreamHandler.php
```diff
@@ -551,25 +549,24 @@ static function () use ($ident, $value, $map, $args): void {
         );
     }
 
-    private function addNotification(array &$params, callable $notify): void
+    private static function addNotification(array &$params, callable $notify): void
```
gmponos, line 552: what's the benefit of having them static? is this for some perf. reasons?

src/Handler/MockHandler.php
```diff
@@ -75,7 +75,8 @@ public function __construct(
         $this->onRejected = $onRejected;
 
         if ($queue) {
-            \call_user_func_array([$this, 'append'], $queue);
+            // array_values included for BC
+            $this->append(...array_values($queue));
```
gmponos, line 79: Nice workaround.

src/Handler/StreamHandler.php
GrahamCampbell: Yeh, I just noticed these private functions could be made static. This also makes them clearer to the programmer that they won't have any side effects on the instance.

GrahamCampbell:
NB This PR also fixes issues with PHP 8, since `call_user_func_array` now gives array keys meaning, and we don't want that.

GrahamCampbell:
For that reason, and chance we could sneak this into 7.1.0 @Nyholm?

Nyholm (approved):
This looks good. 

No, lets not wedge more things into 7.1. I've looked that version in =)
I would be more happy to release a 7.2 or 7.1.x with this later this month

GrahamCampbell:
Cool, just rebased this to fix conflicts. Shame about the false positive from phpstan. https://3v4l.org/p3XfW

Nyholm:
Thank you

PR #2725, opened 2020-07-11
[8.0] Make Pool class final

GrahamCampbell:
Do we wanna `@final` phpdoc it in 7.1 too, if we want this at all?


GrahamCampbell:
This needs replacing by a PR to mark all classes as final that have a soft-final annotation. That can be done later, once we have definitely stopped adding such annotations.

Issue #2521, opened 2019-12-30
Mark classes with final annotation in 7.1


gmponos:
As per the title says here we can have a discussion about adding final annotation to classes and then making them final on version 8

The reason we want to do that on 7.1 is that we don't want towards version 7 to introduce more major changes than the ones we already have.. 

So the thought is to make them final on version 8. Before that there is a though about adding final annotations to classes on version 7.1

GrahamCampbell:
Big :-1: on this. I am totally against annotations that are not actually enforced, when it comes to something being final.

GrahamCampbell:
I am in favour of making some things properly final in version 8 (like the client, BUT only if we have the interface changes).

gmponos:
> Big 👎 on this.

But what is the reason? What difference will make for you as a developer if there is a final annotation. The reason we want to have final annotation is in order to inform the developers that we will do the change on v8.. something like a deprecation... If I could do it with a deprecation then I would.. If it is possible then let me know.. 

if I am not mistaken symfony does the same with annotations as well..

GrahamCampbell:
Because it's being marked as final in a missleading way. It's not actually final.

GrahamCampbell:
> if I am not mistaken symfony does the same with annotations as well..

Really? Can you show me a class from Synfony 5.0 that is annotated final, but is not actually final.

gmponos:
https://github.com/symfony/symfony/blob/4.4/src/Symfony/Bundle/FrameworkBundle/Controller/AbstractController.php#L65

GrahamCampbell:
That's marking a method as final, not a class?

gmponos:
For the reason that it is used I fail to see the difference but never the less here is another example.. https://github.com/symfony/symfony/blob/4.4/src/Symfony/Bundle/FrameworkBundle/Command/ConfigDebugCommand.php#L28

> It's not actually final.

Yes.. that's the goal.. otherwise it would've been made on v7.

GrahamCampbell:
Hmmm. I guess symfony has a special reason for doing this. Are they only doing it to 4.4 classes because of the simultaneous release of 5.0 where they become enforced? This is kinda difference to just randomly doing it in 4.2.0.

gmponos:
> Hmmm. I guess symfony has a special reason for doing this. 

[...]

PR #2726, opened 2020-07-12
Use final annotation to mark classes as final

gmponos:
Mark classes as final

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -24,9 +24,3 @@ parameters:
 			message: "#^Result of && is always false\\.$#"
 			count: 1
 			path: src/Middleware.php
-
-		-
-			message: "#^Unsafe usage of new static\\(\\)\\.$#"
-			count: 1
-			path: src/Pool.php
-
--- a/src/Client.php
+++ b/src/Client.php
@@ -10,6 +10,9 @@ use Psr\Http\Message\RequestInterface;
 use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\UriInterface;
 
+/**
+ * @final
+ */
 class Client implements ClientInterface, \Psr\Http\Client\ClientInterface
 {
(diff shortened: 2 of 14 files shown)


gmponos:
This is related with #2725  and it also closes #2521

gmponos:
Are there more classes that we should mark as final?

GrahamCampbell:
I think this is enough for now. You'll need to remove that extra entry from the phpstan baseline file too. ;)

vaclavvanik:
MessageFormatter does not implement interface. If it will be final there is no way have to use own implementation.

GrahamCampbell:
> MessageFormatter does not implement interface. If it will be final there is no way have to use own implementation.

Why would someone want to extend that class, though?

GrahamCampbell:
If you wanna have your own formatter class, it doesn't need to extend Guzzle's. You can just call the public API of the original class from your new class, without needing to extend it.

gmponos:
> If you wanna have your own formatter class, it doesn't need to extend Guzzle's. You can just call the public API of the original class from your new class, without needing to extend it.

But you can't inject your custom class into the log middleware.. there might be some value having an interface for `MessageFormatter`.

Would appreciate any feedback on that.

GrahamCampbell:
Hmmm, that's true. Is anyone actually using a custom impl?

vaclavvanik:
Actually I use my own implementation. Currently I extend MessageFormatter
because there is no interface. My use case is that I do not log message
body greater than 64 kB.

Dne ne 12. 7. 2020 11:33 uživatel Graham Campbell <notifications@github.com>
napsal:

GrahamCampbell:
Ok, so I think we can still add the soft final annotation, but we also should add a `MessageFormatterInterface` and change the type on the log middleware to that, which shouldn't be a breaking change.

Nyholm (approved):
Thank you

PR #2728, opened 2020-07-13
Be explicit and replace string casting with __toString function

gmponos:
Why?

Let's just for a moment imagine that `$response->getBody()` will return an int for WHATEVER THE HELL reason.. this will still work:

`(string) $response->getBody()`.. So replacing it with `__toString()` makes things strict. 

Do we care? yes they are edge cases what so ever but I think we should promote this.

This PR also includes https://github.com/guzzle/guzzle/issues/2498#issuecomment-657200787

### DIFF
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -7,6 +7,7 @@ $config = PhpCsFixer\Config::create()
         'array_syntax' => ['syntax' => 'short'],
         'binary_operator_spaces' => ['operators' => ['=>' => null]],
         'blank_line_after_opening_tag' => true,
+        'cast_spaces' => ['space' => 'single'],
         'class_attributes_separation' => ['elements' => ['method']],
         'compact_nullable_typehint' => true,
         'concat_space' => ['spacing' => 'one'],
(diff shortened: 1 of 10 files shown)


GrahamCampbell:
Hmmm, I'd be against calling magic methods like this. I wouldn't be surprised if those sorts of methods were made non-callable or non-public in the future.

> $response->getBody() will return an int

mfn:
I very much would applaud this.

It's not "just magic", it's a contract method. It should be promoted more because most people don't realize that `getContents()` doesn't rewind before reading whereas `(string)` / `__toString()` does.

However being at least able to jump to the interface definition helps _a bit_ to better understand this.

>  I wouldn't be surprised if those sorts of methods were made non-callable or non-public in the future.

Source? :)

GrahamCampbell:
> Source? :)

Just my wild speculation. Why should __toString be any different to __get. We don't directly call that method (under usual circumstances).

mfn:
Ok, well if that's a concern, the interface should change it to `toString()`, but that's a different issue then 😄

GrahamCampbell:
Right, but my point is that the PSR-7 implementation specifically chooses to not do that, and to use the magic method. Unfortunately neither the spec or the meta document actually spell out how this method was intended to be called.

sagikazarmark:
Stringable makes it kind explicit, isn't it? https://wiki.php.net/rfc/stringable

PR #2729, opened 2020-07-14
Add MessageFormatterInterface

vaclavvanik:
`MessageFormatterInterface` allow custom formatting implementations discussed in [#PR2726](https://github.com/guzzle/guzzle/pull/2726).

One question - move log formats to MessageFormatterInterface? Imho it is ok.

### DIFF
--- a/src/MessageFormatter.php
+++ b/src/MessageFormatter.php
@@ -43,8 +43,6 @@ class MessageFormatter implements MessageFormatterInterface
      * @link https://httpd.apache.org/docs/2.4/logs.html#common
      *
      * @var string
-     *
-     * @deprecated log format constants will be moved into MessageFormatterInterface in guzzlehttp/guzzle:8.0.
      */
     public const CLF = "{hostname} {req_header_User-Agent} - [{date_common_log}] \"{method} {target} HTTP/{version}\" {code} {res_header_Content-Length}";
     public const DEBUG = ">>>>>>>>\n{request}\n<<<<<<<<\n{response}\n--------\n{error}";
--- a/src/MessageFormatter.php
+++ b/src/MessageFormatter.php
@@ -38,6 +38,19 @@ use Psr\Http\Message\ResponseInterface;
 class MessageFormatter implements MessageFormatterInterface
 {
     /**
+     * Apache Common Log Format.
+     *
+     * @link https://httpd.apache.org/docs/2.4/logs.html#common
+     *
+     * @var string
+     *
+     * @deprecated log format constants will be moved into MessageFormatterInterface in guzzlehttp/guzzle:8.0.
+     */
+    public const CLF = "{hostname} {req_header_User-Agent} - [{date_common_log}] \"{method} {target} HTTP/{version}\" {code} {res_header_Content-Length}";
+    public const DEBUG = ">>>>>>>>\n{request}\n<<<<<<<<\n{response}\n--------\n{error}";
+    public const SHORT = '[{ts}] "{method} {target} HTTP/{version}" {code}';
+
+    /**
      * @var string Template used to format log messages
      */
     private $template;
--- a/src/MessageFormatter.php
+++ b/src/MessageFormatter.php
@@ -32,20 +32,11 @@ use Psr\Http\Message\ResponseInterface;
  * - {res_headers}:    Response headers
  * - {req_body}:       Request body
  * - {res_body}:       Response body
+ *
+ * @final
  */
-class MessageFormatter
+class MessageFormatter implements MessageFormatterInterface
 {
-    /**
-     * Apache Common Log Format.
-     *
-     * @link https://httpd.apache.org/docs/2.4/logs.html#common
-     *
-     * @var string
-     */
-    public const CLF = "{hostname} {req_header_User-Agent} - [{date_common_log}] \"{method} {target} HTTP/{version}\" {code} {res_header_Content-Length}";
-    public const DEBUG = ">>>>>>>>\n{request}\n<<<<<<<<\n{response}\n--------\n{error}";
-    public const SHORT = '[{ts}] "{method} {target} HTTP/{version}" {code}';
-
     /**
      * @var string Template used to format log messages
      */
--- a/src/MessageFormatterInterface.php
+++ b/src/MessageFormatterInterface.php
@@ -1,0 +1,33 @@
+<?php
+
+namespace GuzzleHttp;
+
+use Psr\Http\Message\RequestInterface;
+use Psr\Http\Message\ResponseInterface;
+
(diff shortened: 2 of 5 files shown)


src/MessageFormatterInterface.php
```diff
@@ -0,0 +1,33 @@
...
+    /**
+     * Apache Common Log Format.
+     *
+     * @link https://httpd.apache.org/docs/2.4/logs.html#common
+     *
+     * @var string
+     */
+    public const CLF = "{hostname} {req_header_User-Agent} - [{date_common_log}] \"{method} {target} HTTP/{version}\" {code} {res_header_Content-Length}";
+    public const DEBUG = ">>>>>>>>\n{request}\n<<<<<<<<\n{response}\n--------\n{error}";
+    public const SHORT = '[{ts}] "{method} {target} HTTP/{version}" {code}';
```
gmponos, line 19: Although I am 👍 for this.. it might be a BC for some ppl that might have been extending the `MessageFormatter` class implementation and overriding these constants.. now they wont be able to.

vaclavvanik: I supposed that `MessageFormatter` will be marked as final and thus extending will be prohibited.

gmponos: Yes.. but we add the final annotation and not the final php keyword in order not to hard enforce this rule on 7. So developers can still extend this class. We will enforce the final on v8.. on v8 we could move these to the interface as you did..

vaclavvanik: Ok. For v7 I move consts back to `MessageFormatter`, ok?

GrahamCampbell: I'd say put them in both, but annotate the ones in the class as deprecated.

gmponos: https://3v4l.org/hbjIH

GrahamCampbell: Oh, if you put them only in the interface, are they still accessible as if they were on the class, then, so not a BC break?

gmponos: It is a breaking change for ppl that use to extend the MessageFormatter class.

vaclavvanik: I moved format consts back into `MessageFormatter` and annotated they will be moved in v8.

src/MessageFormatter.php
```diff
@@ -32,15 +32,19 @@
...
+class MessageFormatter implements MessageFormatterInterface
 {
     /**
      * Apache Common Log Format.
      *
      * @link https://httpd.apache.org/docs/2.4/logs.html#common
      *
      * @var string
+     *
+     * @deprecated log format constants will be moved into MessageFormatterInterface in guzzlehttp/guzzle:8.0.
```
gmponos, line 47: We can not add a deprecation without providing an alternative.. Yes.. these will be moved to the interface but currently they are not there..

src/MessageFormatterInterface.php
GrahamCampbell: Is it? https://3v4l.org/acm6d

gmponos: https://3v4l.org/nipM9

gmponos: after the move to the interface ppl will have problem

src/MessageFormatter.php
vaclavvanik: Which annotation you suggest?

gmponos: No annotation. Leave things as they were...

GrahamCampbell: Wait, why did you wanna move it to the interface, anyway?

vaclavvanik: As this is (Apache) common log format it may be directly in interface. imho.

gmponos: Even if this is a standard format.. you never know why someone might have overwritten it.. it is a BC.. deprecated SHOULD go with an alternative IMHO. Saying to ppl that we deprecate stuff but the alternative is on a FUTURE version is not good.

GrahamCampbell: Yep. I please keep the constants on the original class, and don't annotate them with anything. In 8.x we can make the class final, and move the constants to the interface.

src/MessageFormatterInterface.php
gmponos: W8 W8.. my previous code is wrong...

Out of PHP documentation:

> It's possible for interfaces to have constants. Interface constants work exactly like class constants except they cannot be overridden by a class/interface that inherits them. 

but check this:

https://3v4l.org/qrZmn

why is this working 🤔

GrahamCampbell: :trollface: D'no

gmponos: So if it's a first level child it works as expected https://3v4l.org/ctK02

gmponos: Concluding this is a PHP bug https://bugs.php.net/bug.php?id=73348

src/MessageFormatter.php
vaclavvanik: Annotation removed.

gmponos:
TBH... don't know why BC is failing

vaclavvanik:
```
[BC] CHANGED: The parameter $formatter of GuzzleHttp\Middleware::log() changed from GuzzleHttp\MessageFormatter to a non-contravariant GuzzleHttp\MessageFormatterInterface
1 backwards-incompatible changes detected
```

But I do not know how to fix it.

Nyholm:
As we noticed with the Guzzle 6.5 release. People have multiple installations of Guzzle on the same system. 

If I use a Guzzle 7.0 MessageFormatter with a 7.1 Middleware, then I get an error. 

I suggest to remove the type hint and verify the class/interface of the object instead. That will work for every weird edge case.

Nyholm:
Friendly ping

Nyholm (approved):
Thank you

src/MessageFormatter.php
```diff
@@ -32,8 +32,10 @@
  * - {res_headers}:    Response headers
  * - {req_body}:       Request body
  * - {res_body}:       Response body
+ *
+ * @final
```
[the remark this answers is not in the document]
GrahamCampbell: Correct. This class is not meant to be extended. It will become properly final in Guzzle 8.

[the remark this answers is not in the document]
GrahamCampbell: It is intended that you use composition instead of extension.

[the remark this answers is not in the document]
TheLevti: Thank you very much 👌

PR #2730, opened 2020-07-16
Updated PHP version in documentation

edwinvdpol:
As composer.json and contributing guidelines state, the minimum version is 7.2.5.


GrahamCampbell:
I think the minimum version of cURL is probably outdated too.

PR #2731, opened 2020-07-17
Throwing errors because mark is not enrolled in github sponsors

hannesvdvreken:
Screenshot 2020-07-17 at 18 42 28

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,3 +1,3 @@
 # These are supported funding model platforms
 
-github: [Nyholm, gmponos, sagikazarmark, alexeyshockov, GrahamCampbell]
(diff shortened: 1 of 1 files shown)


gmponos:
Hi.

Even if it shows an error what's the problem with that? Does it block you from funding the rest of the maintainers? Otherwise I don't see any reason removing Mark.

GrahamCampbell:
Well, if Mark doesn't have a sponsors account, maybe he wants to be funded in a different way. We should ask. :)

hannesvdvreken:
@sagikazarmark Sign up! https://github.com/sponsors

Nyholm:
Mark, let me know when you signed up and want to be on this list.

sagikazarmark:
Sadly, GitHub sponsors is not available in Hungary. :(

PR #2736, opened 2020-07-23
Release 7.0.2

GrahamCampbell:


CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ Please refer to [UPGRADING](UPGRADING.md) guide for upgrading to a major version
 
 ## UNRELEASED
 
+## 7.0.2 - 2020-07-23
+
+* Fix type error when providing query parameters [#2724](https://github.com/guzzle/guzzle/pull/2724)
```
gmponos, line 9: Hi.. IIRC PHP is not complaining because of this.. because we don't have strict types enabled.

I don't think it's worthe the trouble of tagging.

GrahamCampbell: Oh, I thought we did!

GrahamCampbell: In which case, I would have only targetted 7.1.x, with the fix. Oh well. 😆

PR #2741, opened 2020-08-04
Fix response body streams in Windows

### LINKED ISSUE #2735: Incorrect EOF detection for response body streams on Windows causing missing last chunks
**Guzzle version(s) affected**: 7.0.1, 6.5.5
**PHP version**: PHP 7.4.8 (x64, VC15, NTS)
**cURL version**: cURL 7.70.0

**Description**  

When reading the response body stream using a loop similar to:

```php
	while (true) {
		$eof = $response->getBody()->eof();

		if ($eof) break;

		$chunk = $response->getBody()->read(4096);
	}
```

the last chunk will be missing if the response length is between `$n * 4096 + 1` (inclusive)  and `($n + 1) * 4096` (exclusive) with `$n` being an even integer. In other words: If the last chunk were in the second half of a 8kB block or if reading the full file requires an even number of iterations.

The response would appear to be truncated at exactly a 4096 Byte boundary.

**How to reproduce**  

Running the following example script with `SHOW_BUG` on Windows will print `BUG FOR SIZE` for sizes 4097, 8191, 12289, 16383. 

```php
<?php
const SHOW_BUG = true;

require('vendor/autoload.php');

$sizes = [ 4095, 4096, 4097, 8191, 8192, 8193, 12287, 12288, 12289, 16383, 16384, 16385 ];

foreach ($sizes as $size) {
	printf("Downloading size(%d)\n", $size);

	if (SHOW_BUG) {
		$options = [
		];
	} else {
		$options = [
			'sink' => \GuzzleHttp\Psr7\stream_for(fopen("php://temp", "w+"))
		];
	}

	$response = $client->get('https://gist.githubusercontent.com/TimWolla/1c56cbb9924e2fdf13b5cd1597ac981f/raw/02885d2eab708619d8a24a77155a47e93453ea2f/'.$size.'.txt', $options);
	$i = 0; $total = 0;
	while (true) {
		printf("Iteration %d: ", ++$i);

		$eof = $response->getBody()->eof();
		printf("EOF(%s) ", var_export($eof, true));

		if ($eof) break;

		$chunk = $response->getBody()->read(4096);
		$len = strlen($chunk);
		$total += $len;
		printf("bytes(%d) total(%d)", $len, $total);
		echo "\n";
	}
	echo "\n";
	if ($total !== $size) {
		printf("BUG FOR SIZE %d\n", $size);
	}

	echo "\n";
}
```

The bug can also be reproduced using a simplified example only relying on `curl`:

```php
<?php
const SHOW_BUG = true;

if (SHOW_BUG) {
	$filename = "php://temp";
}
else {
	$filename = __DIR__."/temp";
}

$sizes = [ 4095, 4096, 4097, 8191, 8192, 8193, 12287, 12288, 12289, 16383, 16384, 16385 ];

foreach ($sizes as $size) {
	printf("Downloading size(%d)\n", $size);

	$ch = curl_init();
	curl_setopt($ch, CURLOPT_URL, 'https://gist.githubusercontent.com/TimWolla/1c56cbb9924e2fdf13b5cd1597ac981f/raw/02885d2eab708619d8a24a77155a47e93453ea2f/'.$size.'.txt');
	curl_setopt($ch, CURLOPT_FILE, $f = fopen($filename, "w+"));
	curl_exec($ch);
	curl_close($ch);

	fseek($f, 0);

	$i = 0; $total = 0;
	while (true) {
		printf("Iteration %d: ", ++$i);
		
		$eof = feof($f);
		printf("EOF(%s) ", var_export($eof, true));

		if ($eof) break;

		$chunk = fread($f, 4096);
		$len = strlen($chunk);
		$total += $len;
		printf("bytes(%d) total(%d)", $len, $total);
		echo "\n";
	}
	echo "\n";
	if ($total !== $size) {
		printf("BUG FOR SIZE %d\n", $size);
	}

	echo "\n";
}
```

You can find a complete reproducer environment within the following guzzle.tar.gz: [guzzle.tar.gz](https://github.com/guzzle/guzzle/files/4953004/guzzle.tar.gz). `test.php` is the Guzzle based reproducer. `test2.php` is the raw curl based reproducer.

Example output of `test.php`:

```

Downloading size(4095)
Iteration 1: EOF(false) bytes(4095) total(4095)
Iteration 2: EOF(true) 

Downloading size(4096)
Iteration 1: EOF(false) bytes(4096) total(4096)
Iteration 2: EOF(true) 

Downloading size(4097)
Iteration 1: EOF(false) bytes(4096) total(4096)
Iteration 2: EOF(true) 
BUG FOR SIZE 4097

Downloading size(8191)
Iteration 1: EOF(false) bytes(4096) total(4096)
Iteration 2: EOF(true) 
BUG FOR SIZE 8191

Downloading size(8192)
Iteration 1: EOF(false) bytes(4096) total(4096)
Iteration 2: EOF(false) bytes(4096) total(8192)
Iteration 3: EOF(false) bytes(0) total(8192)
Iteration 4: EOF(true) 

Downloading size(8193)
Iteration 1: EOF(false) bytes(4096) total(4096)
Iteration 2: EOF(false) bytes(4096) total(8192)
Iteration 3: EOF(false) bytes(1) total(8193)
Iteration 4: EOF(true) 

Downloading size(12287)
Iteration 1: EOF(false) bytes(4096) total(4096)
Iteration 2: EOF(false) bytes(4096) total(8192)
Iteration 3: EOF(false) bytes(4095) total(12287)
Iteration 4: EOF(true) 

Downloading size(12288)
Iteration 1: EOF(false) bytes(4096) total(4096)
Iteration 2: EOF(false) bytes(4096) total(8192)
Iteration 3: EOF(false) bytes(4096) total(12288)
Iteration 4: EOF(true) 

Downloading size(12289)
Iteration 1: EOF(false) bytes(4096) total(4096)
Iteration 2: EOF(false) bytes(4096) total(8192)
Iteration 3: EOF(false) bytes(4096) total(12288)
Iteration 4: EOF(true) 
BUG FOR SIZE 12289

Downloading size(16383)
Iteration 1: EOF(false) bytes(4096) total(4096)
Iteration 2: EOF(false) bytes(4096) total(8192)
Iteration 3: EOF(false) bytes(4096) total(12288)
Iteration 4: EOF(true) 
BUG FOR SIZE 16383

Downloading size(16384)
Iteration 1: EOF(false) bytes(4096) total(4096)
Iteration 2: EOF(false) bytes(4096) total(8192)
Iteration 3: EOF(false) bytes(4096) total(12288)
Iteration 4: EOF(false) bytes(4096) total(16384)
Iteration 5: EOF(false) bytes(0) total(16384)
Iteration 6: EOF(true) 

Downloading size(16385)
Iteration 1: EOF(false) bytes(4096) total(4096)
Iteration 2: EOF(false) bytes(4096) total(8192)
Iteration 3: EOF(false) bytes(4096) total(12288)
Iteration 4: EOF(false) bytes(4096) total(16384)
Iteration 5: EOF(false) bytes(1) total(16385)
Iteration 6: EOF(true) 
```

**Possible Solution**  

My understanding of this bug is that directly writing into a `php://temp` stream using `CURLOPT_FILE` is unsafe on Windows. It appears as if PHP does not correctly recognize the stream size increasing there.

This issue can be worked around by always using `CURLOPT_WRITEFUNCTION` to manually write into the stream (as used when a `sink` is explicitely specified). Another alternative would be using the [`tmpfile()`](https://www.php.net/manual/en/function.tmpfile.php) function, which always creates a filesystem backed file.

This most likely is a bug in either PHP or the curl extension, but it's easy enough to work around in Guzzle, that's why I am filing this issue here. I expect a faster turnaround time from you than from PHP.

**Additional context**  

Find the reproducer environment and example outputs of the scripts below:

[guzzle.tar.gz](https://github.com/guzzle/guzzle/files/4953019/guzzle.tar.gz)
[test2-show-bug-false.txt](https://github.com/guzzle/guzzle/files/4953020/test2-show-bug-false.txt)
[test2-show-bug-true.txt](https://github.com/guzzle/guzzle/files/4953021/test2-show-bug-true.txt)
[test-show-bug-false.txt](https://github.com/guzzle/guzzle/files/4953022/test-show-bug-false.txt)
[test-show-bug-true.txt](https://github.com/guzzle/guzzle/files/4953023/test-show-bug-true.txt)

The application in question is hosted by our customers on-premise. We are using Guzzle 6.5.x, because we can't rely on PHP 7.2 being available on customer's webhosts. Please apply a fix to both Guzzle 6.5 as well as Guzzle 7.

TimWolla:
Fixes #2735

### DIFF
--- a/tests/Handler/CurlFactoryTest.php
+++ b/tests/Handler/CurlFactoryTest.php
@@ -757,4 +757,32 @@ class CurlFactoryTest extends TestCase
 
         self::assertEmpty($factory->release($easyHandle));
     }
+
+    /**
+     * https://github.com/guzzle/guzzle/issues/2735
+     */
+    public function testBodyEofOnWindows()
+    {
+        $expectedLength = 4097;
+
+        Server::flush();
+        Server::enqueue([
+            new Psr7\Response(200, [
+                'Content-Length' => $expectedLength,
+            ], \str_repeat('x', $expectedLength))
+        ]);
+
+        $handler = new Handler\CurlMultiHandler();
+        $request = new Psr7\Request('GET', Server::$url);
+        $promise = $handler($request, []);
+        $response = $promise->wait();
+        $body = $response->getBody();
+
+        $actualLength = 0;
+        while (!$body->eof()) {
+            $chunk = $body->read(4096);
+            $actualLength += \strlen($chunk);
+        }
+        self::assertSame($expectedLength, $actualLength);
+    }
 }
--- a/src/Handler/CurlFactory.php
+++ b/src/Handler/CurlFactory.php
@@ -6,7 +6,6 @@ use GuzzleHttp\Exception\ConnectException;
 use GuzzleHttp\Exception\RequestException;
 use GuzzleHttp\Promise\FulfilledPromise;
 use GuzzleHttp\Promise\PromiseInterface;
-use GuzzleHttp\Psr7;
 use GuzzleHttp\Psr7\LazyOpenStream;
 use GuzzleHttp\TransferStats;
 use GuzzleHttp\Utils;
@@ -395,29 +394,27 @@ class CurlFactory implements CurlFactoryInterface
 
         // Do not connect a sink for HEAD requests.
         if ($easy->request->getMethod() !== 'HEAD') {
-            if (isset($options['sink'])) {
-                $sink = $options['sink'];
-                if (!\is_string($sink)) {
-                    $sink = \GuzzleHttp\Psr7\stream_for($sink);
-                } elseif (!\is_dir(\dirname($sink))) {
-                    // Ensure that the directory exists before failing in curl.
-                    throw new \RuntimeException(\sprintf(
-                        'Directory %s does not exist for sink value of %s',
-                        \dirname($sink),
-                        $sink
-                    ));
-                } else {
-                    $sink = new LazyOpenStream($sink, 'w+');
-                }
-                $easy->sink = $sink;
-                $conf[\CURLOPT_WRITEFUNCTION] = static function ($ch, $write) use ($sink): int {
-                    return $sink->write($write);
-                };
-            } else {
+            if (!isset($options['sink'])) {
                 // Use a default temp stream if no sink was set.
-                $conf[\CURLOPT_FILE] = \fopen('php://temp', 'w+');
-                $easy->sink = Psr7\stream_for($conf[\CURLOPT_FILE]);
+                $options['sink'] = \fopen('php://temp', 'w+');
             }
+            $sink = $options['sink'];
+            if (!\is_string($sink)) {
+                $sink = \GuzzleHttp\Psr7\stream_for($sink);
+            } elseif (!\is_dir(\dirname($sink))) {
+                // Ensure that the directory exists before failing in curl.
+                throw new \RuntimeException(\sprintf(
+                    'Directory %s does not exist for sink value of %s',
+                    \dirname($sink),
+                    $sink
+                ));
+            } else {
+                $sink = new LazyOpenStream($sink, 'w+');
+            }
+            $easy->sink = $sink;
+            $conf[\CURLOPT_WRITEFUNCTION] = static function ($ch, $write) use ($sink): int {
+                return $sink->write($write);
+            };
         }
 
         $timeoutRequiresNoSignal = false;


TimWolla:
Note:

The change causes this assertion to fail: https://github.com/guzzle/guzzle/blob/e8ed4dbf49b260ff129ff0e0400718c3269971bf/tests/Handler/CurlFactoryTest.php#L83

I'm not sure what exactly that assertion is supposed to test and why `CURLOPT_WRITEFUNCTION` is in there, but `CURLOPT_FILE` is not. I believe the test is incorrect, but I'd like your feedback regarding that.

Nyholm:
There is no way for me to verify the correctness of this. Could you add a test? 
Please also make sure that test is running on a windows ci.

TimWolla:
> There is no way for me to verify the correctness of this. Could you add a test?

I will have to check. I already have some issues getting the test suite to run on my local Ubuntu machine in Docker. And then there's the open question regarding the failing test.

My issue #2735 contains a simple reproducer script, though. It should be easyish turning that into a test, I suppose. The only important thing is that the response has the appropriate byte length and that the code performs an explicit loop checking for `->eof()`.

> Please also make sure that test is running on a windows ci.

From the looks of it there is no existing set-up for Windows CI for Guzzle and I never used any Windows CI before, so I don't really know where to start setting that up.

Nyholm:
I’m afraid that we are in the same boat. :/ 
I’ve not used windows for the past two decades. I guess GitHub actions got some docs how to set things up. I think searching in their docs is a good place to start.

TimWolla:
Okay, I'll have a look at getting GitHub Actions on Windows to work in my fork.

On more Guzzle specific question for now: Could you please tell me how to best integrate this test within your existing test organization? Would a `CurlFactoryTest::testBodyEofOnWindows()` using `Server::enqueue()` with an appropriately sized response body be correct?

GrahamCampbell:
This should target master for 7.1.1 inclusion.

Nyholm:
Could you target master and rebase this PR?

TimWolla:
> Could you target master and rebase this PR?

@Nyholm Sure, can do. It doesn't rebase cleanly, so I will have to re-check that conflict resolution is good.

While I do so: Could you please tell me how I should go about adding a test for this change? See the second paragraph in: https://github.com/guzzle/guzzle/pull/2741#issuecomment-696907947

TimWolla:
Okay, I rebased onto master. The test that previously failed still fails: https://github.com/guzzle/guzzle/pull/2741#issuecomment-668513936. I still claim that the test is bogus, but I leave this evaluation up to you.

Nyholm:
I dont know the answers to your questions without looking into this further. 

Im find it hard to believe that a current test is wrong.

TimWolla:
> Im find it hard to believe that a current test is wrong.

The failing assertion was introduced here: https://github.com/guzzle/guzzle/commit/0a3065ea4639c1df8b9220bc8ca3fb529d7f8b52#diff-e61f3c0116b6df977a5a8e4c3df2ccb3R76

It previously was adjusted (without further comment) to remove one constant here: https://github.com/guzzle/guzzle/commit/15557b1523bf556a7d2ca8915a8a18cc261f06d4#diff-e61f3c0116b6df977a5a8e4c3df2ccb3L79-R79

I think it's overly specific and tests stuff that is not directly related to the purpose of the actual test.

Nyholm:
> Would a CurlFactoryTest::testBodyEofOnWindows() using Server::enqueue() with an appropriately sized response body be correct?

Try. If you find the bug on master and your patch fixes it then it is an excellent test.

Nyholm:
That test is verifying the HEAD request. 

I looked closer to it, it seams valid.

TimWolla:
> Try. If you find the bug on master and your patch fixes it then it is an excellent test.

Okay. Here is the proof that my test detects the bug: https://github.com/TimWolla/guzzle/pull/3/checks?check_run_id=1164888042.

You can see that my patch fixes the bug, because that new test does not fail in this PR.

> That test is verifying the HEAD request.
> 
> I looked closer to it, it seams valid.

I don't claim that the test itself is invalid. I claim that the `self::assertArrayNotHasKey($check, $_SERVER['_curl']);` assertion in the test tests an implementation detail.

**If** the test should actually test that implementation detail (which is a decision I leave up to you), then `\CURLOPT_FILE` should be added to that array, because it also is `unset()` here: https://github.com/guzzle/guzzle/blob/aec11522a3fafd7ab01017b4fa4fb87b4236b5b3/src/Handler/CurlFactory.php#L262-L267

Do you agree here?

tests/Handler/CurlFactoryTest.php
```diff
@@ -757,4 +757,32 @@ public function testRelease()
...
+        $promise = $handler($request, []);
+        $response = $promise->wait();
+        $body = $response->getBody();
+
+        $actualLength = 0;
+        while (!$body->eof()) {
+            $chunk = $body->read(4096);
+            $actualLength += \strlen($chunk);
+        }
+        self::assertSame($expectedLength, $actualLength);
```
Nyholm, line 786: Great!

Nyholm:
> If the test should actually test that implementation detail (which is a decision I leave up to you), then \CURLOPT_FILE should be added to that array, because it also is unset() here:

In the current master: No, 
In your PR: Hm.. we should never use the`CURLOPT_FILE`, right?

> **If** the test should actually test that implementation detail

You might be correct that we should not test that, But we need some way to test the curl options in a HEAD request.

TimWolla:
> In the current master: No,

@Nyholm Let's us discuss this in my new PR #2778 that adjusted the code + test. When that one is merged (or closed without merge) I'll rebase this Windows PR.

Nyholm:
I fail to understand why this PR includes a `CURLOPT_WRITEFUNCTION` on the HEAD request.

TimWolla:
> I fail to understand why this PR includes a `CURLOPT_WRITEFUNCTION` on the HEAD request.

Because it previously included `CURLOPT_FILE` on a HEAD request and this PR replaces the use of `CURLOPT_FILE` by `CURLOPT_WRITEFUNCTION`, because the use of `CURLOPT_FILE` is broken on Windows. `CURLOPT_FILE` is as useful / useless on a HEAD request as `CURLOPT_WRITEFUNCTION` is.

I recommend reviewing the diff with whitespace changes hidden. It will make the change *much* clearer.

Nyholm:
Thank you for having patience with me. 

I see now that we are unsetting before we deal with the sink stuff.

TimWolla:
> Hm.. I wonder if #2778 starts to make sense now.

I certainly believe it does. And the fact that #2778 is green reinforces that believe. In any case either the assertion should be removed (because it tests an implementation detail) or #2778 should be merged (because the test is incomplete).

Nyholm:
Could you please rebase this?

TimWolla:
Okay, rebased and green. Any change for a backport [of the relevant change](https://github.com/guzzle/guzzle/pull/2741/commits/2f9f5753b60466d75b5a13220fe56de1bcbf621d?w=1) to 6.5?

TimWolla:
Quoting myself from issue #2735:

> The application in question is hosted by our customers on-premise. We are using Guzzle 6.5.x, because we can't rely on PHP 7.2 being available on customer's webhosts. Please apply a fix to both Guzzle 6.5 as well as Guzzle 7.

I would like to avoid patching our local copy or pointing composer to a fork until we upgrade the minimum requirement to 7.2. But of course I don't want to add too much of a maintenance burden for you either :smiley:

Nyholm (approved):
Thank you.

Nyholm:
Regarding 6.5. 

Part of me want to put in 2-3 hours to test, merge and release this on 6.5. Especially since you did a loot of good work here. But given that 7.1 has not been actively supported since 2018. I feel like the better is to encourage people to upgrade their system. 

If they now going to upgrade now, then when?

TimWolla:
> But given that 7.1 has not been actively supported since 2018.

We try to anchor the supported PHP versions to the stock PHP versions of major Linux distributions. In this case the “blocker” is Debian Stretch (oldstable), shipping PHP 7.0, which is going to receive LTS support until 2022 and it was in active support back when the PHP version support for the current version of our software was decided. Technically there is also Ubuntu Xenial with PHP 7.0, but that one only lives until April 2021 and already got 2 subsequent LTS versions.

We plan to up the minimum PHP version with the next release, so I guess I'll just apply the patch to our local tree then. Thanks.

Nyholm:
Hm. I understand. 

If I had a magic wand, I would make sure that everybody understands the importance of not running out of date software.

PR #2761, opened 2020-09-19
Merge 6.5 -> 7.0

GrahamCampbell:
Part 2 of 3 of the 3rd bullet of https://github.com/guzzle/guzzle/pull/2712.

### DIFF
--- a/.github/workflows/static.yml
+++ b/.github/workflows/static.yml
@@ -16,7 +16,7 @@ jobs:
         uses: actions/checkout@v2
 
       - name: PHPStan
-        uses: OskarStark/phpstan-ga@0.12.28
+        uses: OskarStark/phpstan-ga@0.12.41
         env:
           REQUIRE_DEV: true
         with:
--- a/tests/Exception/RequestExceptionTest.php
+++ b/tests/Exception/RequestExceptionTest.php
@@ -130,21 +130,6 @@ class RequestExceptionTest extends TestCase
         self::assertStringEndsWith('response', $e->getMessage());
     }
 
-    public function testCreatesExceptionWithoutPrintableBody()
-    {
-        $response = new Response(
-            500,
-            ['Content-Type' => 'image/gif'],
-            $content = \base64_decode('R0lGODlhAQABAIAAAAAAAP///yH5BAEAAAAALAAAAAABAAEAAAIBRAA7') // 1x1 gif
-        );
-        $e = RequestException::create(new Request('GET', '/'), $response);
-        self::assertStringNotContainsString(
-            $content,
-            $e->getMessage()
-        );
-        self::assertInstanceOf(\GuzzleHttp\Exception\RequestException::class, $e);
-    }
-
     public function testHasStatusCodeAsExceptionCode()
     {
         $e = RequestException::create(new Request('GET', '/'), new Response(442));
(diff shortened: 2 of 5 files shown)


Nyholm:
Could you tell me why this is needed?

GrahamCampbell:
Because the tests fail with the new upcoming psr7 package, without this.

Nyholm:
Sorry. I still don’t quite understand. 

Are you saying tests on guzzle master fails with guzzke/psr master

GrahamCampbell:
Yes, with psr 1.x-dev.

Nyholm:
I don't think we should make more 7.0.x releases. Next release will be 7.1.0. 

I've added https://github.com/guzzle/guzzle/pull/2763 as a test with guzzle/psr7.

tests/Exception/RequestExceptionTest.php
```diff
@@ -130,21 +130,6 @@ public function testExceptionMessageIgnoresEmptyBody()
         self::assertStringEndsWith('response', $e->getMessage());
     }
 
-    public function testCreatesExceptionWithoutPrintableBody()
```
Nyholm, line 133: I've added an issue for this https://github.com/guzzle/guzzle/issues/2765

.github/workflows/static.yml
```diff
@@ -16,7 +16,7 @@ jobs:
         uses: actions/checkout@v2
 
       - name: PHPStan
-        uses: OskarStark/phpstan-ga@0.12.28
+        uses: OskarStark/phpstan-ga@0.12.41
```
Nyholm, line 19: I dont mind this, but as you can see in https://github.com/OskarStark/phpstan-ga/issues/31, we are already running 0.12.42 =)

Nyholm (changes requested):
I think this PR should be closed. 

We could consider use parts of it to fix https://github.com/guzzle/guzzle/issues/2765 but Im sure there are better solutions than just remove the test..

Also, we should not target 7.0 branch.

tests/Exception/RequestExceptionTest.php
GrahamCampbell: We already discussed this a few months ago and decided this was not a bug.

.github/workflows/static.yml
GrahamCampbell: 0.12.42 isn't available on docker hub.

GrahamCampbell: So I couldn't use it to generate the baseline locally.

GrahamCampbell:
I think we should just merge this, then I can do a 7.0 -> 7.1 merge, after.

GrahamCampbell:
Doesn't matter if there will not be another 7.0.x release. Indeed, there are already other commits on 7.0 that will never be tagged.

GrahamCampbell:
Will just target master then, and skip the 7.0 merge, since that seems to be what you want. :)

GrahamCampbell:
Makes little difference, other than the fact the github branch thing will show 6.5 as being ahead of 7.0, but not 7.1.

Nyholm:
If you just make a PR that will fix https://github.com/guzzle/guzzle/issues/2765, then I would be super happy =)

Issue #2765, opened 2020-09-20
Make the test pass with guzzle/psr7 version 1.7.0


Nyholm:
It has not yet been released but there are one test failing. See https://github.com/guzzle/psr7/pull/330

> There was 1 failure:
>
> 1) GuzzleHttp\Tests\Exception\RequestExceptionTest::testCreatesExceptionWithoutPrintableBody
Failed asserting that Binary String: 0x536572766572206572726f723a2060474554202f6020726573756c74656420696e2061206035303020496e7465726e616c20536572766572204572726f726020726573706f6e73653a0a47494638396101000100800000000000ffffff21f90401000000002c000000000100010000020144003b0a does not contain "GIF89a����!�,D;".
>
> /Users/tobias/Workspace/PHPStorm/guzzle/tests/Exception/RequestExceptionTest.php:141

GrahamCampbell:
We already discussed this a few months ago and decided this was not a bug. We just should delete the test.

Nyholm:
Okey. I miss that discussion (or I dont remember it). If that is the case, feel free to submit a PR to remove the test. 

Make sure we target master.

PR #2768, opened 2020-09-20
Merge 6.5 -> 7.1

GrahamCampbell:
This will fix https://github.com/guzzle/guzzle/issues/2765

### DIFF
--- a/tests/ClientTest.php
+++ b/tests/ClientTest.php
@@ -750,7 +750,7 @@ class ClientTest extends TestCase
         $client = new Client(['handler' => $mockHandler]);
 
         $this->expectException(\GuzzleHttp\Exception\InvalidArgumentException::class);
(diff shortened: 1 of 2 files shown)


tests/ClientTest.php
```diff
@@ -750,7 +750,7 @@ public function testExceptionOnInvalidIdn()
         $client = new Client(['handler' => $mockHandler]);
 
         $this->expectException(\GuzzleHttp\Exception\InvalidArgumentException::class);
-        $this->expectExceptionMessage('IDN conversion failed (errors: IDNA_ERROR_LEADING_HYPHEN)');
```
Nyholm, line 753: Could you remind me why this is needed?

GrahamCampbell: Because the polyfill has a different error message sometimes, and I think also the extension depending on how it was compiled. We are only interested in checking that IDN conversion failed, and not the exact string of the message.

Nyholm (approved):
Thank you

PR #2770, opened 2020-09-21
Remove composer normalize dependency

### LINKED ISSUE #2643: Remove composer normalize from dev deps and use a phar
The https://github.com/ergebnis/composer-normalize/releases now distributes a phar. When we have added that I think it was not.

I am not that familiar with GH actions to perform the change. Will try to learn it as well.. until then maybe someone can give it a shot.

sagikazarmark:
Fixes #2643


Nyholm (approved):
Thank you

gmponos (approved):

Issue #2265, opened 2019-02-06
StreamHandler calls `stream_set_timeout(false, ...)`  when 'read_timeout` is set and `fopen()` fails


zdenek-crha:
**Guzzle version(s) affected**: 6.3.3

**Description**  
StreamHandler::createStream() function does not handle 'read_timeout' optional configuration on `fopen()` call failure. It attempts to call `stream_set_timeout()` on `false` (boolean) value and triggers an error

**How to reproduce**  

Try to access api on port where nothing is listening.

**Possible solution**  

[fix_set_timeout_on_fopen_failure.txt](https://github.com/guzzle/guzzle/files/2835512/fix_set_timeout_on_fopen_failure.txt)

**Additional context**  

Error message:
```
Error creating resource: [message] fopen(http://example.com:/path/to/endpoint): failed to open stream: Connection refused
stream_set_timeout() expects parameter 1 to be resource, boolean given
```

Environmnent:
```
$ php --version
PHP 7.1.26-1+0~20190113101810.12+stretch~1.gbp7077bb (cli) (built: Jan 13 2019 10:18:10) ( NTS )
Copyright (c) 1997-2018 The PHP Group
Zend Engine v3.1.0, Copyright (c) 1998-2018 Zend Technologies
    with Zend OPcache v7.1.26-1+0~20190113101810.12+stretch~1.gbp7077bb, Copyright (c) 1999-2018, by Zend Technologies
    with Xdebug v2.6.1, Copyright (c) 2002-2018, by Derick Rethans

$ composer show | grep guzzle
guzzlehttp/guzzle                     6.3.3  Guzzle is a PHP HTTP client library
guzzlehttp/promises                   v1.3.1 Guzzle promises library
guzzlehttp/psr7                       1.4.2  PSR-7 message implementation that also provides common utility methods
```

gmponos:
This is also mentioned by PHPStan https://github.com/guzzle/guzzle/blob/master/phpstan-baseline.neon#L544

I would accept a PR about this..

PR #2772, opened 2020-09-23
Don't set stream timeout when fopen() call has failed (#2265)

zdenek-crha:
The connection to the remote crashes with cryptic error:

`stream_set_timeout() expects parameter 1 to be resource, boolean given`

when call to the `fopen()` method has failed for some reason. The root
cause is that code that tries to set timeout on the stream does not
check `fopen()` result for error indication.

Add check that we really do have valid resource before trying to set
timeout on it.

### DIFF
--- a/src/Handler/StreamHandler.php
+++ b/src/Handler/StreamHandler.php
@@ -341,7 +341,7 @@ class StreamHandler
                     );
                 }
 
-                if (isset($options['read_timeout'])) {
+                if (false !== $resource && isset($options['read_timeout'])) {
                     $readTimeout = $options['read_timeout'];
                     $sec = (int) $readTimeout;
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
That variable cannot be false there.

Nyholm:
This cannot be merged without a test.

zdenek-crha:
@GrahamCampbell Yeah, I see it now. I've had that patch sitting on disk since I've submitted the ticked. Since the original ticket is still open and patch applied cleanly, I did not do much checking.

I think this PR can be discarded and the #2265 could be closed since it has been fixed in different way.

GrahamCampbell:
Yes, this was already fixed in Guzzle 7.0.0. This PR can be closed.

Nyholm:
Thank you

PR #2773, opened 2020-09-23
Add Windows CI build

TimWolla:
As discussed in https://github.com/guzzle/guzzle/pull/2741#issuecomment-696907947 this adds Windows CI based on GitHub Actions. <strike>A single curl test fails. I did not yet look into it in detail. It might or might not be some limitation of the Windows platform.</strike>

I also added an uglyish workaround for https://github.com/php-http/client-integration-tests/issues/46 to get things started. That one should probably be cleaned up. It looks like @Nyholm also is a contributor to the client-integration-tests package, maybe he can help getting the script fixed?

Feel free to push into the windows-ci-clean branch of my fork. I left the "Allow edits by maintainers" checkbox enabled 😄

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -70,3 +70,44 @@ jobs:
 
       - name: Run tests
         run: make test
+
+  build-windows:
+    name: Build on Windows
+    runs-on: windows-latest
+    strategy:
+      max-parallel: 10
+      matrix:
+        php: ['7.2', '7.3', '7.4']
+
+    steps:
+      - name: Set up PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: ${{ matrix.php }}
+          coverage: none
+          extensions: mbstring, intl
+          tools: composer:v2
+
+      - name: Set up Node
+        uses: actions/setup-node@v1
+        with:
+          node-version: '14.x'
+
+      - name: Setup Problem Matchers for PHPUnit
+        run: echo "::add-matcher::${{ runner.tool_cache }}/phpunit.json"
+
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Download dependencies
+        run: composer update --no-interaction --prefer-dist
+
+      - name: Start test servers
+        shell: bash
+        run: |
+          node tests/server.js &
+          # See https://github.com/php-http/client-integration-tests/issues/46 for the background of the line below.
+          php -S 127.0.0.1:10000 -t "vendor/php-http/client-integration-tests/fixture" &
+
+      - name: Run tests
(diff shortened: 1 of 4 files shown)


TimWolla:
All green now :+1:

.github/workflows/ci.yml
```diff
@@ -70,3 +70,44 @@ jobs:
...
+      - name: Setup Problem Matchers for PHPUnit
+        run: echo "::add-matcher::${{ runner.tool_cache }}/phpunit.json"
+
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Download dependencies
+        run: composer update --no-interaction --prefer-dist
+
+      - name: Start test servers.
```
GrahamCampbell, line 105: remove the period

GrahamCampbell (changes requested):

.github/workflows/ci.yml
```diff
@@ -70,3 +70,44 @@ jobs:
...
+        shell: bash
+        run: |
+          node tests/server.js &
+          # See https://github.com/php-http/client-integration-tests/issues/46 for the background of the line below.
+          php -S 127.0.0.1:10000 -t "vendor/php-http/client-integration-tests/fixture" &
```
Nyholm, line 110: I dont think this is an ugly fix. I think this makes a lot of sense.

GrahamCampbell: Ugly in the sense it bypasses `make`?

Nyholm: ¯\_(ツ)_/¯ 

Whatever works =)

TimWolla: IMO ugly, because it fails to adapt to changes in php-http/client-integration-tests. Of course also ugly, because it bypasses make.

.github/workflows/ci.yml
TimWolla: Fixed, amended and force pushed.

GrahamCampbell (approved):

sagikazarmark (approved):
Thanks!

Keep in mind though, Microsoft dropped support for PHP, so new releases won't come out for Windows, only patch versions for existing ones.

GrahamCampbell:
> Keep in mind though, Microsoft dropped support for PHP, so new releases won't come out for Windows, only patch versions for existing ones.

That's not true. Microsoft have stopped offering support and resources to the PHP core team. PHP will continue to be supported on Windows by the community.

Nyholm (approved):
Thank yuo

sagikazarmark:
> PHP will continue to be supported on Windows by the community (in the same way it always was).

Ah, good to know.

PR #2775, opened 2020-09-25
Use fixed version of psalm

Nyholm:
3.14.2 was the first tagged version

### DIFF
--- a/psalm.baseline.xml
+++ b/psalm.baseline.xml
@@ -1,12 +1,12 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<files psalm-version="3.11.5@3c60609c218d4d4b3b257728b8089094e5c6c6c2">
+<files psalm-version="3.14.2@3538fe1955d47f6ee926c0769d71af6db08aa488">
   <file src="src/Client.php">
-    <NullArgument occurrences="1">
-      <code>null</code>
-    </NullArgument>
     <PossiblyUndefinedVariable occurrences="1">
       <code>$options</code>
     </PossiblyUndefinedVariable>
+    <UndefinedConstant occurrences="1">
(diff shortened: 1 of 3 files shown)


psalm.baseline.xml
```diff
@@ -1,12 +1,12 @@
...
+<files psalm-version="3.14.2@3538fe1955d47f6ee926c0769d71af6db08aa488">
   <file src="src/Client.php">
-    <NullArgument occurrences="1">
-      <code>null</code>
-    </NullArgument>
     <PossiblyUndefinedVariable occurrences="1">
       <code>$options</code>
     </PossiblyUndefinedVariable>
+    <UndefinedConstant occurrences="1">
+      <code>\IDNA_DEFAULT</code>
```
GrahamCampbell, line 8: The new docker image seems to mean we loose all static analysis of extensions not installed in the image... Not a problem, for undefined constants, but really unfortunately for all the function calls that are no longer analyzed...

GrahamCampbell (changes requested):

psalm.baseline.xml
Nyholm: Which function calls?

GrahamCampbell: All functions from extensions not installed in the image are no longer have their uses type-checked. Like ext-intl.

Nyholm: I only find this: https://github.com/guzzle/guzzle/blob/879a6c664fda9d33f274d7a9b3cd65ac2f1b62a6/src/Utils.php#L389

GrahamCampbell: Right, but I suppose my point was what other extensions are being skipped now, that we didn't notice?

Nyholm:
You are correct. There is no win using this PR.

PR #2776, opened 2020-09-25
Fix test randomly fails on Windows

Nyholm:
This test seams to fail 50% of the times because we cannot cleanup after ourselves.


GrahamCampbell:
Can we have a more reliable unlink?

Nyholm:
Maybe we could... but it is in temporary directory... I dont really care too much if the file is still there or not.

GrahamCampbell:
The issue arises when the tests are run more than once, though?

Nyholm:
Nope, we are selecting a new random file. 

Good question!

GrahamCampbell (approved):

Nyholm:
Thank yuo

PR #2778, opened 2020-09-25
Do not connect a sink for HEAD requests

TimWolla:
This PR was created in response to the test discussion in #2741: https://github.com/guzzle/guzzle/pull/2741#issuecomment-698808528

### DIFF
--- a/src/Handler/CurlFactory.php
+++ b/src/Handler/CurlFactory.php
@@ -393,29 +393,33 @@ class CurlFactory implements CurlFactoryInterface
             }
         }
 
-        if (isset($options['sink'])) {
-            $sink = $options['sink'];
-            if (!\is_string($sink)) {
-                $sink = \GuzzleHttp\Psr7\stream_for($sink);
-            } elseif (!\is_dir(\dirname($sink))) {
-                // Ensure that the directory exists before failing in curl.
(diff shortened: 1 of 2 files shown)


Nyholm (approved):
I agree with this. 

@GrahamCampbell do you have any comments?

(here is a better [diff](https://github.com/guzzle/guzzle/pull/2778/files?diff=unified&w=1))

GrahamCampbell:
Does a similar change need to be made to the stream handler?

TimWolla:
> Does a similar change need to be made to the stream handler?

It appears as if a check for HEAD requests is already implemented there, making the answer to your question “no”. That one uses strcasecmp, though: https://github.com/guzzle/guzzle/blob/7881628c7fd44078a4d33376fe9cef23dcd26b4f/src/Handler/StreamHandler.php#L120

Should I adjust CurlFactory to make use of `strcasecmp`?

TimWolla:
> Should I adjust CurlFactory to make use of `strcasecmp`?

As discussed with nyholm on php-http Slack: The use of `strcasecmp` technically is incorrect, because the HTTP method is case sensitive. That comparison will not be fixed, but I won't introduce any more case-insensitive comparisons.

Nyholm:
Thank you

PR #2787, opened 2020-09-30
6.x will only get security fixes

Nyholm:


gmponos (approved):

gmponos:
🙏

Nyholm:
Thanky ou

GrahamCampbell:
Wonder if that table should have a maximum PHP version in the bounds too. Only Guzzle 7 will run on PHP 8.

Nyholm:
I just checked. Guzzle 3, 4, 5 and 6 has no upper bound to the PHP version in composer.json. That was introduced in Guzzle 7.0

So if you ask packagist, then Guzzle 7 is the only one that CANNOT run on PHP 8. 

For that reason, I dont think we should add maximum PHP version in this table. Let's focus on PHP 8 support for Guzzle 7 instead.

GrahamCampbell:
Just because there is no upper bound, doesn't mean that we actually support those higher PHP versions. We should indicate to users what we know works, but not pretend Guzzle 4 works on PHP 8 when it doesn't.

PR #2794, opened 2020-10-01
Upgrade to phpstan 0.12.48

GrahamCampbell:


Nyholm:
We should figure out a rule for this.. I mean.. I dont want to merge PRs like this every time PHPStan/Psalm/PHP-cs-fixer releases a new version. At the same time I do want to lock the versions but not fall behind too much. 

Do you have any suggestions?

sagikazarmark:
I tend to upgrade tooling around releases (actually a couple days/a week before releases so I have time to fix bugs 😄 )

Unless there is a feature or bug that affects us, there is no need to update tooling every day.

Nyholm:
Okey. So let's make that the rule then. We should not update tooling unless there is a feature or bug that affect us.

Nyholm:
That rule takes to effect now =)

PR #2795, opened 2020-10-01
Support setting truncateBodyAt param in the http errors middleware

GrahamCampbell:
Closes https://github.com/guzzle/guzzle/pull/2272.

### DIFF
--- a/src/BodySummarizer.php
+++ b/src/BodySummarizer.php
@@ -1,0 +1,28 @@
+<?php
+
+namespace GuzzleHttp;
+
+use Psr\Http\Message\MessageInterface;
+
+final class BodySummarizer implements BodySummarizerInterface
+{
+    /**
+     * @var int|null
+     */
+    private $truncateAt;
+
+    public function __construct(int $truncateAt = null)
+    {
+        $this->truncateAt = $truncateAt;
+    }
+
+    /**
+     * Returns a summarized message body.
+     */
+    public function summarize(MessageInterface $message): ?string
+    {
+        return $this->truncateAt === null
+            ? \GuzzleHttp\Psr7\Message::bodySummary($message)
+            : \GuzzleHttp\Psr7\Message::bodySummary($message, $this->truncateAt);
+    }
+}
--- a/src/Exception/RequestException.php
+++ b/src/Exception/RequestException.php
@@ -56,18 +56,18 @@ class RequestException extends TransferException implements RequestExceptionInte
     /**
      * Factory method to create a new exception with a normalized error message
      *
-     * @param RequestInterface                         $request               Request sent
-     * @param ResponseInterface                        $response              Response received
-     * @param \Throwable|null                          $previous              Previous exception
-     * @param array                                    $handlerContext        Optional handler context
-     * @param null|callable(MessageInterface): ?string $bodySummarySummarizer Optional body summarizer
+     * @param RequestInterface                         $request        Request sent
+     * @param ResponseInterface                        $response       Response received
+     * @param \Throwable|null                          $previous       Previous exception
+     * @param array                                    $handlerContext Optional handler context
+     * @param null|callable(MessageInterface): ?string $bodySummarizer Optional body summarizer
      */
     public static function create(
         RequestInterface $request,
         ResponseInterface $response = null,
         \Throwable $previous = null,
         array $handlerContext = [],
-        callable $bodySummarySummarizer = null
+        ?callable $bodySummarizer = null
     ): self {
         if (!$response) {
             return new self(
@@ -105,9 +105,9 @@ class RequestException extends TransferException implements RequestExceptionInte
             $response->getReasonPhrase()
         );
 
-        $summary = $bodySummarySummarizer === null
+        $summary = $bodySummarizer === null
             ? \GuzzleHttp\Psr7\Message::bodySummary($response)
-            : $bodySummarySummarizer($response);
+            : $bodySummarizer($response);
 
         if ($summary !== null) {
             $message .= ":\n{$summary}\n";
--- a/src/Exception/RequestException.php
+++ b/src/Exception/RequestException.php
@@ -2,6 +2,8 @@
 
 namespace GuzzleHttp\Exception;
 
+use GuzzleHttp\BodySummarizer;
+use GuzzleHttp\BodySummarizerInterface;
 use Psr\Http\Client\RequestExceptionInterface;
 use Psr\Http\Message\MessageInterface;
 use Psr\Http\Message\RequestInterface;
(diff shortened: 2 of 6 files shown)


src/HandlerStack.php
```diff
@@ -44,10 +44,10 @@ class HandlerStack
      *                                                                          handler is provided, the best handler for your
      *                                                                          system will be utilized.
      */
-    public static function create(?callable $handler = null): self
+    public static function create(?callable $handler = null, array $options = []): self
```
Nyholm, line 47: Do we really need to add a second parameter here?

src/Exception/RequestException.php
```diff
@@ -102,13 +105,15 @@ public static function create(
             $response->getReasonPhrase()
         );
 
-        $summary = \GuzzleHttp\Psr7\Message::bodySummary($response);
+        $summary = $bodySummarizer === null
+            ? \GuzzleHttp\Psr7\Message::bodySummary($response)
+            : $bodySummarizer($response);
```
Nyholm, line 110: PHP-http has the concept of "message formatter" which would be ideal here. 

Im not a bug fan of using a callable without a technical contract (eg interface).

Nyholm (commented):
Thank you for working on this

src/Exception/RequestException.php
GrahamCampbell: I don't think Guzzle depends on php-http, intentionally so?

src/HandlerStack.php
GrahamCampbell: Yeh, so that options can be easily passed to middlewares. I anticipate we may want other options in the future, which is why I made this an array. Without this options array, if one wants to change the body summary, they'd have to create a stack themselves and add all the middlewares, rather than just passing an option to here.

src/Exception/RequestException.php
Nyholm: It doesn’t and it shouldn’t. But I like the concept. We could add a similar interface to guzzle.

src/HandlerStack.php
Nyholm: Let’s add it here when we have more and “other options”. 
At the moment, I’m happy with letting users recreate the stack themselves.

GrahamCampbell: It would already be useful for me. ;)

src/Exception/RequestException.php
GrahamCampbell: I mean, given that it is `callable(MessageInterface): string` and not just `callable`, I think it's fine, and an interface would be overkill, but I can add it if you really want.

src/HandlerStack.php
specialtactics: If we add it now, won't it obviously be better for forwards compatibility later?

specialtactics (approved):
Great work!

src/Exception/RequestException.php
Nyholm: Yeah. I believe that would be a good idea.

src/HandlerStack.php
sagikazarmark: I'm not a huge fan of this either. `HandlerStack::create` provides you a sane default stack. If you need this (in my opinion hideous) feature, create a factory function. It's literally 5 LOC.

src/Exception/RequestException.php
GrahamCampbell: We already have a different `MessageFormatterInterface`. I guess I should name this new one `BodySummarizerInterface`?

Nyholm: Yeah. Sounds good

src/BodySummarizer.php
```diff
@@ -0,0 +1,28 @@
+<?php
+
+namespace GuzzleHttp;
+
+use Psr\Http\Message\MessageInterface;
+
+final class BodySummarizer implements BodySummarizerInterface
```
sagikazarmark, line 7:
```suggestion
final class TruncateBodySummarizer implements BodySummarizer
```

Just an idea!

sagikazarmark (commented):
Thanks @GrahamCampbell

src/BodySummarizer.php
GrahamCampbell: I went with `BodySummarizerInterface` rather than `BodySummarizer` because of the naming conventions of the other interfaces in this package.

GrahamCampbell: NB This class does more than just truncate. It also may throw away bodies that cannot be rendered easily.

sagikazarmark: Well, I'd love to get rid of those interface suffixes 😄 

But I'm okay with delaying that to Guzzle 8 (if feasible at all)

src/Middleware.php
```diff
@@ -49,22 +49,24 @@ static function (ResponseInterface $response) use ($cookieJar, $request): Respon
      * Middleware that throws exceptions for 4xx or 5xx responses when the
      * "http_error" request option is set to true.
      *
+     * @param int|null $truncateBodyAt The length of the body to include in an exception message before truncation.
+     *
      * @return callable(callable): callable Returns a function that accepts the next handler.
      */
-    public static function httpErrors(): callable
+    public static function httpErrors(int $truncateBodyAt = null): callable
```
Nyholm, line 56: Let's use our `BodySummarizerInterface` interface here too.

src/Exception/RequestException.php
```diff
@@ -102,13 +106,13 @@ public static function create(
             $response->getReasonPhrase()
         );
 
-        $summary = \GuzzleHttp\Psr7\Message::bodySummary($response);
+        $summary = ($bodySummarizer ?? new BodySummarizer())->summarize($response);
```
Nyholm, line 109: I agree with this line because it keeps the same behavior. 

In Guzzle8 I would like to remove this fallback. If you don't provide a `BodySummarizerInterface` then you will not get a summary.

Nyholm (changes requested):
I think this looks way better now. Thank you. 

Just one more thing

src/Exception/RequestException.php
GrahamCampbell: Oh. In which case I'd advocate for changing the parameter order, perhaps, or maybe providing more than one create method. Like `createWithException` or something.

src/Middleware.php
GrahamCampbell: Yeh, that's fair enough now that we have the class that implements this.

Nyholm (approved):
Thank you. I think this looks good. 

@GrahamCampbell Is it ready to be merged?

GrahamCampbell:
@Nyholm Yep.

Nyholm:
Thank you

PR #2796, opened 2020-10-02
Break up MessageFormatter's format into multiple methods

jamesaspence:
The `MessageFormatter` has a large method, `format`, that handles formatting everything in a `preg_replace_callback` closure.

This PR breaks that up into separate methods for each formatting, as well as `handleFormatMatch`, which is the switch statement to handle ALL matches. This allows a developer to extend this class and override specific formatting or switches - i.e. they could change what formats are expected, or change how they're formatted.

### DIFF
--- a/src/MessageFormatter.php
+++ b/src/MessageFormatter.php
@@ -381,7 +381,8 @@ class MessageFormatter implements MessageFormatterInterface
     protected function formatUnexpectedMatch(
         string $match,
         RequestInterface $request,
-        ?ResponseInterface $response = null): string
+        ?ResponseInterface $response = null
+    ): string
     {
         // handle prefixed dynamic headers
         if (\strpos($match, 'req_header_') === 0) {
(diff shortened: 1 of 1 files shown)


mfn:
Hmm, the class was marked `@final` recently, only in phpdoc though => https://github.com/guzzle/guzzle/pull/2729/files#diff-2409f16b7e066e2ba170aacad6d078b7R36

jamesaspence:
ah, bummer - the reason I did this is my coworker and I recently wanted to implement logging but filter out some headers - in particular, `Authorization` and similar headers so they wouldn't be included in the logs. Doing that was way more tedious than expected, and would've been much easier if we could've just overrode `headers` or one of the formatting methods directly.

I didn't actually _change_ any code, just shuffled it around and separated it into separate methods.

Nyholm:
@mfn is correct. 

You should provide your own MessageFormatter object. That is both simpler and more correct.

PR #2797, opened 2020-10-04
Adding some highly opinionated CS fixes

Nyholm:
Basically: 
- If a method has 3 or less arguments, lets not do them on multiple lines. 
- Make sure lines are readable rather than short. 
- Make exceptions on one line

### DIFF
--- a/src/Cookie/CookieJarInterface.php
+++ b/src/Cookie/CookieJarInterface.php
@@ -36,10 +36,7 @@ interface CookieJarInterface extends \Countable, \IteratorAggregate
      * @param RequestInterface  $request  Request that was sent
      * @param ResponseInterface $response Response that was received
      */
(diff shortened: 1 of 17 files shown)


GrahamCampbell:
The rule I usually use is 120 characters is fine for single line, but behond that, make multiline. That is instead of "3 params on one line".

mfn:
No automation tools for this exist, do they?

GrahamCampbell:
Single line exceptions can be enforced by php-cs-fixer.

Nyholm:
I know that there is a rule like that, but we are currently not enforcing it. I just took the simple methods that were split into multiple lines and made them more readable.

TimWolla:
As a tiny data point: I agree with this change. It's much more readable.

Issue #2799, opened 2020-10-06
Regression in #2778 - Call to a member function getSize() on null


hedii:
**Guzzle version(s) affected**: 7.1.1
**PHP version**: 7.4.11 (hint: `php --version`)
**cURL version**: 7.72.0 (hint: `php -i | grep cURL`)

**Description**  
Since upgrading to 7.1.1 (i have ^7.0.1 in my composer.json), i get a lot of the same error on my log panel. It comes after making a head request with an `accept-encoding` header. I think #2778 is responsible for this error i did not get in previous guzzle versions (7.1.0 and 6.5.x)

```json
{
  "class": "Error",
  "message": "Call to a member function getSize() on null",
  "code": 0,
  "file": "/var/www/html/xxxxxx/vendor/guzzlehttp/guzzle/src/Handler/EasyHandle.php:84",
  "trace": [
    "/var/www/html/xxxxxx/vendor/guzzlehttp/guzzle/src/Handler/CurlFactory.php:580",
    "/var/www/html/xxxxxx/vendor/guzzlehttp/guzzle/src/Handler/CurlHandler.php:44",
    "/var/www/html/xxxxxx/vendor/guzzlehttp/guzzle/src/Handler/Proxy.php:32",
    "/var/www/html/xxxxxx/vendor/guzzlehttp/guzzle/src/Handler/Proxy.php:55",
    "/var/www/html/xxxxxx/vendor/guzzlehttp/guzzle/src/PrepareBodyMiddleware.php:35",
    "/var/www/html/xxxxxx/vendor/guzzlehttp/guzzle/src/Middleware.php:30",
    "/var/www/html/xxxxxx/vendor/guzzlehttp/guzzle/src/RedirectMiddleware.php:71",
    "/var/www/html/xxxxxx/vendor/guzzlehttp/guzzle/src/Middleware.php:60",
    "/var/www/html/xxxxxx/vendor/guzzlehttp/guzzle/src/HandlerStack.php:75",
    "/var/www/html/xxxxxx/vendor/guzzlehttp/guzzle/src/Client.php:330",
    "/var/www/html/xxxxxx/vendor/guzzlehttp/guzzle/src/Client.php:167",
    "/var/www/html/xxxxxx/vendor/guzzlehttp/guzzle/src/Client.php:186",
    "/var/www/html/xxxxxx/vendor/guzzlehttp/guzzle/src/ClientTrait.php:61",
    "..."
  ]
}
```

**How to reproduce**  
[...]

PR #2801, opened 2020-10-06
Fix handling of HEAD requests with Accept-Encoding header

TimWolla:
Fixes guzzle/guzzle#2799
Resolves guzzle/guzzle#2800

### DIFF
--- a/src/Handler/EasyHandle.php
+++ b/src/Handler/EasyHandle.php
@@ -78,6 +78,7 @@ final class EasyHandle
 
         if (!empty($this->options['decode_content'])
             && isset($normalizedKeys['content-encoding'])
+            && $this->request->getMethod() !== 'HEAD'
         ) {
             $headers['x-encoded-content-encoding']
                 = $headers[$normalizedKeys['content-encoding']];
--- a/tests/Handler/CurlFactoryTest.php
+++ b/tests/Handler/CurlFactoryTest.php
@@ -358,6 +358,25 @@ class CurlFactoryTest extends TestCase
         );
     }
 
+    /**
+     * https://github.com/guzzle/guzzle/issues/2799
+     */
+    public function testDecodesGzippedResponsesWithHeaderForHeadRequest()
+    {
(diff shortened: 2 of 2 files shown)


src/Handler/StreamHandler.php
```diff
@@ -114,7 +114,11 @@ private function createResponse(
         $status = (int) $parts[1];
         $reason = $parts[2] ?? null;
         $headers = Utils::headersFromLines($hdrs);
-        [$stream, $headers] = $this->checkDecode($options, $headers, $stream);
+
+        if (\strcasecmp('HEAD', $request->getMethod())) {
```
TimWolla, line 118: As discussed previously this is not correct, but I'm using `strcasecmp` for consistency with the check below.

Nyholm (commented):
I think this looks good. 

Currently we have some requests that never have a `sink`. In 7.0 we always had a `sink`. Would it be a good idea to go back to "always having a sink". I fear that we will have to do more patches like this in the future.

TimWolla:
> Would it be a good idea to go back to "always having a sink".

I'll leave this decision up to you maintainers. We don't currently have a need for HEAD requests and thus are not affected by "not having a sink". Just make sure to not break Windows :smiley:

Nyholm:
@GrahamCampbell what do you think?

GrahamCampbell:
I think always having a sync, at least for 7.x, is the most sensible thing to do.

TimWolla:
Just to make sure you are not waiting for me: I don't plan on making any more changes to this PR, because it would require too much of an effort. My understanding is that this PR does not fit your plan for 7.x. Please either:

- close this one and resolve the issue yourself (e.g. by reverting #2778)
- merge this PR as-is,
- or push into the `head-accept-encoding` branch of fork (it should allow for pushes from maintainers) and merge afterwards.

Nyholm:
That is fine. I'll continue this work within the next few days if nobody beats me to it. 

Thank you

PR #2804, opened 2020-10-09
Make sure the Request always has a body object

Nyholm:
This PR should not be squashed. It includes #2800 (from @hedii) and tests from #2801 (by @TimWolla). It will revert changes in #2778 but make sure it keeps working on Windows.


tests/ClientTest.php
```diff
@@ -805,4 +805,16 @@ public function testIdnWithRedirect()
...
     }
+
+    /**
+     * https://github.com/guzzle/guzzle/issues/2799
+     */
+    public function testGuzzleHeadRequestWithAcceptEncodingHeader(): void
+    {
+        $client = new Client(['headers' => ['accept-encoding' => 'gzip, deflate, br']]);
+
+        $response = $client->head('https://www.youtube.com/watch?v=kiHojsMTBeA');
```
TimWolla, line 816: I believe this tests makes an actual network request. I am not sure whether this is a good idea. My 2 tests should detect the same bugs as this one, but go to the test server instead and check additional assertions.

Nyholm: Oops =)

It is a good clip though.

Nyholm: Fixed

Nyholm: Hm.. no.. It does not show the bug now..hm..

TimWolla: The mock response needs to have a `content-encoding: gzip` response header.

Nyholm:
I removed the test in `ClientTest` as it is covered by `testDecodesGzippedResponsesWithHeaderForHeadRequest`

TimWolla (approved):
LGTM

tests/Handler/StreamHandlerTest.php
```diff
@@ -201,6 +201,24 @@ public function testAutomaticallyDecompressGzip()
...
+                'Content-Encoding' => 'gzip',
+                'Content-Length'   => \strlen($content),
+            ], $content)
+        ]);
+        $handler = new StreamHandler();
+        $request = new Request('HEAD', Server::$url);
+        $response = $handler($request, ['decode_content' => true])->wait();
+
+        // Verify that the content-length matches the encoded size.
+        self::assertTrue(!$response->hasHeader('content-length') || $response->getHeaderLine('content-length') == \strlen($content));
```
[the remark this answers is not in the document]
Nyholm: Im honestly not sure. This is just following a pattern. We do the same thing in `testAutomaticallyDecompressGzip` and `testDecodesGzippedResponsesWithHeader`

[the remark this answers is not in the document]
TimWolla: I re-used the implementation of `testAutomaticallyDecompressGzip`. Also most real-world responses with `content-encoding: gzip` will also be `transfer-encoding: chunked`, because the gzip compression is performed in a streaming fashion.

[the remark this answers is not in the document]
GrahamCampbell: I think we should make the tests as strict as possible.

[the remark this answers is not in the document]
Nyholm: I'll leave changing this to a separate PR. Then I can change it on 4(+?) places. 

I'll do it right away.

PR #2805, opened 2020-10-09
Make tests more strict

Nyholm:
Follow up from @GrahamCampbell comment: https://github.com/guzzle/guzzle/pull/2804#pullrequestreview-505676141


Nyholm:
Hm... It looks like the code dont like this change.

GrahamCampbell:
Why is there no content-length header?

Nyholm:
I think is is some weird stuff with the PSR-7 InflateStream.. but Im not sure.

PR #2806, opened 2020-10-09
Cover case where curl_multi_select return 0

### LINKED ISSUE #2390: curl_multi_select can return immediately ignoring timeout, with result 0 and error code 0

Nyholm:
This will fix #2390

### DIFF
--- a/src/Handler/CurlMultiHandler.php
+++ b/src/Handler/CurlMultiHandler.php
@@ -157,10 +157,15 @@ class CurlMultiHandler
         // Step through the task queue which may add additional requests.
         P\Utils::queue()->run();
 
-        if ($this->active && \curl_multi_select($this->_mh, $this->selectTimeout) === -1) {
-            // Perform a usleep if a select returns -1.
-            // See: https://bugs.php.net/bug.php?id=61141
-            \usleep(250);
+        if ($this->active) {
+            $fd = \curl_multi_select($this->_mh, $this->selectTimeout);
+            if ($fd === -1 || $fd === 0) {
+                // Perform a usleep if a select returns -1.
+                // See: https://bugs.php.net/bug.php?id=61141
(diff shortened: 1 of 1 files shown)


Nyholm:
Im not too confident in this fix. It is also a workaround for super old PHP version... I rather close this.

PR #2807, opened 2020-10-09
Soft mark CURL_VERSION_STR as private

GrahamCampbell:
To be made private in in 8.0, or just removed and inlined. D'no why this constant was bothered with?


GrahamCampbell:
Suppose this was so the context array key could be known.

PR #2812, opened 2020-10-19
Update docs to reflect correct exception when using timeout option

bpedroza:
Per https://github.com/guzzle/guzzle/blob/master/UPGRADING.md 
>Class GuzzleHttp\Exception\ConnectException now extends GuzzleHttp\Exception\TransferException instead of GuzzleHttp\Exception\RequestException.

Updating docs to reflect this.


Nyholm (approved):
Thank you

PR #2813, opened 2020-10-23
Correct tree of exceptions

roenschg:
The ConnectException extends the TransferException not the RequestException


Nyholm (approved):
Thank you

PR #2820, opened 2020-11-15
Removed outdated version number.

joachim-n:
Having the version number here is superfluous (because it's implicit from the branch, and generated docs pages will use the branch to tell the user the version they are reading about) and adds to maintainability burden (it's wrong in the docs for Guzzle 7!).


Nyholm (approved):
Thank you.

PR #2821, opened 2020-11-16
Fix progress aborted by callback

fortuneman1:


GrahamCampbell:
@fortuneman1 Thanks for the PR. What does this fix? Can you include some tests?

PR #2825, opened 2020-11-30
Update CI, use better integration with psalm, phpstan and php-ci-fixer (also fix one deprecated call)

### LINKED ISSUE #2824: Call to undefined function GuzzleHttp\\Psr7\\stream_for()
**Guzzle version(s) affected**: 7.2.0
**PHP version**: 7.4.12
**cURL version**: 7.73.0

**Description**  
It looks like the recent switch to static util functions in 7.2.0 (https://github.com/guzzle/guzzle/pull/2712) missed a call in
https://github.com/guzzle/guzzle/blob/master/src/Handler/CurlFactory.php#L400

**How to reproduce**  
The request is created deep within another library I'm using so I'm not sure how to make a minimal reproduction.

**Possible Solution** 
Within GuzzleHttp/Handler/CurlFactory.php, change:
```php
if (!\is_string($sink)) {
    $sink = \GuzzleHttp\Psr7\stream_for($sink);
} elseif (!\is_dir(\dirname($sink))) {
```
to:
```php
if (!\is_string($sink)) {
    $sink = \GuzzleHttp\Psr7\Utils::streamFor();
} elseif (!\is_dir(\dirname($sink))) {
```
I'm not sure what the purpose of the $sink parameter is and how it would convert to the new syntax, otherwise I'd make a pull request.

Thanks!

GrahamCampbell:
And re-worked static analysis so that this kind of error cannot happen again.

---

Fixes #2824.

### DIFF
--- a/.github/workflows/static.yml
+++ b/.github/workflows/static.yml
@@ -21,7 +21,6 @@ jobs:
           php-version: '7.4'
           coverage: none
           extensions: mbstring, intl
-          tools: composer
 
       - name: Download dependencies
         run: composer update --no-interaction --no-progress
@@ -46,7 +45,6 @@ jobs:
           php-version: '7.4'
           coverage: none
           extensions: mbstring
-          tools: composer
 
       - name: Download dependencies
         run: composer update --no-interaction --no-progress
@@ -71,7 +69,6 @@ jobs:
           php-version: '7.4'
           coverage: none
           extensions: mbstring, intl
-          tools: composer
 
       - name: Download dependencies
         run: composer update --no-interaction --no-progress
--- a/.github/workflows/static.yml
+++ b/.github/workflows/static.yml
@@ -15,12 +15,22 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v2
 
-      - name: PHPStan
-        uses: docker://oskarstark/phpstan-ga:0.12.48
-        env:
-          REQUIRE_DEV: true
+      - name: Setup PHP
+        uses: shivammathur/setup-php@v2
         with:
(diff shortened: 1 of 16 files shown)


Nyholm (commented):
There is a lot of unrelated CI changes here too. 

Could you limit the changes in this PR to only fix the deprecated function call?

GrahamCampbell:
The reason for the CI changes was to prevent this error from being re-introduced, or similar errors.

.github/workflows/static.yml
```diff
@@ -30,31 +40,44 @@ jobs:
...
-          args: --dry-run --diff-format udiff
+          php-version: '7.4'
+          coverage: none
+          extensions: mbstring
+          tools: composer
+
+      - name: Download dependencies
+        run: composer update --no-interaction --no-progress
+
+      - name: Download PHPStan
```
Nyholm, line 54: Phpstan?

GrahamCampbell: fixed ;)

.github/workflows/static.yml
```diff
@@ -15,12 +15,22 @@ jobs:
...
-        env:
-          REQUIRE_DEV: true
+      - name: Setup PHP
+        uses: shivammathur/setup-php@v2
         with:
-          args: analyze --no-progress
+          php-version: '7.4'
+          coverage: none
+          extensions: mbstring, intl
+          tools: composer
```
Nyholm, line 24: This is not needed.

Nyholm (approved):

Nyholm (approved):
Thank you

PR #2826, opened 2020-12-03
options json encode use options

wuchenge:
add json_encode options like JSON_UNESCAPED_UNICODE

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -369,7 +369,13 @@ class Client implements ClientInterface, \Psr\Http\Client\ClientInterface
         }
 
         if (isset($options['json'])) {
-            $options['body'] = Utils::jsonEncode($options['json']);
+            $json_type = 0;
+            if (isset($options['json_type'])) {
+                $json_type = $options['json_type'];
+                unset($options['json_type']);
+            }
+            $options['body'] = Utils::jsonEncode($options['json'], $json_type);
+            unset($json_type);
             unset($options['json']);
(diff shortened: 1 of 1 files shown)


Nyholm:
Thank you for this PR. 

I dont know why this is needed. I am also missing a test.

GrahamCampbell:
Closing due to inactivity.

wuchenge:
> Thank you for this PR.
> 
> I dont know why this is needed. I am also missing a test.

> use Chinese or other language

wuchenge:
use Chinese or other language

src/Client.php
```diff
@@ -369,7 +369,13 @@ private function applyOptions(RequestInterface $request, array &$options): Reque
         }
 
         if (isset($options['json'])) {
-            $options['body'] = Utils::jsonEncode($options['json']);
+            $json_type = 0;
+            if (isset($options['json_type'])) {
```
mfn, line 373: At the very least it should probably be called `json_encoding_options`

Since the encoder here is not configurable, I could see the value having _another_ configuration for this.

However, if it's so important, users might set `body` themselves (though it comes with missing convenience of automatically setting the content-type).

🤷‍♀️

wuchenge: users set body themselves with json_encode with json_type then could't use optioins['json'].

LiJiaHeer (commented):
agree

PR #2834, opened 2020-12-22
fix: ensure 5.3 cannot be installed for PHP 8.0

bshaffer:
See https://github.com/guzzle/guzzle/pull/2833

guzzle 5.3 is not compatible with PHP 8.0. Update `composer.json` to ensure this is taken into account by composer.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,7 @@
         }
     ],
     "require": {
-        "php": "^5.4||^7.0",
+        "php": "^5.4|^7.0",
         "guzzlehttp/ringphp": "^1.1",
         "react/promise": "^2.2"
     },
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,7 @@
         }
     ],
     "require": {
-        "php": ">=5.4.0",
+        "php": "^5.4||^7.0",
         "guzzlehttp/ringphp": "^1.1",
         "react/promise": "^2.2"
     },


GrahamCampbell:
This PR won't work, because composer can just resolve the previous version of the 5.3.x release which says it works on PHP 8.

GrahamCampbell:
Moreover, Guzzle 5 is EOL anyway.

bshaffer:
It _will work_, as libraries can use `^5.3.5`, and `--prefer-lowest` would then not install `5.3.5` on PHP 8.0 as it's been properly marked as incompatible.

I understand Guzzle 5 is EOL, but I imagine it still receives lots of downloads, and this change would be useful for some people hoping to ensure the proper version is installed and it'd be nice to fix the bug.

But no hard feelings if you'd prefer not to make this change.  I added the following workaround to my github actions:

```
- if: ${{ matrix.php == '8.0' || matrix.composer-flags == '--prefer-lowest' }}
   name: Update guzzlehttp/ringphp dependency          
```

I'd prefer not to need this, however. Thanks for your time!

GrahamCampbell:
I don't think many libraries will do that. Indeed, nobody should still be using Guzzle 5 anyway. It will not be getting any security fixes.

GrahamCampbell:
And even if a library does that, older versions of the library exist... This is why upper bounds on dependencies are so important, since they cannot be retroactively fixed.

GrahamCampbell:
While, Guzzle 7 has taken care to use upper bounds, other major projects are actually deleting them, so this problem with prefer-lowest is going to become a real issue for people using phpunit and symfony in the future!

bshaffer:
That's unfortunate, I don't know why anyone would see that as a good idea. Define the versions your library works with accurately! 

Using `--prefer-lowest` is a useful way to test multiple major versions of dependencies, but there are other ways as well. If y'all are refusing security fixes for this branch then it makes sense to not merge. However, I might suggest removing the `5.3` branch, as to me that implies it's still accepting PRs.

Nyholm:
Thank you for this PR. 

I understand your situation but I don't think we should make more effort into 5.x. It only gets a few thousands downloads per month since 2017. 

I dont think we can remove the 5.3 branch because we have tags to it. I think it will make those tags uninstallable... I might be wrong but Im not interested to take a chance =)

bshaffer:
@Nyholm FYI removing a branch definitely doesn't remove the git tags

Nyholm:
Thank you for letting me know.

PR #2837, opened 2020-12-23
Use RequestOptions constants in Client

### LINKED ISSUE #2836: Use RequestOptions constants in Client
**Description**
This is a suggestion for tidying up the code a little. It's not a bug report, possibly not a feature request, but closer to a feature request than a bug report.

`GuzzleHttp\Client` in many places references `$options` array values with their literal keys instead of using defined `GuzzleHttp\RequestOptions` constants.

**Example**
Use `$options[RequestOptions::MULTIPART]` instead of `$options['multipart']`.

**Additional context**
I was looking through the Guzzle code to understand how multipart request bodies are created. Whilst doing so, I was confused as to why RequestOptions::MULTIPART exists but is not used in places where it looks like it should/could do.

I'm happy to create a PR to tidy this up, just documenting for now for my own sanity!

`GuzzleHttp\Client` contains the following references to options array keys literally where in all cases a constant is defined for the same:

- `$options['auth']`
- `$options['body']`
- `$options['decode_content']`
- `$options['form_params']`
- `$options['headers']`
- `$options['json']`
- `$options['multipart']`
- `$options['query']`
- `$options['sink']`    
- `$options['version']`

The above can be replaced with:

- `$options[RequestOptions::AUTH]`
- `$options[RequestOptions::BODY]`
- `$options[RequestOptions::DECODE_CONTENT]`
- `$options[RequestOptions::FORM_PARAMS]`
- `$options[RequestOptions::HEADERS]`
- `$options[RequestOptions::JSON]`
- `$options[RequestOptions::MULTIPART]`
- `$options[RequestOptions::QUERY]`
- `$options[RequestOptions::SINK]`    
- `$options[RequestOptions::VERSION]`

webignition:
Replacing `GuzzleHttp\Client` occurrences of `$options[<key>]` with `$options[RequestOptions::*]`.

Fixes #2836

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -376,12 +376,12 @@ class Client implements ClientInterface, \Psr\Http\Client\ClientInterface
             $options['_conditional']['Content-Type'] = 'application/json';
         }
 
-        if (!empty($options['decode_content'])
-            && $options['decode_content'] !== true
+        if (!empty($options[RequestOptions::DECODE_CONTENT])
+            && $options[RequestOptions::DECODE_CONTENT] !== true
         ) {
             // Ensure that we don't have the header in different case and set the new value.
             $options['_conditional'] = Psr7\Utils::caselessRemove(['Accept-Encoding'], $options['_conditional']);
-            $modify['set_headers']['Accept-Encoding'] = $options['decode_content'];
+            $modify['set_headers']['Accept-Encoding'] = $options[RequestOptions::DECODE_CONTENT];
         }
 
         if (isset($options[RequestOptions::BODY])) {
--- a/src/Client.php
+++ b/src/Client.php
@@ -414,8 +414,8 @@ class Client implements ClientInterface, \Psr\Http\Client\ClientInterface
             }
         }
 
-        if (isset($options['query'])) {
-            $value = $options['query'];
+        if (isset($options[RequestOptions::QUERY])) {
+            $value = $options[RequestOptions::QUERY];
             if (\is_array($value)) {
                 $value = \http_build_query($value, '', '&', \PHP_QUERY_RFC3986);
             }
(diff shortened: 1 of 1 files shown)


webignition:
Assuming these changes are valid, I'm not sure if `GuzzleHttp\Client` lines 230 to 235 should also be updated. Those lines look to use literals that should be replaced by `RequestOptions` constants but I can't be sure it is not just a case of coincidental naming.

src/Client.php
```diff
@@ -343,57 +343,57 @@ private function applyOptions(RequestInterface $request, array &$options): Reque
...
             // Ensure that we don't have the header in different case and set the new value.
             $options['_conditional'] = Psr7\Utils::caselessRemove(['Content-Type'], $options['_conditional']);
             $options['_conditional']['Content-Type'] = 'application/x-www-form-urlencoded';
         }
 
-        if (isset($options['multipart'])) {
-            $options['body'] = new Psr7\MultipartStream($options['multipart']);
+        if (isset($options[RequestOptions::MULTIPART])) {
+            $options[RequestOptions::BODY] = new Psr7\MultipartStream($options[RequestOptions::MULTIPART]);
             unset($options['multipart']);
```
webignition, line 368: This occurrence of `$options['multipart']` has been intentionally left unchanged.

If replaced with `$options[RequestOptions::MULTIPART]` (which is equivalent as `RequestOptions::MULTIPART` === `multipart`), psalm fails with the error:

```
::error file=src/Client.php,line=450,col=22::Cannot iterate over string
```

I can't see any meaningful reason why this failure should be happening.

GrahamCampbell:
Related: https://github.com/guzzle/guzzle/pull/2683.

GrahamCampbell:
Thanks for all your hard work here @webignition, but we've decided to close this PR.

Issue #1656, opened 2016-11-15
Redirect not working when there's no domain in 'Location' header


gytislakavicius:
When a response contains 'Location' header which doesn't have a domain (e.g. '/not/working/properly.jpg'), Guzzle fails to build a correct Uri. Instead of using a domain from original request it throws an exception:
`Unable to parse URI: /not/working/properly.jpg`

On a quick look Psr7\Uri::resolve does some Uri merging if two UriInterface's are passed to it. So I think that if on \GuzzleHttp\RedirectMiddleware:213, UriInterface would be passed instead of naked 'Location' header string - it should work as supposed. Not 100% sure about that though.

sagikazarmark:
This is indeed strange, because we even have a test case for that: https://github.com/guzzle/psr7/blob/master/tests/UriResoverTest.php#L73

haelbichalex:
Did somebody find a fix for this?
my url should be `example.com/test.html` and I get `example.comtest.html`, when there is no full domain in Location header.

Nokim:
Did you tried to set 'base_uri' in client config?

algorythm:
Did you ever work on this? I have the exact same issue. When a website redirects to an absolute path, it fails.

Example: Let's say that a request to `http://example.com` returns a 301 to `/us/en`, then GuzzleHttp throws an exception with message:

> Redirect URI, /us/en, does not use one of the allowed redirect protocols: http, https

mrimann:
Just stumbled over the same issue:

```
➜  ~ curl -I http://www.somevaliddomain.tld
HTTP/1.1 302 Moved Temporarily
Date: Mon, 24 Feb 2020 15:13:57 GMT
Server: Apache
Connection: close
X-Powered-By: pimcore
Location: /de
X-Powered-By: PleskLin
Content-Type: text/html; charset=UTF-8
```

Calling that domain in the browser via "http://www.somevaliddomain.tld" results in being forwarded to "http://www.somevaliddomain.tld/de" - but Guzzle throws the following error at me:
```
Redirect URI, /de, does not use one of the allowed redirect protocols: http, https
```

My Guzzle-Client is not using a Base-URI, has the option `'allow_redirects' => true` set upon initialization - and the request is done against a domain.

Expected behaviour would be like in the Browser - Guzzle shall follow the `/de` redirection and interpret it as `somevaliddomain.tld/de` in this case.

mrimann:
Problem still happening.

[...]

PR #2838, opened 2020-12-31
Adding protocol when missing from the url

alex-moreno:
As per issue https://github.com/guzzle/guzzle/issues/1656

### DIFF
--- a/src/RedirectMiddleware.php
+++ b/src/RedirectMiddleware.php
@@ -5,6 +5,7 @@ namespace GuzzleHttp;
 use GuzzleHttp\Exception\BadResponseException;
 use GuzzleHttp\Exception\TooManyRedirectsException;
 use GuzzleHttp\Promise\PromiseInterface;
+use GuzzleHttp\Psr7\Uri;
 use Psr\Http\Message\RequestInterface;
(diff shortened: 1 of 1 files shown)


alex-moreno:
um, not sure how to solve this

```Argument 1 of strpos expects string, Psr\Http\Message\UriInterface provided with a __toString method```

alex-moreno:
> um, not sure how to solve this
> 
> `Argument 1 of strpos expects string, Psr\Http\Message\UriInterface provided with a __toString method`

actually fixed

GrahamCampbell:
Is omitting the protocol even permitted by the HTTP spec? If not, I don't think we should support it, and people need to fix their server implementations.

alex-moreno:
> Is omitting the protocol even permitted by the HTTP spec? If not, I don't think we should support it, and people need to fix their server implementations.

Good point

Nyholm (changes requested):
what if Im redirected to `tcp://` or ´foobar://´

I vote for closing this.

GrahamCampbell:
I agree with @Nyholm.

GrahamCampbell:
Thanks for your time preparing this PR, but we don't feel that it is something we should support.

PR #2845, opened 2021-01-28
Remove curl auth on cross-domain redirects

### LINKED ISSUE #1477: Problem with 302 redirect and Auth on initial (but not redirected) link
I am downloading a file with Guzzle. The original url, via https, requires Digest authorization, then returns a 302 redirect to another url on a different domain (http) that does NOT require authorization.

When I try to use Guzzle to download the file following redirects automatically, Guzzle makes the initial connection, gets back a 401, sends again with the correct authorization headers, gets back the 302 redirect--all good. 

Then it tries to send the same exact Authorization headers with the request to the new url, which gets back a 400 bad request, I assume because the redirected URL does not expect authorization.

I have gotten around this by capturing the redirect url via an ```on_stats``` function, setting ```allow_redirects``` to false, and creating a completely separate request for the redirected url. But it seems that there should be a better way to handle this.

kkopachev:
Fixes #1477

### DIFF
--- a/src/RedirectMiddleware.php
+++ b/src/RedirectMiddleware.php
@@ -88,6 +88,16 @@ class RedirectMiddleware
         $this->guardMax($request, $response, $options);
         $nextRequest = $this->modifyRequest($request, $options, $response);
 
+        // If authorization is handled by curl, unset it if host is different.
+        if ($request->getUri()->getHost() !== $nextRequest->getUri()->getHost()
+            && defined('\CURLOPT_HTTPAUTH')
+        ) {
+            unset(
+                $options['curl'][\CURLOPT_HTTPAUTH],
+                $options['curl'][\CURLOPT_USERPWD]
+            );
+        }
+
         if (isset($options['allow_redirects']['on_redirect'])) {
             ($options['allow_redirects']['on_redirect'])(
                 $request,
--- a/tests/RedirectMiddlewareTest.php
+++ b/tests/RedirectMiddlewareTest.php
@@ -301,6 +301,64 @@ class RedirectMiddlewareTest extends TestCase
     }
 
     /**
+     * @testWith ["digest"]
+     *           ["ntlm"]
+     */
+    public function testRemoveCurlAuthorizationOptionsOnRedirect($auth)
+    {
+        if (!defined('\CURLOPT_HTTPAUTH')) {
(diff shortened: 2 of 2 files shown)


TimWolla (approved):
As a non-maintainer that eyeballed the diff this LGTM.

GrahamCampbell:
Do we actually want to support someone circumventing Guzzle's abstraction, and setting auth headers directly through curl options?

TimWolla:
@GrahamCampbell My understanding is that this is related to this: https://github.com/guzzle/guzzle/blob/3a0543e4f589e21e8d65b984e544967c9de2b8b2/src/Client.php#L405-L413

GrahamCampbell:
Ah, I see that "todo" is what needs resolving.

kkopachev:
I looked at the todo item before submitting this PR and resolved to submitting PR anyway. I have 2 reasons for that:
* This PR fixes issue. Probably not the ideal way for ideal world, but it fixes it now.
* It is 6 years since this [todo was first created](https://github.com/guzzle/guzzle/commit/1a9ad6b55368283aaf73672a519e5177530991fb#diff-40d7473316ef0bb3ba4c4d937b16ae3d1cc47e0d65815034ee2b5be6a74cb11eR445). Digest auth provide no advantages over https+basic auth and therefore not so much popular and usage would probably decline over time. NTLM auth is [explicitly disabled](https://github.com/guzzle/guzzle/blob/master/src/Handler/StreamHandler.php#L292-L295) at stream handler because such auth requires keep-alive connection and [stream handler](https://github.com/guzzle/guzzle/blob/3a0543e4f589e21e8d65b984e544967c9de2b8b2/src/Handler/StreamHandler.php#L256-L262) [closes stream](https://github.com/guzzle/guzzle/blob/3a0543e4f589e21e8d65b984e544967c9de2b8b2/src/Handler/StreamHandler.php#L204) after each request. 

tl;dr: a fix is better than no fix for issue being open for so long with low chances of proper fix.

kkopachev:
Is there anything else I could do for this PR?

GrahamCampbell (approved):

kkopachev:
ping? Somebody wants to re-run all checks and merge it in?

kkopachev:
Rebased on latest master. Please take a look.

Nyholm (approved):
Thank you

Issue #1885, opened 2017-07-26
Files are not sent (using 'multipart' when header 'Content-Type' is set manually)


jeromejaglale:
| Q            | A
| ------------ | ---
| Bug?         | not sure
| Version      | 6.3.0

#### Actual Behavior

When using 'multipart', the form data is not sent when the global header 'Content-Type' is manually set to 'multipart/form-data'.

#### Possible Solutions

Mention in the documentation, under the 'multipart' section, that 'Content-Type' shouldn't be set manually.

tomzx:
Do you mean you're doing something like the following request?

```php
$response = $client->request('POST', 'upload.php', [
    'headers' => [
        'Content-Type' => 'multipart/form-data',
    ],
    'multipart' => [
        [
            'name'     => 'test',
            'contents' => fopen('index.php', 'r'),
        ]
    ],
]);
```

Files are still sent to the remote server (as observed by outputting `$_SERVER['CONTENT_LENGTH']`), but the server will generate the following warning `Warning: Missing boundary in multipart/form-data POST data in Unknown on line 0`.

mubat:
I think, this is because the client does not attach the value of the boundary to `Content-Type` header.
Request from guzzle came with `'Content-Type' => 'multipart/form-data',` only, but from browser header is `'Content-Type' => 'multipart/form-data; boundary=-----44cf242ea3173cfa0b97f80c68608c4c' for example request:
```http
-----44cf242ea3173cfa0b97f80c68608c4c
Content-Disposition: form-data; name="image_01"; filename="lorem-ipsum.png"
Content-Type: image/png; charset=binary

�PNG
���...etc etc...
���,
-----44cf242ea3173cfa0b97f80c68608c4c
Content-Disposition: form-data; name="testkey"

testvalue
-----44cf242ea3173cfa0b97f80c68608c4c
Content-Disposition: form-data; name="otherkey"

othervalue
-----44cf242ea3173cfa0b97f80c68608c4c--

```

kvnvandenbrande:
I spent a day looking for a solution this issue.. The removal the manual header was the solution. The suggested solution from @jeromejaglale would have helped me much faster in finding the problem!

crjaCC:
I just had this same issue.

thucpug:
can you show code post image by form data as your reply

ksgush1:
Wow! Yes a hint in the docs would have saved a ton of time.

roelywoely:
[...]

PR #2860, opened 2021-02-26
Add a note to not set the content-type manually

Cryt1c:
This PR adds a note in the docs that the Content-Type must not be set manually on multipart/form-data requests.
https://github.com/guzzle/guzzle/issues/1885

### DIFF
--- a/docs/request-options.rst
+++ b/docs/request-options.rst
@@ -636,7 +636,7 @@ multipart
 
 .. note::
 
-    Do not set the `Content-Type` manually when you are using `multipart/form-data`
+    Do not set the `Content-Type` manually when you are using `multipart/form-data`.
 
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
You could set it manually if you want, but you'd have to make sure you get it right.

GrahamCampbell:
`'multipart/form-data'` is not the correct content type.

GrahamCampbell:
The correct content type is described in the issue you linked to.

Cryt1c:
Thank you for your Feedback @GrahamCampbell ! 
This is the correct manual Content-Type, right?
`'Content-Type' => 'multipart/form-data; boundary=-----44cf242ea3173cfa0b97f80c68608c4c'`
Shall we put a note of this into the docs?

Issue #2863, opened 2021-03-05
[Bug] MessageFormatter makes the response body empty


root-aza:
**Guzzle version(s) affected**:  7.2.0
**PHP version**: 8.0.2
**cURL version**: 7.74.0

**Description**  
When use the template {response} for the MessageFormatter with the built-in logger midleware Middleware::Log, the response body becomes empty.

**How to reproduce**  
```php
<?php

use GuzzleHttp\Client;
use GuzzleHttp\Psr7\Request;

$client   = new Client();
$request  = new Request('GET', 'https://jsonplaceholder.typicode.com/todos/1');
$response = $client->sendRequest($request);

$formatter = new \GuzzleHttp\MessageFormatter('{response}');
$message   = $formatter->format($request, $response);

if ($response->getBody()->getContents() === '') { // WTF???
    echo 'empty body' . \PHP_EOL;
    echo $message . \PHP_EOL;
}
```

**Possible Solution**  
Rewind the stream at retrieving response content.

Nyholm:
I know this is super weird, however, it is not wrong. `getContents()` says: "Print the rest of the stream". You are making an assumption that the stream pointer is at the beginning. There is no place where PSR7 or Guzzle promise the position of the pointer. 

You should always use `$response->getBody()->__toString()`

root-aza:
Oh. Sory. Then I close the issue.

Nyholm:
No worries. As I said: This is weird and super confusing.

PR #2864, opened 2021-03-05
[Fix] MessageFormatter makes the response body empty

root-aza:
[Issue #2863](https://github.com/guzzle/guzzle/issues/2863)


Nyholm:
Im 👎 

https://github.com/guzzle/guzzle/issues/2863#issuecomment-792247910

PR #2867, opened 2021-03-07
Remove else if and put early return on choose handler

Calison25:
I was pass for a problem with guzzle, and try to figure out was doing and find this method, so I think he can be a little better, and remove else if and put early return for make it more readable and faster.


docker-compose.yml
```diff
@@ -0,0 +1,14 @@
+version: '3.6'
```
GrahamCampbell, line 1: Please don't add this unrelated change.

GrahamCampbell (changes requested):

GrahamCampbell:
What does this fix, please? :)

docker-compose.yml
Calison25: sorry, my mistake with docker-compose.

Calison25:
> What does this fix, please? :)

Nothing, only make method more readable, with early return, I don't find bug in this method, only try to make more readable.

src/Utils.php
```diff
@@ -90,26 +90,38 @@ public static function debugResource($value = null)
...
-        } elseif (!$handler) {
+        if (!$handler) {
             throw new \RuntimeException('GuzzleHttp requires cURL, the allow_url_fopen ini setting, or a custom HTTP handler.');
         }
 
         return $handler;
     }
 
+    /**
+     * @return callable|CurlHandler|CurlMultiHandler|null
```
GrahamCampbell, line 106: wrong phpdoc. should be `(callable(\Psr\Http\Message\RequestInterface, array): \GuzzleHttp\Promise\PromiseInterface)|null`

GrahamCampbell (changes requested):

src/Utils.php
```diff
@@ -90,26 +90,38 @@ public static function debugResource($value = null)
...
+        if (\ini_get('allow_url_fopen')) {
+            return $handler ? Proxy::wrapStreaming($handler, new StreamHandler()) : new StreamHandler();
         }
 
-        if (\ini_get('allow_url_fopen')) {
-            $handler = $handler
-                ? Proxy::wrapStreaming($handler, new StreamHandler())
-                : new StreamHandler();
-        } elseif (!$handler) {
+        if (!$handler) {
```
GrahamCampbell, line 98: `$handler === null`

GrahamCampbell (changes requested):

src/Utils.php
Calison25: Changed, it was generated automatically by PHP storm.

GrahamCampbell:
I am not convinced this makes it easier to read, actually. Now is a bit confusing, since the curl handler is the only thing returned from getHandler, and the main method possibly builds a stream handler. I think it was better before. Since this is not broken, I'm going to close this. Thanks anyway. :)

src/Utils.php
Calison25: changed too.

PR #2874, opened 2021-03-21
Test with guzzle/PSR7 2.0

Nyholm:


.github/workflows/ci.yml
```diff
@@ -43,6 +43,10 @@ jobs:
       max-parallel: 10
       matrix:
         php: ['7.2', '7.3', '7.4', '8.0']
+        psr7: ['^1.7']
+        include:
+          - php: '7.4'
+            psr7: '2.0.x-dev'
```
GrahamCampbell, line 49: `^2.0@dev`?

GrahamCampbell (changes requested):

composer.json
```diff
@@ -69,5 +69,7 @@
         "psr-4": {
             "GuzzleHttp\\Tests\\": "tests/"
         }
-    }
+    },
+    "prefer-stable": true,
+    "minimum-stability": "dev"
```
GrahamCampbell, line 74: Why is that needed?

GrahamCampbell: Better to just have the dev on the single dependency we want, otherwise now we run the risk of accidentally needing other dev deps and the CI not catching it.

GrahamCampbell (changes requested):

composer.json
Nyholm: This allows us to do `vendor/package:^3.0` before 3.0 is released. 

That will come in handy when Guzzle officially supports `guzzlehttp/psr7:2.0` which should happen before 2.0 is released. 

It also has the benefit that we dont need to update the CI when 2.0 is released.

GrahamCampbell: Doing `@dev` on the end of the individual version constraint we want also allows us to do that.

Nyholm:
This PR adding two new jobs. One that will test with `@alfa`, `@beta` and stable releases of ^2.0. And another job that will always test with the latest.

.github/workflows/ci.yml
```diff
@@ -43,6 +43,12 @@ jobs:
       max-parallel: 10
       matrix:
         php: ['7.2', '7.3', '7.4', '8.0']
+        psr7: ['^1.7']
+        include:
+          - php: '7.4'
+            psr7: '^2.0'
+          - php: '7.4'
+            psr7: 'dev-master'
```
GrahamCampbell, line 51: `^2.0@dev` does this already, and is robust again branch renames.

GrahamCampbell (changes requested):

.github/workflows/ci.yml
Nyholm: The PR is updated

GrahamCampbell (approved):

Nyholm:
Thank you

PR #2877, opened 2021-03-21
Prepare release 7.3.0

Nyholm:


CHANGELOG.md
```diff
@@ -2,6 +2,14 @@
 
 Please refer to [UPGRADING](UPGRADING.md) guide for upgrading to a major version.
 
+## 7.3.0 - 2021-03-21
+
+### Added
+
+- Support for DER and P12 certificates [#2413](https://github.com/guzzle/guzzle/pull/2413(
+- Support the cURL (http://) scheme for StreamHandler proxies [#2850](https://github.com/guzzle/guzzle/pull/2850)
+- Support for `guzzlehttp/psr7` [#2874](https://github.com/guzzle/guzzle/pull/2874)
```
GrahamCampbell, line 11: This should be omitted. I assume we don't want to allow 2.x in here until 2.0.0 stable is out. That could coincide with Guzzle 7.4.0.

GrahamCampbell (changes requested):

CHANGELOG.md
Nyholm: Nope. 

That is exactly what we do want. We want to allow people to install it and report bugs before 2.0 is released.

GrahamCampbell: Then we need to update Guzzle's composer.json to allow `^2.0@dev`.

CHANGELOG.md
```diff
@@ -2,6 +2,18 @@
 
 Please refer to [UPGRADING](UPGRADING.md) guide for upgrading to a major version.
 
+## 7.3.0 - 2021-03-21
+
+### Added
+
+- Support for DER and P12 certificates [#2413](https://github.com/guzzle/guzzle/pull/2413(
+- Support the cURL (http://) scheme for StreamHandler proxies [#2850](https://github.com/guzzle/guzzle/pull/2850)
+- Support for `guzzlehttp/psr7:2.0` [#2878](https://github.com/guzzle/guzzle/pull/2878)
```
GrahamCampbell, line 11: Maybe call it `^2.0`. Don't want people trying to copy this as a version constraint.

GrahamCampbell (approved):

CHANGELOG.md
Nyholm:
```suggestion
- Support for `guzzlehttp/psr7:^2.0` [#2878](https://github.com/guzzle/guzzle/pull/2878)
```

Sure!

PR #2868, opened 2021-03-10
Bumped branch alias

GrahamCampbell:
The next version will be 7.3.0 on this branch, and no, we shouldn't change this to 7.x-dev. 7.3-dev gives more information to composer.

PR #2880, opened 2021-03-21
Adding CI job to update branch alias when needed

Nyholm:
Follow up from #2868

### DIFF
--- a/.github/workflows/branch-alias.yml
+++ b/.github/workflows/branch-alias.yml
@@ -1,0 +1,66 @@
+name: Update branch alias
+
+on:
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Hmmm, this is not necessarily correct. The branch alias only needs bumping when the next new feature is merged, otherwise the branch exists to provide the next patch release for the current major.minor series.

Nyholm:
Yes, but it is way harder to automate if merged PR is a feature or not. 

This will make sure that the branch alias is always one minor version higher than the released version.

Issue #2846, opened 2021-01-28
Middleware::retry() (and others) typing


zerkms:
**Description**

Its return type is declared as

```
@return callable Returns a function that accepts the next handler.
```

while it in fact should be

```
@return callable(callable):callable Returns a function that accepts the next handler.
```

because `HandlerStack::push()` is declared as

```
    /**
     * Push a middleware to the top of the stack.
     *
     * @param callable(callable): callable $middleware Middleware function
     * @param string                       $name       Name to register for this middleware.
     */
    public function push(callable $middleware, string $name = ''): void
    {
        $this->stack[] = [$middleware, $name];
        $this->cached = null;
    }
```

**Example**

At the moment this code

```php
            $stack->push(Middleware::retry(
                $this->retryDecider($logger),
                function (int $retries): int {
                    return $retries * 1000;
                }
            ));
```

does not pass type check, because `Middleware::retry` return value is more generic than `$stack->push` argument parameter type.

**Additional context**

GrahamCampbell:
I'd be happy to review a PR that added more precise typing information (and possibly eliminated some lines from our static analyzer baseline files),

zerkms:
@GrahamCampbell instead of reducing baseline psalm I increased it significantly, sorry :-D

zerkms:
Not stale - there is a PR

zerkms:
I hate leaving such messages and I understand it's very low priority, but "not stale" :-)

PR #2885, opened 2021-03-26
[WIP] Fixed retry middleware return type

zerkms:
And also enabled level 1 psalm checks with strict typing

This is intended to fix https://github.com/guzzle/guzzle/issues/2846 and start working on better typing for the library.

Yep, I know level=1 + strict typing might be extreme, but from my personal experience switching from l3 to l2 then l1 then strict typing makes things even harder: you might end up fixing the same thing multiple times, which in fact increases amount of job to be done.


zerkms:
And the second commit (05936a0d3a40dca0b36f658f5650aba30a253e8a) types `\GuzzleHttp\Utils` (but one warning, see https://github.com/vimeo/psalm/issues/5490)

zerkms:
I have no prior experience with phpstan, hence I don't look at its results yet.

Nevertheless, I would be keen to hear from core team before I continue investing even more time into typing the library's codebase.

psalm.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0"?>
 <psalm
-    errorLevel="3"
+    totallyTyped="true"
```
GrahamCampbell, line 3: totallyTyped is deprecated. please set `errorLevel="1"`

GrahamCampbell (changes requested):

GrahamCampbell:
Going through this in detail. Lots of these changes look incorrect or unnecessary.

Issue #2894, opened 2021-05-30
Typo or PHP voodoo magic?


sualko:
Hi everyone. Can someone explain this line?

https://github.com/guzzle/guzzle/blob/de6f1e58e735754b888649495ed4cb9ae3b19589/tests/Cookie/SetCookieTest.php#L377

@SamuelWei @FelixJacobi and I think the `!=` should be a `-`.

SamuelWei:
@GrahamCampbell Any updates on this or your PR?

sualko:
@villfa can you maybe look into this issue?

villfa:
@sualko I'm just a contributor of this project, not a maintainer.

Anyway I have open a PR to this problem #2977

PR #2896, opened 2021-05-31
Fixed typo in SetCookieTest

GrahamCampbell:
Closes https://github.com/guzzle/guzzle/issues/2894.


Nyholm:
Oops. It seams like a few tests depended on this.

GrahamCampbell:
Hmm. These tests seem bad. Unix timestamps are not allowed: https://datatracker.ietf.org/doc/html/rfc7231#section-7.1.1.1.

PR #2897, opened 2021-06-03
Update readme for PHP8 workaround because of phpunit/phpunit not supp

hopocode:
PHP8 doues not work proper becaus of phpunit/phpunit dos nut support it jet. 
guzzle_php8_not_supported

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,17 +9,17 @@
 Guzzle is a PHP HTTP client that makes it easy to send HTTP requests and
 trivial to integrate with web services.
 
-- Simple interface for building query strings, POST requests, streaming large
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Guzzle's own dev-dependencies don't impact your app's.

hopocode:
It is not about my App. Guzzle has in composer php8+, but it is not running in php8+ because phphunit/phpunit does nut run PHP8

GrahamCampbell:
phpunit does run on PHP 8 and Guzzle doesn't require phpunit. It's in its isolated dev-dependencies.

PR #2904, opened 2021-06-18
Improve the comment regarding `CURLOPT_ENCODING` handling

TimWolla:
While researching how the `decode_content` option is meant to be used and
looking into the Guzzle source code I stumbled over this part of the code,
because it was non-obvious to me that providing an empty string to
`CURLOPT_ENCODING` does not *disable* the decoding, but instead enables *all*
decoders.


Nyholm (approved):
Thank you

PR #2909, opened 2021-07-01
Inclusion of safe array sum avoiding "unsupported operand types"

guil95:
Inclusion of safe array sum avoiding "unsupported operand types"

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -253,7 +253,7 @@ class Client implements ClientInterface, \Psr\Http\Client\ClientInterface
             $defaults['proxy']['no'] = \explode(',', $cleanedNoProxy);
         }
 
-        $this->config = $config + $defaults;
+        $this->config = $this->sumArray($config, $defaults);
 
         if (!empty($config['cookies']) && $config['cookies'] === true) {
             $this->config['cookies'] = new CookieJar();
@@ -301,7 +301,7 @@ class Client implements ClientInterface, \Psr\Http\Client\ClientInterface
         }
 
         // Shallow merge defaults underneath options.
-        $result = $options + $defaults;
+        $result = $this->sumArray($options, $defaults);
 
         // Remove null values.
         foreach ($result as $k => $v) {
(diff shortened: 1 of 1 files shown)


src/Client.php
```diff
@@ -253,7 +253,7 @@ private function configureDefaults(array $config): void
             $defaults['proxy']['no'] = \explode(',', $cleanedNoProxy);
         }
 
-        $this->config = $config + $defaults;
+        $this->config = (array) $config + (array) $defaults;
```
GrahamCampbell, line 256: it is impossible for either of these to not be an array

GrahamCampbell (changes requested):

src/Client.php
guil95: The same goes for line 304?

guil95: Today I had a problem for not having an array sum check on line 304 of this “Client” so I did the casting in the two places where I found the problem, if one of the values ​​is null it will give the error "types not supported operands ", by making this change we will no longer have this problem

GrahamCampbell:
What version of PHP are you running?

GrahamCampbell:
Does https://github.com/guzzle/guzzle/pull/2910 fix the issue?

guil95:
PHP 7.4.9

guil95:
It may still not fix, I believe that by casting to array we are preventing it and making sure it won't cause the problem...

GrahamCampbell:
Can you please test that change. I am not merging code that casts things that are already arrays.

guil95:
> Does #2910 fix the issue?
It may still not fix, I believe that by casting to array we are preventing it and making sure it won't cause the problem...

guil95:
> Você pode testar essa mudança. Não estou mesclando código que converte coisas que já são matrizes.

If they were already matrices I wouldn't break my application, but you have the power if you don't want to develop defensively, that's ok :D

GrahamCampbell:
There's a difference between defensive and impossible.

guil95:
> Existe uma diferença entre defensivo e impossível.

image

It's not impossible

GrahamCampbell:
And that's because of `this->config` being null, and not any of the changes you made. Please confirm #2910 fixes your issue.

guil95:
> E isso por `this->config`ser nulo, e não por nenhuma das alterações que você fez. Confirme se o [# 2910](https://github.com/guzzle/guzzle/pull/2910) corrige o seu problema.

This is the danger of using arrays for everything, I believe this casting would not affect the application at all, whether in performance or anything else, but that's ok

GrahamCampbell:
Please try the fix. This conversation is not helpful.

PR #2910, opened 2021-07-01
Fix uninitialized property

GrahamCampbell:


Nyholm:
I dont see that the `Client` can be initialised and `Client::$config` is null.

I saw the conversation in #2909. I think that it is fair to close this PR now.

PR #2914, opened 2021-07-17
Be more strict with types

### LINKED ISSUE #2888: setExpires/getMaxAge: Notice: A non well formed numeric value encountered
**Guzzle version(s) affected**: 7.3.0
**PHP version**: 7.3.19-1~deb10u1
**cURL version**: 7.66.0

**Description**  
The line [$this->setExpires(\time() + $this->getMaxAge());](https://github.com/guzzle/guzzle/blob/master/src/Cookie/SetCookie.php#L87) generates a "Notice: A non well formed numeric value encountered" when the received Cookie contains an invalid "Max-Age" for the cookie.

**How to reproduce**  
In my case the server returned two comma separated values for Max-Age: "63071999,9843797"

**Possible Solution**  
Either explicitly cast the return value from `$this->getMaxAge()` to "int", or add an additional validation for "digits only".

Nyholm:
Make sure getMaxAge() always returns null or int. Also trigger deprecations if wrong types are passed to SetCookie

This will fix #2888

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -33,7 +33,7 @@
         "guzzlehttp/promises": "^1.4",
         "guzzlehttp/psr7": "^1.7 || ^2.0",
         "psr/http-client": "^1.0",
-        "symfony/deprecation-contracts": "^2.4"
+        "symfony/deprecation-contracts": "^2.2"
     },
     "provide": {
         "psr/http-client-implementation": "1.0"
--- a/composer.json
+++ b/composer.json
@@ -32,7 +32,8 @@
         "ext-json": "*",
         "guzzlehttp/promises": "^1.4",
         "guzzlehttp/psr7": "^1.7 || ^2.0",
-        "psr/http-client": "^1.0"
+        "psr/http-client": "^1.0",
+        "symfony/deprecation-contracts": "^2.4"
     },
     "provide": {
         "psr/http-client-implementation": "1.0"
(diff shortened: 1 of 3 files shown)


GrahamCampbell:
We should probably turn off the authoritative phpdoc types setting in phpstan (or whatever it's called).

Nyholm:
Yeah... But I rather just add this to the baseline. 

Note how I prepare the user and code for version 8.

GrahamCampbell:
Ok, sure. Re-gen the baseline then.

Nyholm:
Baseline is updated and the CI is green.

composer.json
```diff
@@ -32,7 +32,8 @@
         "ext-json": "*",
         "guzzlehttp/promises": "^1.4",
         "guzzlehttp/psr7": "^1.7 || ^2.0",
-        "psr/http-client": "^1.0"
+        "psr/http-client": "^1.0",
+        "symfony/deprecation-contracts": "^2.4"
```
GrahamCampbell, line 36: `^2.2` would be better I think. There are no code changes between 2.2.0 and 2.4.0: https://github.com/symfony/deprecation-contracts/compare/v2.2.0...v2.4.0.

GrahamCampbell (changes requested):
One tiny comment. Then good to merge.

composer.json
Nyholm: Good idea!

PR #2916, opened 2021-07-17
Throw exception if you are using all wrong headers

### LINKED ISSUE #2911: strtolower() expects parameter 1 to be string, int given
**Guzzle version(s) affected**: 7.3
**PHP version**: 7.4.19
**cURL version**: 7.76.1

I get this error:

`strtolower() expects parameter 1 to be string, int given {"userId":1,"exception":"[object] (TypeError(code: 0): strtolower() expects parameter 1 to be string, int given at /home/vagrant/xxx/vendor/guzzlehttp/psr7/src/Utils.php:28)`

after calling the PayPal API with this command:

```
$response = Http::withHeaders([
            'Accept: application/json',
            'Content-Type: application/json',
        ])
        ->withToken($paypal_account->access_token)
        ->get(env('PAYPAL_LIVE_API_URL').'/v1/reporting/transactions', [
            'start_date' => $start_date,
            'end_date' => $end_date,
            'fields' => $fields,
            'balance_affecting_records_only' => 'N',
            'page_size' => $page_size,
            'page' => $page
        ]);
```

This happend after my composer update updated guzzlehttp from psr7 1.8.2 to 2.0.0.

In the meantime I fixed this for me adding this to my composer.json:

`"guzzlehttp/psr7": "1.8.2",`

Nyholm:
This will fix #2911

Note that this only thrown an exception if you are using all wrong headers. Ie, it check if you send a list or a map.

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -5,6 +5,7 @@ namespace GuzzleHttp;
 use GuzzleHttp\Cookie\CookieJar;
 use GuzzleHttp\Exception\GuzzleException;
 use GuzzleHttp\Exception\InvalidArgumentException;
+use GuzzleHttp\Exception\RequestException;
 use GuzzleHttp\Promise as P;
 use GuzzleHttp\Promise\PromiseInterface;
 use Psr\Http\Message\RequestInterface;
@@ -344,6 +345,9 @@ class Client implements ClientInterface, \Psr\Http\Client\ClientInterface
         ];
 
         if (isset($options['headers'])) {
+            if (array_keys($options['headers']) === range(0, count($options['headers']) - 1)) {
+                throw new RequestException('The headers array must have header name as keys.', $request);
+            }
             $modify['set_headers'] = $options['headers'];
             unset($options['headers']);
         }
--- a/tests/ClientTest.php
+++ b/tests/ClientTest.php
@@ -4,6 +4,7 @@ namespace GuzzleHttp\Tests;
 
 use GuzzleHttp\Client;
 use GuzzleHttp\Cookie\CookieJar;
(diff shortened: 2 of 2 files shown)


GrahamCampbell (approved):
It would be also good to add some additional validation in the psr7 class to ensure headers are actually of type `array<string, string>`.

Nyholm:
Yeah. I agree. But it is a bit tricky. 

Let's discuss that further on the psr7 repo

src/Client.php
```diff
@@ -344,6 +345,9 @@ private function applyOptions(RequestInterface $request, array &$options): Reque
         ];
 
         if (isset($options['headers'])) {
+            if (array_keys($options['headers']) === range(0, count($options['headers']) - 1)) {
+                throw new RequestException('The headers array must have header name as keys.', $request);
```
TimWolla, line 349: My understanding is that this PR is not yet in any release: Should this exception be changed to `InvalidArgumentException` instead? I seems to be more fitting.

GrahamCampbell: I think this was chosen for consistency with what similar code here does. I'm open to review this for Guzzle 8.

TimWolla: Looking further below within this method I am just seeing `InvalidArgumentException` being thrown. That was my primary motivation to suggest the change.

e.g.

https://github.com/guzzle/guzzle/blob/e38c668f3b565bf9d045f20460a9bdf415073bf8/src/Client.php#L437

and

https://github.com/guzzle/guzzle/blob/e38c668f3b565bf9d045f20460a9bdf415073bf8/src/Client.php#L356-L362

PR #2918, opened 2021-07-19
Fix Guzzle 6 on PHP 8.1

alexpott:
Compare https://3v4l.org/u1DEI#v8.0.8  with https://3v4l.org/u1DEI/rfc#vgit.master (PHP 8.1)

Guzzle 7 is not affected.

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -76,11 +76,6 @@ parameters:
 			path: src/Client.php
 
 		-
-			message: "#^Parameter \\#2 \\$prefix of function http_build_query expects string, null given\\.$#"
-			count: 1
-			path: src/Client.php
-
-		-
 			message: "#^Method GuzzleHttp\\\\ClientInterface\\:\\:send\\(\\) has parameter \\$options with no value type specified in iterable type array\\.$#"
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
I don't think we should actually make this change to Guzzle 6. It is EOL. Everyone should upgrade to Guzzle 7. Upgrading is mostly easy.

alexpott:
It'd be really helpful if this would make it into Guzzle 6. Looking at https://github.com/guzzle/guzzle - it's still in security fix mode. Drupal 9 is still dependent on Guzzle 6 and we're getting all the pieces in place to make sure Drupal 10 will ship with the latest Guzzle but we can't update the major version in Drupal 9 because Drupal extensions (i.e. things separate from the main project) depend on the Guzzle 6 API as shipped by Drupal core.

GrahamCampbell:
I think the proper approach here is for those other things to allow both Guzzle 6 and 7. We intentionally made the APIs, almost identical so this could be done.

GrahamCampbell:
Please let me know if there's something missing or changed in Guzzle 7 that prevents you from doing this.

PR #2919, opened 2021-07-20
Fix phpdoc handle_factory array

BafS:
- Array key should be `handle_factory`, not `factory`
- Added @param array information

### DIFF
--- a/src/Handler/CurlHandler.php
+++ b/src/Handler/CurlHandler.php
@@ -24,9 +24,9 @@ class CurlHandler
     /**
      * Accepts an associative array of options:
      *
-     * - factory: Optional curl factory used to create cURL handles.
+     * - handle_factory: Optional curl factory used to create cURL handles.
      *
-     * @param array $options Array of options to use with the handler
+     * @param array{handle_factory?:CurlFactoryInterface} $options Array of options to use with the handler
      */
     public function __construct(array $options = [])
     {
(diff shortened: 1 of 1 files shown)


src/Handler/CurlHandler.php
```diff
@@ -24,9 +24,9 @@ class CurlHandler
     /**
      * Accepts an associative array of options:
      *
-     * - factory: Optional curl factory used to create cURL handles.
+     * - handle_factory: Optional curl factory used to create cURL handles.
      *
-     * @param array $options Array of options to use with the handler
+     * @param array{handle_factory?:CurlFactoryInterface} $options Array of options to use with the handler
```
GrahamCampbell, line 29: `array{handle_factory?: ?CurlFactoryInterface}` is probably the correct type, here.

GrahamCampbell (changes requested):

src/Handler/CurlHandler.php
BafS: Thanks, about the space after the colon it would be great to consolidate around one convention, I did it like that because of https://github.com/guzzle/guzzle/blob/de6f1e58e735754b888649495ed4cb9ae3b19589/src/Handler/HeaderProcessor.php#L19

we may want to update the code there too

GrahamCampbell (approved):

PR #2940, opened 2021-10-08
Enabled typing for tests and updated psalm baseline

zerkms:
Fixes https://github.com/guzzle/guzzle/issues/2895 ?


GrahamCampbell:
Thanks for looking into this, all the same. :)

Sandesh333333:
Thanks

Sandesh333333 (commented):
Hubb

PR #2942, opened 2021-10-13
Throw `InvalidArgumentException` when an incorrect `headers` array is provided

TimWolla:
As discussed in PR #2916 after it was merged an `InvalidArgumentException` is
more fitting, as passing an invalid `headers` array is a clear programming
error that needs to be fixed and not caught.

This is consistent with the validation of the other options, e.g. when using
`multipart` and `form_params` at the same time.

see #2916
see a2b8dd1ad7e733d50e9c7b80cb375e0883a7088d

/cc @GrahamCampbell; thanks


TimWolla:
My pleasure. Happy to give back to libraries we are relying on, even if it's just the little things.

PR #2946, opened 2021-10-17
Replaced implicit URI to string coercion

GrahamCampbell:
When we eventually turn on strict typing, this would have caused issues.


Nyholm (approved):
Thank you

PR #2959, opened 2021-11-27
PHPStan 1.2

kallesommernielsen:
This PR updates PHPStan to 1.2 and fixes a bunch of errors at the same time; The remaining ones are left is related to the PHP 7<>8 stubs which I did not look too deep into.

I did opt for a few expectations, though it did not seem to be the common practice here, especially for the `$depth` parameter on `\json_decode()` which requires `int<1, max>`, though that can be fixed with `@var` tags as an alternative as PHP does not have the concept of signedness when it comes to `int`.

See inline comments for further in-depth thoughts

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ vendor/
 /phpstan.neon
 /phpunit.xml
 
+.idea/
 build/
 artifacts/
 docs/_build
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -5,6 +5,7 @@ includes:
 parameters:
     level: max
     checkMissingIterableValueType: false
+    treatPhpDocTypesAsCertain: false
     paths:
         - src
     bootstrapFiles:
--- a/src/Client.php
+++ b/src/Client.php
@@ -120,6 +120,8 @@ class Client implements ClientInterface, \Psr\Http\Client\ClientInterface
     public function send(RequestInterface $request, array $options = []): ResponseInterface
     {
         $options[RequestOptions::SYNCHRONOUS] = true;
+
+        /** @var ResponseInterface */
         return $this->sendAsync($request, $options)->wait();
     }
 
@@ -134,6 +136,7 @@ class Client implements ClientInterface, \Psr\Http\Client\ClientInterface
         $options[RequestOptions::ALLOW_REDIRECTS] = false;
         $options[RequestOptions::HTTP_ERRORS] = false;
 
+        /** @var ResponseInterface */
         return $this->sendAsync($request, $options)->wait();
     }
 
@@ -184,6 +187,8 @@ class Client implements ClientInterface, \Psr\Http\Client\ClientInterface
     public function request(string $method, $uri = '', array $options = []): ResponseInterface
     {
         $options[RequestOptions::SYNCHRONOUS] = true;
+
+        /** @var ResponseInterface */
         return $this->requestAsync($method, $uri, $options)->wait();
     }
 
@@ -264,6 +269,8 @@ class Client implements ClientInterface, \Psr\Http\Client\ClientInterface
             $this->config['headers'] = ['User-Agent' => Utils::defaultUserAgent()];
         } else {
             // Add the User-Agent header if one was not already set.
+
+            /** @var string $name */
             foreach (\array_keys($this->config['headers']) as $name) {
                 if (\strtolower($name) === 'user-agent') {
                     return;
--- a/src/Cookie/SessionCookieJar.php
+++ b/src/Cookie/SessionCookieJar.php
@@ -70,7 +70,7 @@ class SessionCookieJar extends CookieJar
             foreach ($data as $cookie) {
                 $this->setCookie(new SetCookie($cookie));
             }
-        } elseif (\is_string($data) && \strlen($data)) {
+        } elseif (!\is_string($data) && \strlen($data)) {
             throw new \RuntimeException("Invalid cookie data");
         }
     }
--- a/src/Cookie/SessionCookieJar.php
+++ b/src/Cookie/SessionCookieJar.php
@@ -70,7 +70,7 @@ class SessionCookieJar extends CookieJar
             foreach ($data as $cookie) {
                 $this->setCookie(new SetCookie($cookie));
             }
-        } elseif (\strlen($data)) {
+        } elseif (\is_string($data) && \strlen($data)) {
             throw new \RuntimeException("Invalid cookie data");
         }
     }
--- a/src/Cookie/SessionCookieJar.php
+++ b/src/Cookie/SessionCookieJar.php
@@ -70,7 +70,7 @@ class SessionCookieJar extends CookieJar
             foreach ($data as $cookie) {
                 $this->setCookie(new SetCookie($cookie));
             }
-        } elseif (!\is_string($data) && \strlen($data)) {
+        } elseif (!\is_string($data) || \strlen($data)) {
             throw new \RuntimeException("Invalid cookie data");
         }
     }
--- a/src/Exception/RequestException.php
+++ b/src/Exception/RequestException.php
@@ -91,7 +91,7 @@ class RequestException extends TransferException implements RequestExceptionInte
         }
 
         $uri = $request->getUri();
-        $uri = static::obfuscateUri($uri);
+        $uri = self::obfuscateUri($uri);
 
         // Client Error: `GET /` resulted in a `404 Not Found` response:
         // <html> ... (truncated)
--- a/src/Handler/CurlFactory.php
+++ b/src/Handler/CurlFactory.php
@@ -113,6 +113,8 @@ class CurlFactory implements CurlFactoryInterface
         // Return the response if it is present and there is no error.
         $factory->release($easy);
 
+        assert($easy->response !== null);
+
         // Rewind the body of the response if possible.
         $body = $easy->response->getBody();
         if ($body->isSeekable()) {
@@ -125,6 +127,7 @@ class CurlFactory implements CurlFactoryInterface
     private static function invokeStats(EasyHandle $easy): void
     {
         $curlStats = \curl_getinfo($easy->handle);
+        assert(\is_array($curlStats));
         $curlStats['appconnect_time'] = \curl_getinfo($easy->handle, \CURLINFO_APPCONNECT_TIME);
         $stats = new TransferStats(
             $easy->request,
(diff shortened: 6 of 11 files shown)


src/Handler/CurlFactory.php
```diff
@@ -113,6 +113,8 @@ public static function finish(callable $handler, EasyHandle $easy, CurlFactoryIn
         // Return the response if it is present and there is no error.
         $factory->release($easy);
 
+        assert($easy->response !== null);
```
kallesommernielsen, line 116: The above call to `$factory->release($easy);` may modify `$easy->response`, so the `!$easy->response` check above will not be sufficient here

src/Exception/RequestException.php
```diff
@@ -91,7 +91,7 @@ public static function create(
         }
 
         $uri = $request->getUri();
-        $uri = static::obfuscateUri($uri);
+        $uri = self::obfuscateUri($uri);
```
kallesommernielsen, line 94: This could potentially be wrong if it is intended to allow child classes to override `obfuscateUri()` and allow that to be applied with this factory method, if that is the case, then the visibility should be loosened to either `protected` or `public`

phpstan.neon.dist
```diff
@@ -5,6 +5,7 @@ includes:
 parameters:
     level: max
     checkMissingIterableValueType: false
+    treatPhpDocTypesAsCertain: false
```
kallesommernielsen, line 8: Since some of the code is a bit pedantic despite the docblock values, this means we already don't trust the user input

.gitignore
```diff
@@ -6,6 +6,7 @@ vendor/
 /phpstan.neon
 /phpunit.xml
 
+.idea/
```
kallesommernielsen, line 9: PHPStorm folder, wasn't sure what the convention was regarding editor specific ignores so I left it here

src/Cookie/SessionCookieJar.php
```diff
@@ -70,7 +70,7 @@ protected function load(): void
             foreach ($data as $cookie) {
                 $this->setCookie(new SetCookie($cookie));
             }
-        } elseif (\strlen($data)) {
+        } elseif (!\is_string($data) || \strlen($data)) {
```
kallesommernielsen, line 73: This could potentially be wrong because `$data` may still be `bool|int|float|string` at this point, though it seems like the intended effect as introduced in 2ba19b2c8b22794071f6053db2c43773aff06e59 means it could be simplified to:

```suggestion
        } elseif ($data) {
```

(or simply an `else`)

src/Handler/CurlMultiHandler.php
```diff
@@ -53,7 +53,7 @@ class CurlMultiHandler
     /**
      * @var array<mixed> An associative array of CURLMOPT_* options and corresponding values for curl_multi_setopt()
      */
-    private $options = [];
+    private $options;
```
kallesommernielsen, line 56: No need to initialize this as we are already writing to it inside the constructor, as I don't assume the code is intended to be run with `\ReflectionClass::newInstanceWithoutConstructor()`

src/Utils.php
```diff
@@ -21,10 +21,10 @@ final class Utils
      */
     public static function describeType($input): string
     {
-        switch (\gettype($input)) {
-            case 'object':
+        switch (true) {
+            case \is_object($input):
                 return 'object(' . \get_class($input) . ')';
-            case 'array':
+            case \is_array($input):
```
kallesommernielsen, line 27: PHPStan cannot understand this as `$input` still is `mixed` despite calling `gettype()`, so by calling the introspection functions allows PHPStan to understand the intended type here

src/Handler/StreamHandler.php
```diff
@@ -517,6 +517,8 @@ private function add_progress(RequestInterface $request, array &$options, $value
             $params,
             static function ($code, $a, $b, $c, $transferred, $total) use ($value) {
                 if ($code == \STREAM_NOTIFY_PROGRESS) {
+                    assert(\is_callable($value));
```
kallesommernielsen, line 520: This could potentially be fixed by a `callable` docblock, but given the assumption that `$value` is only used as such when `$code` is `\STREAM_NOTIFY_PROGRESS`, it will have to be a union and need either an assertion or a `@var` tag override

src/Handler/CurlMultiHandler.php
```diff
@@ -32,9 +32,9 @@ class CurlMultiHandler
     private $selectTimeout;
 
     /**
-     * @var resource|\CurlMultiHandle|null the currently executing resource in `curl_multi_exec`.
+     * @var int the currently executing resource in `curl_multi_exec`.
      */
-    private $active;
+    private $active = 0;
```
kallesommernielsen, line 37: Technically not needed to be initialized as the code checks will still evaluate `null` to falsy, but could potentially be an issue later on if typed properties is added when the minimum version of PHP is upped to at least 7.4

src/Handler/CurlFactory.php
```diff
@@ -338,6 +348,7 @@ private function applyHeaders(EasyHandle $easy, array &$conf): void
      */
     private function removeHeader(string $name, array &$options): void
     {
+        /** @var string $key */
```
kallesommernielsen, line 351: This is not ideal, but requires `$options` to be an array-shape, as there is no sanity check that the `_headers` index actually exists here which would also allow better type inference

src/Handler/CurlFactory.php
```diff
@@ -142,12 +145,19 @@ private static function invokeStats(EasyHandle $easy): void
...
         $ctx = [
             'errno' => $easy->errno,
             'error' => \curl_error($easy->handle),
             'appconnect_time' => \curl_getinfo($easy->handle, \CURLINFO_APPCONNECT_TIME),
-        ] + \curl_getinfo($easy->handle);
-        $ctx[self::CURL_VERSION_STR] = \curl_version()['version'];
+        ] + $curlStats;
+
+        /** @var array<array-key, mixed> $curlVersion */
+        $curlVersion = \curl_version();
```
kallesommernielsen, line 158: PHPStan and PSalm seems to disagree here, where PHPStan will correctly assume that both `\curl_getinfo()` and `\curl_getversion()` can return `false`.

`\curl_getinfo()`:
This seems to be an issue with the PHPStan stubs where `false` is only plausible if both arguments is passed to `\curl_getinfo()` (which we don't here): https://github.com/php/php-src/blob/master/ext/curl/interface.c#L3360

`\curl_version()`:
PHP has a check to see if cURL returns a null pointer, however the structure returned by atleast more modern versions of cURL:
 - `php-src` commit which introduced this check: https://github.com/php/php-src/commit/537284418cde811f6541708a928430c7967d4a27#diff-29336a112e08c5b5caa7950afb6601010867f40a4643a33263646504277f7b33R612
 - `php-src`: https://github.com/php/php-src/blob/master/ext/curl/interface.c#L1691
 - `curl`: https://github.com/curl/curl/blob/63c76681827b5ae9017f6c981003cd75e5f127de/lib/version.c#L503

src/Handler/CurlFactory.php
```diff
@@ -125,6 +127,7 @@ public static function finish(callable $handler, EasyHandle $easy, CurlFactoryIn
     private static function invokeStats(EasyHandle $easy): void
     {
         $curlStats = \curl_getinfo($easy->handle);
+        assert(\is_array($curlStats));
```
kallesommernielsen, line 130: This can be fixed similarly to my other comment below, however it means that the potential type information about `total_time` is lost from being a `float` to `mixed`, while the assertion is not too nice, a better fix would be to have an array-shape for this to preserve that type information

src/Client.php
```diff
@@ -120,6 +120,8 @@ public function sendAsync(RequestInterface $request, array $options = []): Promi
     public function send(RequestInterface $request, array $options = []): ResponseInterface
     {
         $options[RequestOptions::SYNCHRONOUS] = true;
+
+        /** @var ResponseInterface */
```
kallesommernielsen, line 124: All the `@var` tags added here is due to `guzzlehttp/promises`'s interface `PromiseInterface` having `mixed` as the return type

src/Client.php
```diff
@@ -264,6 +269,8 @@ private function configureDefaults(array $config): void
             $this->config['headers'] = ['User-Agent' => Utils::defaultUserAgent()];
         } else {
             // Add the User-Agent header if one was not already set.
+
+            /** @var string $name */
```
kallesommernielsen, line 273: Similar to my other comment in `CurlFactory`, since there is no array-shape here, the value of `$name` will be `array-key` (`int|string`), so PHPStan thinks we're passing `int` to `\strtolower()`.

.gitignore
GrahamCampbell: Please remove this/.

src/Exception/RequestException.php
GrahamCampbell: this is a breaking change. please leave this alone. we can change on guzzle 9

src/Handler/CurlFactory.php
GrahamCampbell, line 116: please don';t use `assert`.

src/Handler/CurlFactory.php
GrahamCampbell, line 130: same

src/Handler/CurlMultiHandler.php
GrahamCampbell: i think this was added on purpose. please check the git blame

src/Handler/StreamHandler.php
GrahamCampbell: again, please don't use the assert function.

src/Utils.php
GrahamCampbell: please revert all changes to this file.

GrahamCampbell (changes requested):

src/Utils.php
```diff
@@ -270,11 +270,14 @@ public static function isHostInNoProxy(string $host, array $noProxyArray): bool
...
     public static function jsonDecode(string $json, bool $assoc = false, int $depth = 512, int $options = 0)
     {
+        assert($depth >= 1);
+
         $data = \json_decode($json, $assoc, $depth, $options);
         if (\JSON_ERROR_NONE !== \json_last_error()) {
             throw new InvalidArgumentException('json_decode error: ' . \json_last_error_msg());
         }
 
+        /** @var object|array|string|int|float|bool|null */
```
Nyholm, line 280: @GrahamCampbell Should this also be reverted?

Is it because we already have `@return` in the doc block?

Nyholm (commented):
Cool. Im happy with this PR after we made changes according to Graham's review.

phpstan.neon.dist
allan-simon: isn't it more because they can't assert that people using guzzle also have phpstan or related, so they can't rely on phpdoc as a safeguard ?

PR #2961, opened 2021-12-01
Allow deprecation-contracts 3

derrabus:


GrahamCampbell (approved):

simondotws:
@derrabus @GrahamCampbell v3.0 of deprecation-contracts requires a minimum version of PHP8.0 so causes issues for people still on PHP7

GrahamCampbell:
It won’t, because we did not drop v2.

GrahamCampbell:
This is only because you generated your lock file using PHP 8 or did not include the proper platform config in your app's composer.json file to force it to fetch PHP 7.4 compatible dependencies.

GrahamCampbell:
> it requires --ignore-platform-reqs to avoid the error.

I would highly, highly recommend not doing that, as the code you install will fail at runtime, and not actually work.

togethernet:
> This is only because you generated your lock file using PHP 8 or did not include the proper platform config in your app's composer.json file to force it to fetch PHP 7.4 compatible dependencies.

What would be the proper platform config in composer.json to force fetch PHP 7.4 compatible dependencies?

mhamann:
You could set (though I don't recommend):

```json
{
  ...
  "config": {
    "platform": {
      "php": "7.4"
    }
  }
}
```

You lose flexibility and can have weird stuff happen when explicitly setting the platform here, because you're basically forcing composer to ignore the actual version of PHP you're using.

Instead, just make sure that you're using the correct version of PHP CLI. For example, I ran into this and in my case, my server was running PHP 7.4, but my command line was using PHP 8.1. So, when doing an install of guzzle, I was getting a version of symfony/deprecation/contracts that was incompatible with the server PHP version.

PR #2976, opened 2022-01-07
precise phpdoc typing of array

allan-simon:
So that phpstan and similar tools catch when you do typo like 

`["body" ,  $body"]` instead of `["body" => $body]`

### DIFF
--- a/src/ClientTrait.php
+++ b/src/ClientTrait.php
@@ -19,9 +19,9 @@ trait ClientTrait
      * relative path to append to the base path of the client. The URL can
      * contain the query string as well.
      *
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
We can consider this for Guzzle 8, but both the interface and the trait would need to be changed, I think.

GrahamCampbell:
Changing phpdoc is breaking as far as I'm concerned. It still constitutes changing the public interface.

allan-simon:
fair enough

GrahamCampbell:
I think probably will actually go with array-shape typing here, instead, as per #3194.

PR #2983, opened 2022-01-30
Add attribute to avoid deprection warnings on PHP 8.1

jtojnar:
PHP 8.1 considers missing return typehints different from the interfaces.

### DIFF
--- a/src/Cookie/CookieJar.php
+++ b/src/Cookie/CookieJar.php
@@ -217,11 +217,13 @@ class CookieJar implements CookieJarInterface
         return true;
     }
 
+    #[\ReturnTypeWillChange]
     public function count()
     {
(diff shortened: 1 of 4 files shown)


jtojnar:
The CI failures seem unrelated. I can try to fix them in a separate PR if you feel it is worth it.

GrahamCampbell:
Guzzle 6 is both EOL and also not compatible with PHP 8. Please upgrade to Guzzle 7.

jtojnar:
Unfortunately, I also need to support PHP 5.6 (since that is what some LTS distros still have) so upgrading is not possible for me at the moment. But if Guzzle 6 is EOL, I guess I need to do it in a different way.

PR #2988, opened 2022-02-24
Set a default ssl.peer_name context in StreamHandler

TimWolla:
This is required when using the `force_ip_resolve` option with the stream
handler:

As that option will cause the StreamHandler to manually resolve the hostname
and then replace the hostname with the resolved IP address in the URI, PHP
will use that IP address by default in the SNI of the TLS handshake.

Set an explicit ssl.peer_name within the stream's context based on the hostname
in the URL to fix this.

Setting a proper SNI is independent from TLS certificate validation, thus this
value must not be dependent on the `verify` option.

A test cannot be added, due to a lack of TLS support with the current testing
infrastructure. TLS support cannot easily be added, because it would require a
separate port and also certificates that would need to be commited to the
repository. However correctness can be verified by setting `force_ip_resolve`
to `v4` and attempting to make a request to `https://www.example.com/`. It will
fail without this commit and work with.

### DIFF
--- a/tests/Handler/Network/StreamHandlerTest.php
+++ b/tests/Handler/Network/StreamHandlerTest.php
@@ -1,0 +1,71 @@
+<?php
+
+namespace GuzzleHttp\Test\Handler\Network;
+
+use GuzzleHttp\Client;
+use GuzzleHttp\Handler\StreamHandler;
+use GuzzleHttp\HandlerStack;
+use GuzzleHttp\Psr7\Request;
+use GuzzleHttp\RequestOptions;
+use PHPUnit\Framework\TestCase;
+
+/**
+ * @covers \GuzzleHttp\Handler\StreamHandler
+ */
+class StreamHandlerTest extends TestCase
+{
+    public function setUp(): void
+    {
+        if (!($_SERVER['GUZZLE_TEST_ALLOW_NETWORK'] ?? false)) {
+            self::markTestSkipped("This test requires the GUZZLE_TEST_ALLOW_NETWORK environment variable.");
+        }
+    }
+
+    public function testSslRequestWorks()
+    {
+        $handler = new StreamHandler();
+
+        $response = $handler(
+            new Request('GET', 'https://www.example.com/'),
+            [
+                RequestOptions::STREAM => true,
+            ]
+        )->wait();
+
+        self::assertSame(200, $response->getStatusCode());
+        self::assertStringContainsString('<h1>Example Domain</h1>', (string)$response->getBody());
+    }
+
+    public function testSslRequestWorksWithForceIpResolve()
+    {
+        $handler = new StreamHandler();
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
What if Guzzle follows a redirect to another host. Won't this break that?

TimWolla:
It _should_ not: `getDefaultContext` already is responsible for setting the request headers (including `host` and possible authorization).

However I did not test this. If you'd like to hold off merging this, until I tested this: This is fine for me.

GrahamCampbell:
Thanks. Let us know the outcome, please. :)

TimWolla:
Will do, but this is not going to happen today. I will try to get around to it next week, but can't promise anything.

If you planned to release today, then please don't wait for this. I don't think this fix is super important, I just noticed this issue by accident while working on #2989.

GrahamCampbell:
No worries, and no, we don't plan to release today. Maybe we will do releases on this repo and the psr7 repo, this time next weekend.

TimWolla:
I've added a new TestCase that actually performs networking against third party services, it needs to be explicitly enabled by setting the GUZZLE_TEST_ALLOW_NETWORK environment variable.

With that I could confirm that redirects work as expected.

tests/Handler/Network/StreamHandlerTest.php
```diff
@@ -0,0 +1,71 @@
...
+use PHPUnit\Framework\TestCase;
+
+/**
+ * @covers \GuzzleHttp\Handler\StreamHandler
+ */
+class StreamHandlerTest extends TestCase
+{
+    public function setUp(): void
+    {
+        if (!($_ENV['GUZZLE_TEST_ALLOW_NETWORK'] ?? false)) {
```
GrahamCampbell, line 19: please use $_SERVER for reading env variables.

GrahamCampbell (changes requested):

tests/Handler/Network/StreamHandlerTest.php
```diff
@@ -0,0 +1,71 @@
...
+        $handler = new StreamHandler();
+
+        $response = $handler(
+            new Request('GET', 'https://www.example.com/'),
+            [
+                RequestOptions::STREAM => true,
+            ]
+        )->wait();
+
+        self::assertEquals(200, $response->getStatusCode());
```
GrahamCampbell, line 35: please use assertSame where possible

GrahamCampbell (changes requested):

GrahamCampbell (approved):

TimWolla:
On an unrelated note: Should the 8.0/8.1 CI be set to 'Required' in the repository settings?

GrahamCampbell (approved):

Nyholm (approved):
Thank you

PR #2989, opened 2022-02-24
Reject non-HTTP schemes in StreamHandler

TimWolla:
I recommend not to squash this PR when merging, as the commits try to tell a story.

Detailed reasoning can be found in the commit message of each of the both commits.

### DIFF
--- a/src/Handler/StreamHandler.php
+++ b/src/Handler/StreamHandler.php
@@ -266,6 +266,10 @@ class StreamHandler
             $methods = \array_flip(\get_class_methods(__CLASS__));
         }
 
+        if (!\in_array($request->getUri()->getScheme(), ['http', 'https'])) {
+            throw new RequestException(\sprintf("The scheme '%s' is not supported.", $request->getUri()->getScheme()), $request);
+        }
+
         // HTTP/1.1 streams using the PHP stream wrapper require a
         // Connection: close header
         if ($request->getProtocolVersion() == '1.1'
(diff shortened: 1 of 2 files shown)


Nyholm:
What if you want to fetch a file on an FTP server. Then you cannot use the stream handler, right?

GrahamCampbell:
Maybe something to consider for Guzzle 8. I don't feel like we can merge this on Guzzle 7.

TimWolla:
> What if you want to fetch a file on an FTP server. Then you cannot use the stream handler, right?

Right, but you already cannot do this, because FTP is not HTTP and thus no HTTP headers will be set. In fact FTP is how I originally found out about this issue: https://www.woltlab.com/community/thread/294643-jeden-tag-selben-fehler-im-log/ (in German).

So this does not break anything that wasn't broken previously. It just improves the error message.

TimWolla:
Alternatively just take the first commit for now, I've intentionally split this into two separate commits. The first one fixes the underlying issue (and that should be easy to verify).

GrahamCampbell (approved):

Nyholm (approved):
Thank you. 

I do need to squash the commits though. I like a nice history like this: 
Screenshot 2022-03-13 at 17 19 33

TimWolla:
> I do need to squash the commits though. I like a nice history like this:

That's where I disagree. For my / our projects I rather prefer separate small commits (unless they are useless stuff like “fix typo”). But you're the maintainers and need to work with it, so I won't complain :smiley:

Nyholm:
Most people (myself included) don’t do nice commit messages like you show here. I agree that those are pretty and they do make sense. Especially if we also force the PR to be updated with the latest from master before we merge. 

But that is not a good general rule enforce. (Not for an open source project at least) :/
And I rather have just one solution for everything.

PR #2992, opened 2022-03-04
Configure `convertDeprecationsToExceptions="true"` in phpunit.xml.dist

TimWolla:


TimWolla:
see also: php-http/client-integration-tests#57.

.github/workflows/ci.yml
```diff
@@ -18,6 +18,7 @@ jobs:
           php-version: '7.2'
           coverage: none
           extensions: mbstring, intl
+          ini-values: error_reporting=E_ALL
```
GrahamCampbell, line 21: can we do this in bootstrap instead, so direct running locally has this, too?

GrahamCampbell: eg https://github.com/laravel/framework/blob/v5.5.50/tests/bootstrap.php

GrahamCampbell (changes requested):

.github/workflows/ci.yml
TimWolla: Good idea. It appears to work.

GrahamCampbell (changes requested):
CI failing.

GrahamCampbell (approved):

PR #2994, opened 2022-03-13
Upgrade actions and SA tools

GrahamCampbell:


.github/workflows/static.yml
```diff
@@ -37,7 +37,7 @@ jobs:
 
     steps:
       - name: Checkout code
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
 
       - name: Setup PHP
         uses: shivammathur/setup-php@v2
```
TimWolla, line 43: I wonder if the PHP version should also be updated to 8.0 / 8.1 in all workflows. At least the behavior of some of PHP-CS-Fixer's rules change their behavior depending on the PHP version and the current PHP 7.4 is neither the newest PHP, nor the oldest PHP supported by Guzzle.

PR #3000, opened 2022-03-25
Update docs from default_handler to HandlerStack::create

stickeegreg:
The `default_handler` function has been removed and replaced by `HandlerStack::create()`. Update the docs to match.


GrahamCampbell (approved):

Nyholm (approved):
Thank you

PR #3004, opened 2022-04-05
Update .gitattributes

angeljqv:
Make the vendor 140kb smaller, 
is it possible or is there any reason not to do it?

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -13,3 +13,5 @@
 /psalm.xml             export-ignore
 /tests/                export-ignore
 /vendor-bin/           export-ignore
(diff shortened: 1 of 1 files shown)


mfn:
As a user, I welcome have those things right in my local code base to check without requiring going somewhere else on the internet.

erikn69:
140kb seems excessive, it would be better to separate the changelog based on major releases

GrahamCampbell:
I think we can consider this change for Guzzle 8. When our zip is compressed, it's much less than 140kb, and you can strip these files out of vendor if you want.

PR #3013, opened 2022-05-09
avoid depreciation issues in php8.1

Corepex:
**Description:**
wrapped `$this->timeToNext()` with `intval()` to avoid implicit conversion from float to int since this is depricated in php8.1.

### DIFF
--- a/src/Handler/CurlMultiHandler.php
+++ b/src/Handler/CurlMultiHandler.php
@@ -178,7 +178,7 @@ class CurlMultiHandler
         while ($this->handles || !$queue->isEmpty()) {
             // If there are no transfers, then sleep for the next delay
             if (!$this->active && $this->delays) {
-                \usleep($this->timeToNext());
+                \usleep(intval($this->timeToNext()));
             }
             $this->tick();
         }


GrahamCampbell:
Thanks for the PR, but this is incorrect for two reasons:

1. The `timeToNext` already is guaranteed to return an `int`.
2. `intval` should not be used. It's the same as an int cast, but much slower.

Corepex:
> Thanks for the PR, but this is incorrect for two reasons:
> 
> 1. The `timeToNext` already is guaranteed to return an `int`.
> 2. `intval` should not be used. It's the same as an int cast, but much slower.

@GrahamCampbell thanks for the performance hint - until now i didn't know that there's such a big performance difference ... but it is! 👍

solody:
Then why don't we prevent installing Guzzle 6 from php 8.1?
We should put a php 8.1 limitation in composer.json file.

GrahamCampbell:
Because it's too late. If we do that, composer just installs the version of Guzzle that did not have that upper-bound. We'd have to have added it back in 2015, before I was a maintainer.

GrahamCampbell:
My position is somewhere in the middle (the other school of thought is always bound to major.minor, so not even allow PHP 8.2 on guzzle 7 yet, since php does not follow semver), and as such Guzzle 7 has upper bounds on the PHP major version, but not minor version.

PR #3024, opened 2022-06-15
Stop pretending that PHP 8.1 is 8.0 in CI

TimWolla:
This should hopefully no longer be required, as PHP 8.1 is several months out
by now.


GrahamCampbell:
Thanks. I keep forgetting to do this. I notice it every time I do a release. 😆

GrahamCampbell (approved):

PR #3025, opened 2022-06-16
Ordered interfaces

MathiasReker:
Orders the interfaces in an implements or interface extends clause.


GrahamCampbell:
See earlier comments. In future, please don't open loads of PRs without consulting the authors. This helps make sure both of us aren't wasting our time. If you want to propose new CS styles, open a **single PR** with the proposed config changes.

PR #3029, opened 2022-06-16
Declare strict types

MathiasReker:
Force strict types declaration in all files.


GrahamCampbell:
Thanks for the PR< however this doesn't work, because there is still some implicit casting going on.

PR #3030, opened 2022-06-16
Get class to class keyword

MathiasReker:
Replace get_class calls on object variables with class keyword syntax.


GrahamCampbell:
See earlier comments. In future, please don't open loads of PRs without consulting the authors. This helps make sure both of us aren't wasting our time. If you want to propose new CS styles, open a **single PR** with the proposed config changes.

PR #3033, opened 2022-06-16
Use arrow functions

MathiasReker:
Anonymous functions with one-liner return statement must use arrow functions.


GrahamCampbell:

GrahamCampbell:
In this case, we can't use this syntax because it requires PHP 7.4.

PR #3034, opened 2022-06-16
Void return

MathiasReker:
Add void return type to functions with missing or empty return statements, but priority is given to @return annotations.

### DIFF
--- a/src/Handler/EasyHandle.php
+++ b/src/Handler/EasyHandle.php
@@ -104,7 +104,7 @@ final class EasyHandle
      *
      * @throws \BadMethodCallException
      */
-    public function __get($name)
+    public function __get($name): void
     {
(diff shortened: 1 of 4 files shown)


src/Handler/EasyHandle.php
```diff
@@ -104,7 +104,7 @@ public function createResponse(): void
      *
      * @throws \BadMethodCallException
      */
-    public function __get($name)
+    public function __get($name): void
```
GrahamCampbell, line 107: this is a breaking change, please revert. the others are fine.

GrahamCampbell (changes requested):

GrahamCampbell (changes requested):
Please apply the fixes from php-cs-fixer.

GrahamCampbell:
In light of all the other PRs, I'm actually going to close this. This doesn't fix or improve anything for library consumers.

PR #3041, opened 2022-06-16
No singleline whitespace before semicolons

MathiasReker:
Single-line whitespace before closing semicolon are prohibited.


GrahamCampbell:
Thanks for the PR, however we should enable the code style fixer to do this, if we want to enforce this rule. Can you do this, please, instead?

GrahamCampbell:
In light of all the other PRs, I'm actually going to close this.

PR #3046, opened 2022-06-26
rewind response body after format

wjfz:


GrahamCampbell:
Thanks for the PR, but we should not be trying to do this.

wjfz:
@GrahamCampbell 
I think it's a bit necessary. If somebody set a log middleware to response body and the response status code >= 400, the exception can't get the body summary because it's not rewind.

https://github.com/guzzle/psr7/blob/13388f00956b1503577598873fffb5ae994b5737/src/Message.php#L70

GrahamCampbell:
Oh, I think it's that message code that should probably do the re-wind.

GrahamCampbell:
https://github.com/guzzle/psr7/pull/525

wjfz:
Thank you, solve my problem, but i still puzzle about this.
If someone use the formatter in middware to log each response or do something else, the formatter read the response body but not rewind it, it's  destructive.
I think who use it, who restore it maybe better?

GrahamCampbell:
For small streams, people mostly consume them by casting them to a string, which automatically rewinds. A common mistake is to call `->getContent()`. If you must do a method call, do `->__toString()`.

wjfz:
Thank you for your reply.

PR #2929, opened 2021-09-05
Support PHP 8.1

GrahamCampbell:

PR #3054, opened 2022-07-26
Mock Handler for PHP8.1

wouter-toppy:
See https://github.com/guzzle/guzzle/pull/2929


GrahamCampbell:
Guzzle 6 cannot be used on PHP 8, that includes 8.1. For a PHP 8 compatible version of Guzzle, you must upgrade to Guzzle 7. Upgrading is largely easy - we intentionally made it so.

PR #3057, opened 2022-08-05
Add request to delay closure params

mohamednagy:
This PR adds the `$request` object to the delay closure, this would be useful when you want to delay the request based on specific logic which is my case.


GrahamCampbell:
Thanks/

PR #3075, opened 2022-09-26
Corrected docs and default connect timeout value to 300 seconds

aivus:
As per [cURL documentation](https://curl.se/libcurl/c/CURLOPT_CONNECTTIMEOUT.html) if `CURLOPT_CONNECTTIMEOUT` set to `0` it means 300 seconds timeout:

> Set to zero to switch to the default built-in connection timeout - 300 seconds.

Additionally, curl provides 2 options to set connect timeout:
[CURLOPT_CONNECTTIMEOUT](https://curl.se/libcurl/c/CURLOPT_CONNECTTIMEOUT.html)
[CURLOPT_CONNECTTIMEOUT_MS](https://curl.se/libcurl/c/CURLOPT_CONNECTTIMEOUT_MS.html)

As mentioned, `If both CURLOPT_CONNECTTIMEOUT and CURLOPT_CONNECTTIMEOUT_MS are set, the value set last will be used.`

Guzzle code sets `CURLOPT_CONNECTTIMEOUT` to default 150, but then uses `CURLOPT_CONNECTTIMEOUT_MS` to overwrite this value.

This PR replaces `CURLOPT_CONNECTTIMEOUT` on `CURLOPT_CONNECTTIMEOUT_MS` and changes the default value for connect timeout from 150 to 300 to get rid of the wrong behaviour.

### DIFF
--- a/docs/request-options.rst
+++ b/docs/request-options.rst
@@ -282,7 +282,7 @@ connect_timeout
 ---------------
 
 :Summary: Float describing the number of seconds to wait while trying to connect
-        to a server. Use ``0`` to wait indefinitely (the default behavior).
+        to a server. Use ``0`` to wait 300 seconds (the default behavior).
 :Types: float
 :Default: ``0``
 :Constant: ``GuzzleHttp\RequestOptions::CONNECT_TIMEOUT``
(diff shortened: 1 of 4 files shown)


GrahamCampbell:
Hmm. Do you know when this changed. I don't think this was always the case?

aivus:
@GrahamCampbell unfortunately I don't know if it has ever been changed.

gmponos:
https://github.com/guzzle/guzzle/pull/2559

GrahamCampbell:
Hahahaha. I'd totally forgotten I PRed something similar. 😆

PR #3079, opened 2022-10-24
request no encoding if user does not specify acceptable encodings

divinity76:
due to (imo shitty) specs, when the client does not request any encodings, the server is free to pick any encoding it wants. not sending any `Accept-Encoding` headers is equivalent to sending Accept-Encoding: *
(as * - anything - is the default Accept-encoding value)

IMO we shouldn't overwrite curl's Accept-Encoding header at all and the guzzle implementation is stupid, i'd love to hear the guzzle dev's rationale, 
 but at least setting CURLOPT_ENCODING='' + Accept-Encoding: identity is better than CURLOPT_ENCODING='' + Accept-encoding: *  (* is implicit when no accept-encoding is specified)

Lets say for example that curl is compiled with gzip+deflate, and curl *wants* to send `Accept-Encoding: gzip,deflate`, but because Guzzle sets `Accept-Encoding: *` (implicitly), the server opts to use br/brotli, resulting in broken downloads. this wouldn't have happened if guzzle didn't overwrite curl's default accept-encoding, and it wouldn't have happened if guzzle sent `Accept-Encoding: identity`, but it may happen if guzzle sends `Accept-Encoding: *` as it does prior to this patch.

### DIFF
--- a/src/Handler/CurlFactory.php
+++ b/src/Handler/CurlFactory.php
@@ -388,9 +388,9 @@ class CurlFactory implements CurlFactoryInterface
                 // The empty string enables all available decoders and implicitly
                 // sets a matching 'Accept-Encoding' header.
                 $conf[\CURLOPT_ENCODING] = '';
-                // But as the user did not specify any acceptable encodings we need
-                // to overwrite this implicit header with an empty one.
-                $conf[\CURLOPT_HTTPHEADER][] = 'Accept-Encoding:';
+                // But as the user did not specify any acceptable encodings we should
+                // tell the server that no encodings are acceptable..
+                $conf[\CURLOPT_HTTPHEADER][] = 'Accept-Encoding: identity';
             }
         }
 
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
Thanks for the PR, though I don't think this is something we could change in Guzzle 7. It'd have to be something to consider for Guzzle 8. What do you think @TimWolla, @Nyholm?

TimWolla:
I agree that the HTTP spec is shitty with regard to this, but I disagree that Guzzle should magically add an `accept-encoding: identity` header to the request, when the PSR-7 `Request` object did not specify such a header.

divinity76:
what should be done if the server opts to use br/brothli due to a lack of Accept-Encoding headers, and curl was not compiled with br/brotli support then? just let the download become corrupted? or perhaps decode it manually? or something else?

TimWolla:
> what should be done

The correct course of action depends on the use case within your application. If in doubt, set an explicit `accept-encoding: identity` header when creating the `Request` object.

----------

FWIW: The same issue would exist when you set `decode` to `false`. Then no decoders would be enabled, but the `accept-encoding: identity` header would not be set either, thus also allowing the server to send you arbitrary data.

GrahamCampbell:
I agree with @TimWolla. Thanks for taking the time to open up the discussion @divinity76. We're not going to merge this, since there is a mitigation you can apply in your application code in order to insist on no encoding if that's what you want.

Issue #3082, opened 2022-11-07
getenv method doesn't work in laravel config:cache


rezvani:
To optimize performance in laravel we should use this command:
`php artisan config:cache`

This command will be produce config file in cache and then application cannot access to .env  variables.
So getenv method in Util class will be return **null** instead of actual value in ENV

**How to reproduce**
For example in when i need to use proxy in all of requests in laravel packages (such as Telegram Notification) i set HTTPS_PROXY variable in .env. and as usual i will be run `php artisan optimize` command to cache all files and config for better performance.
in this situation, guzzle never use proxy, because `Utils::getenv('HTTPS_PROXY')` return null.

**Possible Solution**  
We can use function_exists method for `config` function to read from that.
        
        if (function_exists('config')){
            return config($name);
        }

        if (\PHP_SAPI === 'cli' && ($value = \getenv($name)) !== false && $value !== null) {
            return (string) $value;
        }

        return null;
    }

GrahamCampbell:
Thanks for getting in touch. One should set this as config when constructing the Guzzle client, passing it through explicitly in the constructor. Using global variables for configuration is an anti-practice that exists only for legacy BC.

PR #3083, opened 2022-11-07
Support laravel config cache in getenv

rezvani:
update getenv method in Util Class to support laravel cache
https://github.com/guzzle/guzzle/issues/3082


GrahamCampbell:
I am strongly against this. This is a major misuse of this function.

PR #3084, opened 2022-11-10
MessageFormatter no longer modifies pointerposition of body-streams

Zombaya:
Currently, the MessageFormatter provided for the logging middleware modifies the internal pointer of the body-streams of the request and response.  This PR makes the MessageFormatter more transparent by resetting the pointer to the original position if it is possible.

# Why
We recently added logs to our application using `GuzzleHttp\Middleware::log()` and `GuzzleHttp\MessageFormatter`.  This caused our existing code to start failing as we were using `$response->getBody()->getContents()`.
As I've now learned, it's probably better if we were to use `(string) $response->getBody()`.
Nevertheless I feel like it would be better if MessageFormatter would no longer modify the internal pointer position so using this middleware is more transparent to the end user.

# Example-code
```php
/*
 * Original code
 */

// $body is not empty
$body = $client->get('www.example.com')->getBody()->getContents();

/*
 * Code modified to use logging middleware
 */
$handler = \GuzzleHttp\HandlerStack::create();
$handler->push(\GuzzleHttp\Middleware::log(new \Psr\Log\NullLogger(),new \GuzzleHttp\MessageFormatter()));
$client = new \GuzzleHttp\Client([
    'handler' => $handler
]);

// Original code: $body is empty
// New code in PR: $body is not empty
$body = $client->get('www.example.com')->getBody()->getContents();

// Original code: $body is not empty
// New code in PR: $body is not empty
$body = (string) $client->get('www.example.com')->getBody();

```

### DIFF
--- a/tests/MessageFormatterTest.php
+++ b/tests/MessageFormatterTest.php
@@ -50,42 +50,49 @@ class MessageFormatterTest extends TestCase
 
     public function formatProvider()
     {
-        $request = new Request('PUT', '/', ['x-test' => 'abc'], Psr7\Utils::streamFor('foo'));
-        $response = new Response(200, ['X-Baz' => 'Bar'], Psr7\Utils::streamFor('baz'));
-        $err = new RequestException('Test', $request, $response);
+        $request = static function () {
+            return new Request('PUT', '/', ['x-test' => 'abc'], Psr7\Utils::streamFor('foo'));
+        };
+        $response = static function () {
+            return new Response(200, ['X-Baz' => 'Bar'], Psr7\Utils::streamFor('baz'));
+        };
+        $err = static function () use ($request, $response) {
+            return [$req = $request(),$res = $response(),new RequestException('Test', $req, $res)];
+        };
 
         return [
-            ['{request}', [$request], Psr7\Message::toString($request)],
-            ['{response}', [$request, $response], Psr7\Message::toString($response)],
-            ['{request} {response}', [$request, $response], Psr7\Message::toString($request) . ' ' . Psr7\Message::toString($response)],
+            ['{request}', [$request()], Psr7\Message::toString($request())],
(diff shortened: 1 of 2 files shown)


tests/MessageFormatterTest.php
```diff
@@ -50,42 +50,49 @@ public function testFormatsTimestamps(string $format, string $pattern)
...
-        $err = new RequestException('Test', $request, $response);
+        $request = static function () {
+            return new Request('PUT', '/', ['x-test' => 'abc'], Psr7\Utils::streamFor('foo'));
+        };
+        $response = static function () {
+            return new Response(200, ['X-Baz' => 'Bar'], Psr7\Utils::streamFor('baz'));
+        };
+        $err = static function () use ($request, $response) {
+            return [$req = $request(),$res = $response(),new RequestException('Test', $req, $res)];
+        };
```
Zombaya, line 61: I replaced the original variables with functions so that each test uses a fresh object, where all streams have a internal pointer position set to 0.

GrahamCampbell:
The problem here is that the application code `$resp->getBody()->getContents();` you have is incorrect. One should have written `(string) $resp->getBody()`.

Zombaya:
I'm now aware the application code I had written was incorrect and have changed it.  I do feel like it's a bit of a footgun as I wrongly assumed that the `getContents()`-method would have gotten me the contents of the body without any side-effects, or that adding the logging middleware would have no side-effects.

I'm aware this was an error in my application code, resulting from lack of knowledge about streams.

I kindly ask you to reconsider this PR as other than restoring the internal pointer position of the streams, this change has no other impact on the library.

Even so, I still like to thank you for the work you do on this project since it brings a lot of value to our project and many with it.

GrahamCampbell:
Unfortunately this is the design of the psr-7 interfaces, which is out of our control. `getContent` has to just get the rest of the data in the stream from where ever the pointer is. ;)

PR #3087, opened 2022-11-28
Cast to array, preventing unsupported operands error

joeytyndale:
We've observed cases where options have come through as something other than an array. This change is covered by existing unit tests and offers a sanity check and prevents Unsupported Operands error from being thrown.


GrahamCampbell:
It is impossible for these not to be arrays. If this is an IDE warning, it's wrong. If you are seeing crashes, upgrade to the latest version.

GrahamCampbell:
The only way this could happen is if you extend this class (not allowed - it is soft-final), and forget to call the parent constructor.

PR #3089, opened 2022-12-05
SetCookie constructor bugfix

idealogica:
The following issue has been fixed:
When using fromString(), for the following kind of incorrectly formed cookie
```fr=synced; Max-Age=604800 Expires=Mon, 12 Dec 2022 13:27:50 GMT; Domain=.emxdgt.com; Path=/; SameSite=None; Secure; HttpOnly```
it gives ```A non well formed numeric value encountered in /app/proxy/vendor/guzzlehttp/guzzle/src/Cookie/SetCookie.php notice```

### DIFF
--- a/src/Cookie/SetCookie.php
+++ b/src/Cookie/SetCookie.php
@@ -82,7 +82,7 @@ class SetCookie
 
         $this->data = $replaced;
         // Extract the Expires value and turn it into a UNIX timestamp if needed
-        if (!$this->getExpires() && $this->getMaxAge()) {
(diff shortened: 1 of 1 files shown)


unixdevil:
have a look to the is numeric :)

GrahamCampbell:
Thanks for the PR, but I don't feel that this is the correct solution. The return type of `getMaxAge` is indicated as `int|null`, so we should probably make it so.

GrahamCampbell:
It looks like missing constructor validation is the issue here.

PR #3098, opened 2023-02-06
PHP 8.2 Dynamic Property Compat

rodnaph:
Adds the `\AllowDynamicProperties` annotation to suppress warnings about dynamic property creation on PHP 8.2.

This has also been raised in the `psr7` repository:

https://github.com/guzzle/psr7/pull/547
https://github.com/guzzle/psr7/pull/519

Are there concerns for supporting older versions of PHP that this is incompatible with?


GrahamCampbell:
Guzzle 6 is not compatible with PHP 8.1 or 8.2. You'll need to use Guzzle 7 there.

GrahamCampbell:
See https://github.com/guzzle/guzzle#version-guidance for notes on what versions are supported where.

PR #3100, opened 2023-02-16
Ability to pass callable as logLevel to Middleware::log()

zlodes:
Hi!

Sometimes we want to choice different levels for different Requests/Response pair. 
Especially for different response status codes.

This PR brings an ability to pass `callable` as `logLevel` param of `Middleware::log()`.


zlodes:
I'll assess why the PHPStan task failed... It works fine locally with the same version of PHPStan.

GrahamCampbell:
Thanks for the PR, but I think we should leave this out of the core.

PR #3113, opened 2023-03-21
Update Testing documentation for test-server package

scorgn:
This MR updates the Testing documentation to clarify that the test-server is part of an external package instead of included directly with Guzzle. Since the test server is not shipped with Guzzle when installed via composer the documentation is a bit confusing for those wanting to use it.


GrahamCampbell:
Thanks for taking the time to prep this PR. 👍

PR #3117, opened 2023-04-11
Fix `NO_PROXY` settings

gergokee:
**Description**
When using the proxy => no request option guzzle fails to ignore the proxy if ENV vars are present on the server i.e. http_proxy or https_proxy.

This is because the no option only applies the CURLOPT_PROXY if the host is not in the list. So when ignoring it doesn't apply the curl option so falling back to the proxy that is set in the ENV var.

Fixes: https://github.com/guzzle/guzzle/issues/3093


GrahamCampbell (changes requested):
The tests are failing here, indicating that this behaviour was possibly intentional? If not, can you update the tests and add new tests please?

gergokee:
> The tests are failing here, indicating that this behaviour was possibly intentional? If not, can you update the tests and add new tests please?

Fixed, added `unset()` instead of setting to empty value. Test are running succeeding now.

oostapov (approved):

GrahamCampbell:
Thanks for working on this. 👍

PR #3124, opened 2023-04-20
Prevent to enable IDN conversion if the intl extension or its polyfill is not available

phansys:
```shell
Fatal error: Uncaught Error: Undefined constant "GuzzleHttp\IDNA_DEFAULT" in /app/vendor/guzzlehttp/guzzle/src/Client.php:218
```


GrahamCampbell:
Thanks, but Guzzle 6 is no longer supported for bug fixes. Please upgrade to Guzzle 7.

GrahamCampbell:
Guzzle 7 has a different default here, btw. We didn't change this on Guzzle 6, because silently doing nothing on some systems would be a breaking change - we want to crash, so people choose to explicitly turn it off.

Issue #3128, opened 2023-05-04
Document that the calling code is responsible for closing streams


andrewnicols:
**Guzzle version(s) affected**: 7.5.0
**PHP version**: 8.1.17
**cURL version**: 8.0.1.

**Description**  
When passing in a stream as the contents argument to a multipart request, the stream is not closed by Guzzle.
As a result, it is not possible to remove the source file that the stream is opening in some OSes (e.g. Windows).

Arguably it's correc that it's not closed, but this should probably be noted in the documentation that the calling code is responsible for closing the stream when finished.

**How to reproduce**  

This is easy to reproduce on Windows with unit tests where a file handle is not closed, and a file path is reused in a subsequent test. See the testcase here:

https://github.com/andrewnicols/guzzleCloseStreams
https://github.com/andrewnicols/guzzleCloseStreams/blob/main/tests/StreamedContentTest.php

This uses a data provider to run the same test with the same stream source five times.
The first two tests close the stream after running the test.
The third test does not.
The fourth test closes it, and the fifth test does not.

You can see in this run: https://github.com/andrewnicols/guzzleCloseStreams/actions/runs/4879658612/jobs/8706434793

The first three tests pass.
All tests after this point fail because the file handle in the third test was not closed.

**Possible Solution**  
Document a note that the calling code is responsible for closing the stream.

**Additional context**  
Seems to only affect Windows.

mfn:
Interesting, in a different use-cse when using a `sink` and providing a "resource", once the Request object gets out of scope, the handle _is_ closed => #2400

GrahamCampbell:
Streams are, in general, closed automatically https://github.com/guzzle/psr7/blob/2.5.0/src/Stream.php#L69-L75, unless they stay referenced by your code or a previous response middleware. The test you have there is invalid, because the mock handler retains a reference to the stream.

andrewnicols:
Hi Graham,

I have added an additional test to:
* not store the Mock in a variable, at all.
* expicitly unset the ref to the client.

Still failing:
https://github.com/andrewnicols/guzzleCloseStreams/blob/main/tests/StreamedContentTest.php#L104-L137
https://github.com/andrewnicols/guzzleCloseStreams/actions/runs/4911531600/jobs/8769649762

This behaviour should, at the very least, be documented.

andrewnicols:
Hi @GrahamCampbell,

[...]

PR #3129, opened 2023-05-04
Add notes about closing streams after use (#3128)

andrewnicols:
As noted in #3128, it would be helpful to add a note about the need to close streams after using them. Failure to do so can cause issues, particularly on Windows.

### DIFF
--- a/docs/psr7.rst
+++ b/docs/psr7.rst
@@ -363,6 +363,12 @@ write-only, read-write, allow arbitrary random access (seeking forwards or
 backwards to any location), or only allow sequential access (for example in the
 case of a socket or pipe).
 
+.. note::
+
+    When a stream is no-longer needed, the ``close()`` method should be called.
+    Guzzle will not close streams automatically when the request or response object
+    is destroyed.
+
 Guzzle uses the ``guzzlehttp/psr7`` package to provide stream support. More
 information on using streams, creating streams, converting streams to PHP
 stream resource, and stream decorators can be found in the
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
Streams are, in general, closed automatically https://github.com/guzzle/psr7/blob/2.5.0/src/Stream.php#L69-L75, unless they stay referenced by your code or a previous response middleware.

andrewnicols:
Hi @GrahamCampbell ,

Thanks for your reply on this.

Whilst this may be the intent, something is preventing this from working properly on Windows. See the following [testcase](https://github.com/andrewnicols/guzzleCloseStreams/blob/main/tests/StreamedContentTest.php) and it's [GHA Run](https://github.com/andrewnicols/guzzleCloseStreams/actions/runs/4911531600/jobs/8769649762).

In this test I'm creating a file using `file_put_contents()`, obtaining a file handle, and then using a Guzzle Client to make a Request. I use `Utils::streamFor(fopen(...))` to create the handle and do not store any reference to either the stream, or the file handle.

The first time that the test runs, everything works correctly.
Any subsequent time, it is not possible to create the file using `file_put_contents`. No reference is kept to either the stream, or the client, within the test code. I believe that happens because the file handle is still open to it.

I've added a second test to the same testcase which is almost identical except that a reference _is_ stored, and is explicitly closed. This test passes on every iteration.

I would also argue that the _intended_ behaviour also should be documented. That is, that a note should be added to explain that the file handle will be closed when the reference to the stream is unset and that, if you keep a handle to the stream, then it will not be closed until you dispose of the handle.

PR #3138, opened 2023-05-15
PR for #3137: Using `\GuzzleHttp\Cookie\SetCookie::fromString` with a `Max-Age` set will always result in a deprecation warning in `\GuzzleHttp\Cookie\SetCookie::setMaxAge`.

### LINKED ISSUE #3137: Using `\GuzzleHttp\Cookie\SetCookie::fromString` with a `Max-Age` set will always result in a deprecation warning in `\GuzzleHttp\Cookie\SetCookie::setMaxAge`.
**Guzzle version(s) affected**: 7.6.0
**PHP version**: (I think irrelevant to this bug)
```
PHP 8.1.12 (cli) (built: Oct 25 2022 18:16:21) (ZTS Visual C++ 2019 x64)
Copyright (c) The PHP Group
Zend Engine v4.1.12, Copyright (c) Zend Technologies
    with Xdebug v3.1.3, Copyright (c) 2002-2022, by Derick Rethans
```
**cURL version**: (I think irrelevant to this bug)
```
curl 7.88.1 (x86_64-w64-mingw32) libcurl/7.88.1 OpenSSL/1.1.1t (Schannel) zlib/1.2.13 brotli/1.0.9 zstd/1.5.4 libidn2/2.3.4 libpsl/0.21.2 (+libidn2/2.3.3) libssh2/1.10.0 nghttp2/1.52.0
Release-Date: 2023-02-20
Protocols: dict file ftp ftps gopher gophers http https imap imaps ldap ldaps mqtt pop3 pop3s rtsp scp sftp smb smbs smtp smtps telnet tftp
Features: alt-svc AsynchDNS brotli HSTS HTTP2 HTTPS-proxy IDN IPv6 Kerberos Largefile libz MultiSSL NTLM PSL SPNEGO SSL SSPI threadsafe TLS-SRP UnixSockets zstd
```

**Description**  
When using  `\GuzzleHttp\Cookie\SetCookie::fromString`, due to the `explode` the `$value` of all `$key`s will be a String.
So this is also the case for the `$value` of `$key` `Max-Age`.

In the reworked `\GuzzleHttp\Cookie\SetCookie::__construct` since version 7.6.0, `\GuzzleHttp\Cookie\SetCookie::setMaxAge` is now always called (previously it wasn't).
Since `$maxAge` is always a String when coming from `\GuzzleHttp\Cookie\SetCookie::fromString`, the `!is_int($maxAge)` will always be true, and thus the deprecation `Since guzzlehttp/guzzle 7.4: Not passing an int or null to GuzzleHttp\Cookie\SetCookie::setMaxAge() is deprecated and will cause an error in 8.0.` will always be fired.

**How to reproduce**  
Since `trigger_deprecation` triggers a silenced deprecation notice, this is sadly not reproducible with tests in Guzzle itself.
It can only be detected in the actual project by setting a breakpoint on the `trigger_deprecation` in `setMaxAge`.

I came from the Land of Drupal, where we don't silence deprecations.
Our daily CI job that runs with the latest updated versions of our dependencies found out about this today, see https://www.drupal.org/pift-ci-job/2666829 and scroll down a bit for messages like these:
```
Remaining indirect deprecation notices (28)

  28x: Since guzzlehttp/guzzle 7.4: Not passing an int or null to GuzzleHttp\Cookie\SetCookie::setMaxAge() is deprecated and will cause an error in 8.0.
```
Here's the Drupal Core issue: https://www.drupal.org/project/drupal/issues/3360288

Not only does this break our CI job, it will also break Guzzle itself when the time comes for version 8.0 and the deprecation message will turn into an error.

**Possible Solution**  
PR #3138 tries to solve this by checking if `$value` for `$key` `Max-Age` is a numeric string, and if so does a cast to int.
(FWIW here: https://www.drupal.org/pift-ci-job/2667171 is a test run which uses #3138 and has no deprecation warnings)
**Additional context**

Boegie:
Attempts to resolve #3137

### DIFF
--- a/src/Cookie/SetCookie.php
+++ b/src/Cookie/SetCookie.php
@@ -60,7 +60,7 @@ class SetCookie
                     if (!\strcasecmp($search, $key)) {
                         if ($search === 'Max-Age') {
                             if (is_numeric($value)) {
-                                $value = (int) $value;
+                                $data[$search] = (int) $value;
                             }
                         } else {
                             $data[$search] = $value;
--- a/src/Cookie/SetCookie.php
+++ b/src/Cookie/SetCookie.php
@@ -58,10 +58,13 @@ class SetCookie
             } else {
                 foreach (\array_keys(self::$defaults) as $search) {
                     if (!\strcasecmp($search, $key)) {
-                        if ($search === 'Max-Age' && is_numeric($value)) {
-                            $value = (int) $value;
+                        if ($search === 'Max-Age') {
+                            if (is_numeric($value)) {
+                                $value = (int) $value;
+                            }
+                        } else {
+                            $data[$search] = $value;
                         }
-                        $data[$search] = $value;
                         continue 2;
                     }
                 }
--- a/src/Cookie/SetCookie.php
+++ b/src/Cookie/SetCookie.php
@@ -58,6 +58,9 @@ class SetCookie
             } else {
                 foreach (\array_keys(self::$defaults) as $search) {
                     if (!\strcasecmp($search, $key)) {
+                        if ($search === 'Max-Age' && is_numeric($value)) {
+                            $value = (int) $value;
+                        }
                         $data[$search] = $value;
                         continue 2;
                     }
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
Fixed in 7.5.3 and 7.6.1.

Issue #3154, opened 2023-07-14
Abandoned php-http/message-factory package is used


mvczajkowski:
Guzzle version(s) affected: 7.7.0
PHP version: 8.2 (hint: php --version)

**Description**  
When using "composer install" a "Package php-http/message-factory is abandoned, you should avoid using it. Use psr/http-factory instead" warning appears. It seems that since 7.7.0 uses php-http/message-factory package which is abandoned.

**How to reproduce**  
Composer Install

**Possible Solution**  
Use psr/http-factory

**Additional context**  
Package was added in [Fixed tests](https://github.com/guzzle/guzzle/commit/aff112895ebd2c59585f5ba3adce0a51c105e164) commit and released in 7.7.0.

GrahamCampbell:
This only in require-dev. When you use guzzle in your own code, this package is not pulled in.

mvczajkowski:
This still isn't proper package and creates warnings. It shouldn't be used there.

GrahamCampbell:
The warnings are totally fine, and don't affect you in any way.

bastien-wink:
We should still plan to migrate to psr/http-factory at some point. It will otherwise show this warning in lots of projects.

GrahamCampbell:
Can you show me where installing guzzle in your app or library shows the warning please?

bastien-wink:
During a `composer install` (including require-dev)

The warning is 
`Package php-http/message-factory is abandoned, you should avoid using it. Use psr/http-factory instead.`

Tested with Composer 2.1.8

GrahamCampbell:
That is not relevant for installing within this repo itself. If you don't see this warning when requiring guzzle as a dependency, then it doesn't matter.

ghost:
My project does not require guzzle at all directly. It requires "league/oauth2-client", which requires guzzle and I sill get the warning.


```
composer install
Installing dependencies from lock file (including require-dev)
Verifying lock file contents can be installed on current platform.
Nothing to install, update or remove
Package php-http/message-factory is abandoned, you should avoid using it. Use psr/http-factory instead.
Generating optimized autoload files
64 packages you are using are looking for funding.
Use the `composer fund` command to find out more!
```

GrahamCampbell:
That's probably not from us. Can you run `composer why php-http/message-factory` please?

ghost:
A thanks, you are right.

PR #3156, opened 2023-07-25
Replace the abandoned `php-http/message-factory` package

kayw-geek:
This package is abandoned and no longer maintained. The author suggests using the [psr/http-factory](https://packagist.org/packages/psr/http-factory) package instead.


nicekiwi:
@kayw-geek ~~why was this closed? This is still an issue.~~

EDIT: It's not an issue. 

For anyone else coming across this, see: https://github.com/guzzle/guzzle/issues/3154

GrahamCampbell:
It's a non-issue, as it's a dev dependency.

PR #3157, opened 2023-08-03
Run CI with PHP 8.3

TimWolla:


GrahamCampbell:
Looks like we forgot https://github.com/guzzle/guzzle/pull/3171, too. ;)

PR #3158, opened 2023-08-03
Remove the need for `AllowDynamicProperties` in CurlMultiHandler

TimWolla:
see guzzle/psr7#547

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -141,6 +141,16 @@ parameters:
 			path: src/Handler/CurlMultiHandler.php
 
 		-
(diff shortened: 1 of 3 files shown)


GrahamCampbell:
Heh. Can you re-gen the psalm baseline plz - I'm happy that we just make it stfu for these changes for now. ;)

TimWolla:
Okay, I have no idea what I did wrong, but CI still is red.

It was not easy to get PHPStan and Psalm working in the first place, because apparently they are broken on PHP 8.2 (and also complain when executed with PHP 7.4 like in CI).

TimWolla:
@GrahamCampbell Got it to work, CI is green now.

GrahamCampbell:
image

PR #3160, opened 2023-08-15
Added 'curl' tp RequestOptions

grogy:
The goal of the merge request is to document the functionality when a user needs to add curl options


GrahamCampbell:
I am not so sure we should allow this, as it breaks the abstraction of the library to know about the underlying handler.

grogy:
Thank you for the comment. Ok, it is good opinion. I closed the merge request.

Issue #3163, opened 2023-08-18
The no proxy request option does not avoid falling back to proxies set in ENV vars


oostapov:
**Guzzle version(s) affected**: 7.7
**PHP version**: 8.1.22 (hint: `php --version`)
**cURL version**: 8.2.1 (hint: `php -i | grep cURL`)

**Description**  
When using the `proxy` => `no` request option guzzle fails to ignore the proxy if ENV vars are present on the server i.e. `http_proxy` or `https_proxy`.

The issue was introduced in https://github.com/guzzle/guzzle/issues/3093 . Original solution in the first commit https://github.com/guzzle/guzzle/pull/3117/commits/e0de379ca55e38bf06a019d46f0359b1ecba721a (setting `CURLOPT_PROXY`  to an empty string) solved the problem but due to a failing test a proposed final fix to unset it with `unset($conf[\CURLOPT_PROXY]);` does not actually do anything as the `$conf[\CURLOPT_PROXY]` option is never defined before being unset.

**How to reproduce**  
Create `http_proxy` & `https_proxy` ENV var pointing to a proxy.
Create `no_proxy` ENV var and make sure it does not include the host that we are about to send a test request to.
Set the `proxy` => `no` request option to the host of a request.
Send a request and the proxy sittings of the ENV will be used.

**Possible Solution**  
The following lines of code in the CurlFactory.php:448 could be changed to set the CURLOPT_NOPROXY to a requesting host string when a proxy is not required.

From:
```
                    if (isset($options['proxy']['no']) && Utils::isHostInNoProxy($host, $options['proxy']['no'])) {
                        unset($conf[\CURLOPT_PROXY]);
                    } else {
                        $conf[\CURLOPT_PROXY] = $options['proxy'][$scheme];
                    }
```
To:

```
                    if (isset($options['proxy']['no']) && Utils::isHostInNoProxy($host, $options['proxy']['no'])) {
                        $conf[\CURLOPT_NOPROXY] = $host;
                    } else {
                        $conf[\CURLOPT_PROXY] = $options['proxy'][$scheme];
                    }
```

**Additional context**  
Another solution is to force the `$conf[\CURLOPT_PROXY]` option to en empty string as it was originally proposed in https://github.com/guzzle/guzzle/pull/3117/commits/e0de379ca55e38bf06a019d46f0359b1ecba721a  but then the test CurlFactoryTest.php:194 should be updated from:

```
        if ($assertUseProxy) {
[...]

PR #3164, opened 2023-08-18
[8.0] Final for no proxy settings for all php versions

gergokee:
Fix for issue mentioned in https://github.com/guzzle/guzzle/issues/3163

### DIFF
--- a/src/Handler/CurlFactory.php
+++ b/src/Handler/CurlFactory.php
@@ -447,10 +447,11 @@ class CurlFactory implements CurlFactoryInterface
                     $host = $easy->request->getUri()->getHost();
                     if (isset($options['proxy']['no'])) {
                         if (Utils::isHostInNoProxy($host, $options['proxy']['no'])) {
-                            // To support older versions of php
+                            // To support older versions of php we need to set the proxy to an empty string
                             $conf[\CURLOPT_PROXY] = '';
                             $conf[\CURLOPT_NOPROXY] = $host;
                         } elseif (empty($options['proxy']['no'])) {
+                            $conf[\CURLOPT_PROXY] = $options['proxy'][$scheme];
                             $conf[\CURLOPT_NOPROXY] = '';
                         }
                     } else {
--- a/src/Handler/CurlFactory.php
+++ b/src/Handler/CurlFactory.php
@@ -445,10 +445,14 @@ class CurlFactory implements CurlFactoryInterface
                 $scheme = $easy->request->getUri()->getScheme();
                 if (isset($options['proxy'][$scheme])) {
                     $host = $easy->request->getUri()->getHost();
-                    if (isset($options['proxy']['no']) && Utils::isHostInNoProxy($host, $options['proxy']['no'])) {
-                        // To support older versions of php
-                        $conf[\CURLOPT_PROXY] = '';
-                        $conf[\CURLOPT_NOPROXY] = $host;
+                    if (isset($options['proxy']['no'])) {
+                        if (Utils::isHostInNoProxy($host, $options['proxy']['no'])) {
+                            // To support older versions of php
+                            $conf[\CURLOPT_PROXY] = '';
+                            $conf[\CURLOPT_NOPROXY] = $host;
+                        } elseif (empty($options['proxy']['no'])) {
+                            $conf[\CURLOPT_NOPROXY] = '';
+                        }
                     } else {
                         $conf[\CURLOPT_PROXY] = $options['proxy'][$scheme];
                     }
--- a/src/Handler/CurlFactory.php
+++ b/src/Handler/CurlFactory.php
@@ -453,6 +453,8 @@ class CurlFactory implements CurlFactoryInterface
                         } elseif (empty($options['proxy']['no'])) {
                             $conf[\CURLOPT_PROXY] = $options['proxy'][$scheme];
                             $conf[\CURLOPT_NOPROXY] = '';
+                        } else {
+                            $conf[\CURLOPT_PROXY] = $options['proxy'][$scheme];
                         }
                     } else {
                         $conf[\CURLOPT_PROXY] = $options['proxy'][$scheme];
--- a/src/Handler/CurlFactory.php
+++ b/src/Handler/CurlFactory.php
@@ -446,7 +446,9 @@ class CurlFactory implements CurlFactoryInterface
                 if (isset($options['proxy'][$scheme])) {
                     $host = $easy->request->getUri()->getHost();
                     if (isset($options['proxy']['no']) && Utils::isHostInNoProxy($host, $options['proxy']['no'])) {
-                        unset($conf[\CURLOPT_PROXY]);
+                        // To support older versions of php
+                        $conf[\CURLOPT_PROXY] = '';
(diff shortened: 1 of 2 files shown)


oostapov (approved):

GrahamCampbell:
Is this an issue with curl or PHP? I am reluctant to revert this and re-break the issue this was fixing. Do you know the exact last version of libcurl that has this issue, if this a PHP issue, the exact last version of PHP to have this issue?

---

I'll leave this PR on hold till we have that information.

oostapov:
> Is this an issue with curl or PHP? I am reluctant to revert this and re-break the issue this was fixing. Do you know the exact last version of libcurl that has this issue, if this a PHP issue, the exact last version of PHP to have this issue?
> 
> I'll leave this PR on hold till we have that information.

It is a curl issue, not php.
cURL version: 8.2.1 (hint: php -i | grep cURL) but it worked the same wrong on earlier 8.1 too

GrahamCampbell:
Ok, so what is the latest version of curl where this is **broken**, so we can decide if this is actually something we want to support? (not what is the latest version of curl)

oostapov:
As I said it is broken on the 8.2.1 which is currently latest stable. It has been broken and worked this way for years. 7.29 is the oldest I can test with and since that version it has never worked correctly.

oostapov:
I believe it has nothing to do with cURL as the cURL behaves quite predictable. It is an improper handling of the proxy settings by the guzzle. If we want to not use proxy for some domain we should force the curl having it in NO_PROXY setting instead of leaving it unset which leads to falling back to the ENV settings

gergokee:
@GrahamCampbell 

In short: my old fix was not a real fix. :/

Explanation:

The reason is that first i added `$conf[\CURLOPT_PROXY] = '';` . This should have been okay. But then tests failed, so i accidentally added this: `unset($conf[\CURLOPT_PROXY]);` . Which is not a real fix, since we wanted to specifically set  CURLOPT_PROXY to be empty, unsetting it does nothing :/ 

The reason now i want both `CURLOPT_PROXY =''` and `CURLOPT_NOPROXY` set to the added values cause in older php versions `CURLOPT_NOPROXY` is not working. So setting both is the ultimate solution.

GrahamCampbell:
I'm wondering if Guzzle 8.0 is the best place to fix this. I'm very conscious of breaking things for people.

src/Handler/CurlFactory.php
```diff
@@ -446,7 +446,9 @@ private function applyHandlerOptions(EasyHandle $easy, array &$conf): void
                 if (isset($options['proxy'][$scheme])) {
                     $host = $easy->request->getUri()->getHost();
                     if (isset($options['proxy']['no']) && Utils::isHostInNoProxy($host, $options['proxy']['no'])) {
-                        unset($conf[\CURLOPT_PROXY]);
+                        // To support older versions of php
+                        $conf[\CURLOPT_PROXY] = '';
+                        $conf[\CURLOPT_NOPROXY] = $host;
                     } else {
                         $conf[\CURLOPT_PROXY] = $options['proxy'][$scheme];
                     }
```
TravisWhitehead, line 461: Although this covers the case where our request matches the `['proxy']['no']` configuration (and we want it not to use the proxy), it misses an inverse case. :)

Image that in the environment variables there's `NO_PROXY="foo.bar"` / `no_proxy="foo.bar"`, and we want to configure guzzle to override/ignore this (to proxy everything, or at least proxy requests to `foo.bar`).

If we set:
```
'proxy' => [
    'http' => 'http://example:8080',
    'https' => 'http://example:8080',
    # Send everything through the proxy
    'no' => []
]
```

Then the environment's behavior will still get used and requests to `foo.bar` will (unexpectedly) not go through the proxy.

Perhaps something like this in the else branch?

```suggestion
                    } else {
                        $conf[\CURLOPT_PROXY] = $options['proxy'][$scheme];
                        $conf[\CURLOPT_NOPROXY] = "";
                    }
```

(Related comment on the issue: https://github.com/guzzle/guzzle/issues/3163#issuecomment-2065612465)

gergokee: hi, well, i think a little correction is needed here, cause your code would unset no proxy even if there is no no proxy value added, we only wanna unset if we set it and set it to be empty, but you are correct that we might need a third case:
'proxy' => [
    'http' => 'http://example:8080',
    'https' => 'http://example:8080',
]

This is the correct i think:
```
                    if (isset($options['proxy']['no'])) {
                        if (Utils::isHostInNoProxy($host, $options['proxy']['no'])) {
                            // To support older versions of php we need to set the proxy to an empty string
                            $conf[\CURLOPT_PROXY] = '';
                            $conf[\CURLOPT_NOPROXY] = $host;
                        } elseif (empty($options['proxy']['no'])) {
                            $conf[\CURLOPT_PROXY] = $options['proxy'][$scheme];
                            $conf[\CURLOPT_NOPROXY] = '';
                        } else {
                            $conf[\CURLOPT_PROXY] = $options['proxy'][$scheme];
                        }
                    } else {
                        $conf[\CURLOPT_PROXY] = $options['proxy'][$scheme];
                    }
```
What do you think @oostapov ?

GrahamCampbell:
While reviewing the PHPStan baseline for warnings that look like real runtime issues, I found a few proxy-related cases that seem best handled in this 8.0 PR rather than as separate 7.x patch changes.

The current baseline points at code paths where `proxy['no']` is passed directly to `Utils::isHostInNoProxy(string, array)`. That means malformed values, and potentially raw `NO_PROXY`-style strings, can fail with a raw `TypeError` before normal option validation. There is also a stream-handler path where a per-scheme proxy value is passed to `parse_proxy(string)` without validating it is a string.

Since this PR is already the 8.0 no-proxy behavior change, could we expand it to cover the full proxy option contract consistently?

Requested scope:

- Apply no-proxy handling and validation consistently in both the cURL and stream handlers.
- Decide the 8.0 contract for `proxy['no']`: either `string[]` only, matching the current docs, or `string|string[]` if we want to accept raw comma-delimited `NO_PROXY` values.
- If `proxy['no']` remains `string[]` only, reject non-arrays with a clear `InvalidArgumentException` before calling `Utils::isHostInNoProxy()`.
- If strings are accepted, normalize comma-delimited values intentionally and update docs/tests because the current docs tell users to pass `explode(',', getenv('NO_PROXY'))`.
- Validate per-scheme proxy entries before assigning them to cURL or passing them to the stream handler's `parse_proxy(string)`.
- Add matching tests for both `CurlFactory` and `StreamHandler`.

Suggested test coverage:

- No-proxy match disables proxy behavior as intended.
- No-proxy miss uses the configured proxy.
- Empty no-proxy list uses the configured proxy.
- Invalid `proxy['no']` fails with a deterministic `InvalidArgumentException`, not a raw `TypeError`.
- Invalid per-scheme proxy values fail with a deterministic `InvalidArgumentException`.
- If comma-delimited strings are accepted, both handlers normalize them consistently.

This keeps the behavior-changing proxy work in the existing 8.0 PR and avoids tightening proxy option handling in patch releases.

GrahamCampbell:
Closing this in favor of #3363, which carries the no-proxy fix forward on the 8.0 branch and also covers the related proxy option validation and stream-handler consistency work.

PR #3165, opened 2023-08-21
Added automatic closing of curl handles on CurlFactory object destruction

### LINKED ISSUE #3162: High memory usage when consistently using new HTTP client instances for each request
**Guzzle version(s) affected**: 7.7.0
**PHP version**: 8.2.7
**cURL version**: 7.88.1

**Description**  

I have a long running PHP script deployed within a docker service, which over its lifetime sends many HTTP requests. 
I have noticed that the memory usage reported by `docker stats` keep increasing and settles at around 2GB of used memory - If I remove the line sending the HTTP requests, then the memory stays low and stable.

Inspection with `memory_get_usage()` in the PHP runtime, does report a significantly lower memory usage than `docker stats` does.
I can reproduce a similar memory gap between the two tools, by simply creating multiple curl handles and never closing them - Leaving me to the thinking that the high memory usage is caused by hanging curl handles?

If I create a single guzzle client instance and reuse that client for all requests, then I do not see any large memory usage. 

**How to reproduce**  

Deploy at docker service with the following script
```
while (true) {
    $guzzle = new Client();

    try {
        $guzzle->get('https://some_url_you_do_not_mind_spamming');
    } catch (Throwable) {
        // Do nothing
    }
}
```

And inspect the memory usage with `docker stats`

The host system is Ubuntu 22.

**Possible Solution**  

It might be related to some curl handles not being closed?

**Additional context**  

I have attempted calling `gc_collect_cycles()` within my script, which does nothing for the high memory.
The high memory usage can also be measured on the host system using tools such as `btop` to measure the difference between "Used memory", "Available memory", "Cached memory" and "Free memory". Killing the docker service after it reaches the 2 GB peak, I can see a similar drop in "Used memory". No affect is present on "Cached memory".

nizacegodk:
Fixes https://github.com/guzzle/guzzle/issues/3162

### DIFF
--- a/src/Handler/CurlFactory.php
+++ b/src/Handler/CurlFactory.php
@@ -627,4 +627,12 @@ class CurlFactory implements CurlFactoryInterface
             return \strlen($h);
         };
     }
+
+    public function __destruct()
+    {
+        foreach ($this->handles as $id => $handle) {
+            \curl_close($handle);
+            unset($this->handles[$id]);
+        }
+    }
 }
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -57,7 +57,7 @@ parameters:
 
 		-
 			message: "#^Parameter \\#1 \\$ch of function curl_close expects resource, CurlHandle\\|resource given\\.$#"
-			count: 1
+			count: 2
 			path: src/Handler/CurlFactory.php
 
 		-
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -47,7 +47,7 @@
     <PossiblyFalseOperand occurrences="1">
       <code>$timeoutRequiresNoSignal</code>
     </PossiblyFalseOperand>
-    <PossiblyInvalidArgument occurrences="13">
+    <PossiblyInvalidArgument occurrences="14">
       <code>$easy-&gt;handle</code>
       <code>$easy-&gt;handle</code>
       <code>$easy-&gt;handle</code>
@@ -61,6 +61,7 @@
       <code>$resource</code>
       <code>$resource</code>
       <code>$sslKey</code>
+      <code>$handle</code>
     </PossiblyInvalidArgument>
     <PossiblyInvalidCast occurrences="1">
       <code>$sslKey</code>


src/Handler/CurlFactory.php
```diff
@@ -627,4 +627,12 @@ private function createHeaderFn(EasyHandle $easy): callable
             return \strlen($h);
         };
     }
+
+    public function __destruct()
+    {
+        foreach ($this->handles as $id => $handle) {
+            \curl_close($handle);
```
mfn, line 634: I assume you included `\curl_close` for PHP 7 compatibility?

Because according to https://www.php.net/manual/en/function.curl-close.php
> This function has no effect. Prior to PHP 8.0.0, this function was used to close the resource.

And according to your reported version https://github.com/guzzle/guzzle/issues/3162
> PHP version: 8.2.7

(so this line would be removed once Guzzle becomes PHP 8 only)

GrahamCampbell:
@nizacegodk Have you confirmed this PR works. According to the docs, this code doesn't do anything on PHP 8?

nizacegodk:
@GrahamCampbell I did some initial testing that concluded that the destructor has an effect, even on PHP 8. The initial testing showed the used memory reached very different levels. but it might just be that the garbage collector behaves differently than expected?

I will do some additional testing to further confirm it was not just a fluke.
What kind of proofs would be satisfactory? Would a reproducible docker image be acceptable?

GrahamCampbell:
Can you try it just with the unset?

nizacegodk:
I just ran the same test setup as initially, and when just using `unset()` the memory leak is still present.
My docker container ended up setteling at just over 2GB of ram as before any code changes.
image

When running with just `curl_close()` and without `unset()`, then the memory settles at around 430 MB. Initially I measured it to 250MB, but there seems to consistently be a short memory plateau around 250MB which made me stop the test too early initially.

image

The speed at which the memory increases is also radically different between having `curl_close()` and not having it. Leaving me with the strong impression that the php docs stating that `curl_close()` as having no effect is untrue.

I would also like to note that even 430MB is a lot of memory, and that I am unsure why such a high memory is still reached. No code except the code in the original issue is executed within the docker container, besides composer autoload.

```php
while (true) {
    $guzzle = new Client();

    try {
        $guzzle->get('https://some_url_you_do_not_mind_spamming');
    } catch (Throwable) {
        // Do nothing
    }
}
```

nizacegodk:
@GrahamCampbell 

I just had a breakthrough in debugging this issue. 
I modified the test code to as follows to get some GC statistics to monitor.

```php
while (true) {
    try {
        $guzzle = new Client();

        $guzzle->get('https://some_url_you_do_not_mind_spamming');
    } catch (Throwable $throwable) {
        // Do nothing
    } finally {
        gc_collect_cycles();

        echo print_r(gc_status(), true);
    }
}
```

When running the above code **WITHOUT** any destructor then the GC stats is as follows.
Runs equals the number of iterations of the while-loop.

Note the following:
- No objects are being collected.
- GC reports 0 root objects
- (Memory reaches almost 2GB)

image

When running the above code **WITH** the destructor then the GC stats is as follows.

Note the following:
- Objects are being collected.
- GC reports consistently 3 root objects
- (Memory is stuck at a low level of 27MB)

image

Leaving me again to the conclusion that the php docs stating that `curl_close()` as having no effect is untrue.

### Php version dump from container
```
php-shell:~/html$ php --version
PHP 8.2.7 (cli) (built: Jun 14 2023 05:12:47) (NTS)
Copyright (c) The PHP Group
Zend Engine v4.2.7, Copyright (c) Zend Technologies
    with Zend OPcache v8.2.7, Copyright (c), by Zend Technologies
    with Xdebug v3.2.1, Copyright (c) 2002-2023, by Derick Rethans
```

mfn:
> Leaving me again to the conclusion that the php docs stating that `curl_close()` as having no effect is untrue.

But the source… https://github.com/php/php-src/blob/de60872cfd229db42d600be343fb13b7ec9bb938/ext/curl/interface.c#L2895-L2912

In https://github.com/php/php-src/commit/b516566b84c210ce6ceddeb238e45d4b1bcb32ce#diff-29336a112e08c5b5caa7950afb6601010867f40a4643a33263646504277f7b33L3231 the actual functionality was removed (`zend_list_close()`)

nizacegodk:
I am sadly not very versed in C nor the PHP-src. But I should have no problem supplying a docker image which reproduces the issue. Using curl_close has an effect on the garbage collector in all of my tests.

It might just be a side effect, and not the close code itself, maybe calling the curl_close method somehow references the underlying curl handle in the context of the garbage collector thereby making it collectable.

As a sliver of data on this, notice the difference between the gc stats, when using curl close and when not.

I can supply a docker imagine tomorrow

mfn:
Indeed I agree this is "fishy" somehow!

nizacegodk:
@mfn @GrahamCampbell I figured out the "fishy" part of this issue.

It is caused by enabling the ElasticPHP APM Agent extension. 
Disabling said extension removes any memory leak from the test script - So they seem to have an issue to resolve on their end.

As a final note, might this not still be beneficial on PHP7.4? Since curl_close is expected there?

mfn:
I've no opinion on that TBH, but want to remark the calling any of those `*_close()` functions is always a bit tricky in case the resource is not valid anymore (for _whatever_ reason).

This then leads to weird issues that there may be an exception thrown on destructing that object, which could be anywhere else, which could be confusing.

What I want to say: in such delicate cases they should probably be protected via a try/catch to ignore any such cases.

And this all for PHP 7.4 neither I nor you are using :)

And since no one came forward running 7.4 and having this issue and it will be dropped eventually, I think just going for the `unset()` is the safer option, because it's void of any such problems.

GrahamCampbell:
I am ok with leaving this change in. It won't have much real-world impact on performance, and works around this issue with the tracing library. When PHP eventually removes this function, I presume in PHP 9, we can re-think the inclusion.

GrahamCampbell:
Thanks for all the effort here, everyone.

src/Handler/CurlFactory.php
andypost: as PHP 8.5 beta1 curl_close() throws deprecation warning https://github.com/php/php-src/commit/9b13bb1ae45f2e8abf6a702b64fd2333a23fc341

PR #3184, opened 2023-10-05
[8.0] Fix Utils::json*() for native `\JSON_THROW_ON_ERROR` option

TimWolla:
If a user passes the native `\JSON_THROW_ON_ERROR` option, these functions would throw a `\JsonException` instead of the expected `\InvalidArgumentException`.

### DIFF
--- a/src/Utils.php
+++ b/src/Utils.php
@@ -273,12 +273,11 @@ EOT
      */
     public static function jsonDecode(string $json, bool $assoc = false, int $depth = 512, int $options = 0)
     {
-        $data = \json_decode($json, $assoc, $depth, $options);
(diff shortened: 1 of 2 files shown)


TimWolla:
Oh, this only exists as of PHP 7.3. Would you consider it appropriate to drop support for PHP 7.2? It's at 1.8% for Guzzle 7 as per: https://packagist.org/packages/guzzlehttp/guzzle/php-stats#7

GrahamCampbell:
We can consider this for Guzzle 8. AWS and others need Guzzle 7 to retain PHP 7.2 support. We're also still fighting to get people to upgrade from Guzzle 5 and 6, so dropping more PHP versions is just gonna make it harder, or give us two 7.x lines to provide security fixes for.

PR #3189, opened 2023-11-02
Add another `base_uri` example in documentation

chapa:
It's not that intuitive that `http://api.com/v1/` + `/resources` would result in `http://api.com/resources` so I think it's good to mention it.

It could actually be merged in an older branch but I didn't know which one to choose, tell me if I should change the target branch.


GrahamCampbell:
Thanks.

PR #3201, opened 2024-03-15
[PHP 8.4] Fixes for implicit nullability deprecation

Ayesh:
Fixes all issues that emits a deprecation notice on PHP 8.4.

See:
 - [RFC](https://wiki.php.net/rfc/deprecate-implicitly-nullable-types)
 - [PHP 8.4: Implicitly nullable parameter declarations deprecated](https://php.watch/versions/8.4/implicitly-marking-parameter-type-nullable-deprecated)

### DIFF
--- a/src/BodySummarizer.php
+++ b/src/BodySummarizer.php
@@ -11,7 +11,7 @@ final class BodySummarizer implements BodySummarizerInterface
      */
     private $truncateAt;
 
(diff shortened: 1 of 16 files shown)


GrahamCampbell:
I've made the fixes via upgrading php-cs-fixer. Thanks anyway.

Issue #2036, opened 2018-03-29
Allow passing certificates as strings


jbaron-gingco:
| Q            | A
| ------------ | ---
| Bug?         | no
| New Feature? | yes
| Version      | 6.3.2 on Ubuntu 17.10 with self-built PHP 7.1.5

TLDR: It would be nice if PEM data could be passed directly as string in the  `cert` and `ssl_key` options.

The "cert" and "ssl_key" options currently require a path to a PEM encoded file as value. This requires that the certificate and its key are stored in a locally accessible storage. Sometimes one does not want that, but only wants to hold certificates in memory (e.g. they are retrieved from some configuration distribution service, from an environment variable, or similar).

It does not seem possible to pass the certificates as a string. I've played around with stream wrappers (e.g. using a file name like `data://text/plain;base64,SSBsb3ZlIFBIUAo=`), but that results in a "file not found" error. This error seems to come directly from OpenSSL or curl.

sagikazarmark:
The problem is that libcurl has no API for passing keys as strings (only as files). So this solution would require Guzzle to temporarily store them as files which would involve a lot of checks, cleanups, etc and it's not something that Guzzle should do.

I'm afraid this is not something that we can implement in Guzzle right now.

LeoMachado94:
Isso já é possível hoje @sagikazarmark ? Estou querendo passar como string..

PR #3202, opened 2024-03-28
Add options to allow passing certificates by string

bernhard-efler:
Since PHP 8.1 [(URL)](https://github.com/php/php-src/pull/7194) and curllib 7.71.0 it is possible to pass SSL Certificates as strings.

To support this two new request options are created.
I since it would be more convinent to change the current options, but that would require a breaking change.

This feature was requestes four years ago in #2036, but at the time curl was ned capable.

### DIFF
--- a/docs/request-options.rst
+++ b/docs/request-options.rst
@@ -1109,6 +1109,28 @@ SSL certificates can be found on the
 `cURL website <http://curl.haxx.se/docs/sslcerts.html>`_.
 
 
+.. _verify_blob-option:
+
+verify_blob
+------
+
+:Summary: Specify the CA bundle to use for SSL certificate verification. When this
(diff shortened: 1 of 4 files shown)


GrahamCampbell:
Unfortunately still not enough time has passed for this to be viable, for the same reasons as previously. I think we should come back to this in another 4 years once we can be sure everyone has a new enough PHP and curl version. I don't really want some Guzzle functionality to fail at runtime because it's not supported (nor to implement a polyfill for the reasons discussed in the linked PR). Thanks anyway.

monterhealth:
This would be a super - distinctive - feature. 2028 still sounds far away :-( The _blob in the configuration seems clearly distinguishable to me from the standard cert and ssl_key options.

PR #3215, opened 2024-04-13
[7.9] Improve `Accept-Encoding` documentation

jtojnar:
https://github.com/guzzle/guzzle/commit/1d46cec19f74f57fe8be671214b4ecf183e5d63f tried to improve the comment but the new content implies an empty header will be sent.

That is not the case. Quoting from <https://curl.se/libcurl/c/CURLOPT_HTTPHEADER.html>:

> If you add a header with no content as in 'Accept:' (no data on the right side of the colon), the internally used header is disabled/removed.

Let’s clarify the point is removing the header to get the default '*'
<https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Encoding#sect2>:

> `*`: Matches any content encoding not already listed in the header. This is the default value if the header is not present. This directive does not suggest that any algorithm is supported but indicates that no preference is expressed.

As was probably intended by the original change that introduced this: https://github.com/guzzle/guzzle/commit/1a9ad6b55368283aaf73672a519e5177530991fb

### DIFF
--- a/docs/request-options.rst
+++ b/docs/request-options.rst
@@ -400,6 +400,28 @@ header of the request.
     // Pass "gzip" as the Accept-Encoding header.
     $client->request('GET', '/foo.js', ['decode_content' => 'gzip']);
 
+.. warning::
+
+    The ``Accept-Encoding`` header will not be sent unless you provide it
+    explicitly, or pass a string value to ``decode_content``. That is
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
Thanks everyone. 🎉

PR #3218, opened 2024-05-19
The CurlMultiHandle causes a busy immediately when the first byte is received

frodeborli:
I've provided a patch to the CurlMultiHandler which leverages the event loop. This allows TaskQueue implementations to properly sleep between each invocation of the \curl_multi_exec, and added a curl_multi_select() statement inside the busy wait loop that is meant to "wrap up" things.

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -117,7 +117,7 @@ parameters:
 
 		-
 			message: "#^Parameter \\#1 \\$mh of function curl_multi_exec expects resource, CurlMultiHandle\\|resource given\\.$#"
-			count: 1
+			count: 2
 			path: src/Handler/CurlMultiHandler.php
 
 		-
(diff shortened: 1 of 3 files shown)


frodeborli:
I'm not sure what I can do with the PHPStan tests. They seem to be unrelated to my patch.

GrahamCampbell:
It's probably because you ran composer and phpstan with a different version of PHP. ~You'd need to use our Makefile. e.g. `make static-psalm-update-baseline`.~ You need to use PHP 7.4.

frodeborli:
I made the commit to the 7.8 branch. Should I also make it for the 7.9 and 8.0?

GrahamCampbell:
> I made the commit to the 7.8 branch. Should I also make it for the 7.9 and 8.0?

No.

GrahamCampbell:
I'm still aware of of this PR, and appreciate the time you took to prepare this. I'm hoping to get to this soon, so this can be included in the next release(s).

GrahamCampbell:
Thanks for working on this. Moving this to #3234.

PR #3224, opened 2024-06-19
proposal timeout type on CurlFactory.php

nao329:
could you please consider about my proposal commit!
cause when I  use this library on my project.
I had trouble about the error of [2024-06-19 19:55:02] local.ERROR: TypeError: Unsupported operand types: string * int in /var/www/vendor/guzzlehttp/guzzle/src/Handler/CurlFactory.php:420


abublihi:
we have the same issue 'types: string * int' 

how to solve it?

GrahamCampbell:
We do not support passing a string there. This is an error in the code on your side.

PR #3229, opened 2024-07-10
Redact the userinfo in cURL error exception messages

TimWolla:
I think this would technically qualify as a security issue, but given that it cannot be actively exploited by an attacker, I'm filing this as a public PR. You might want to consider publishing an advisory nevertheless.

### DIFF
--- a/src/Utils.php
+++ b/src/Utils.php
@@ -381,4 +381,18 @@ EOT
 
         throw new \Error('ext-idn or symfony/polyfill-intl-idn not loaded or too old');
     }
+
+    /**
+     * Redacts the password in the userinfo part of a UriInterface.
+     */
+    public static function redactUserinfoInUri(UriInterface $uri): UriInterface
+    {
+        $userInfo = $uri->getUserInfo();
+
+        if (false !== ($pos = \strpos($userInfo, ':'))) {
+            return $uri->withUserInfo(\substr($userInfo, 0, $pos), '***');
+        }
+
+        return $uri;
+    }
 }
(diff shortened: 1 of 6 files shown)


GrahamCampbell (changes requested):
The baseline file needs to be generated using the same version of PHP as the CI.

TimWolla:
@GrahamCampbell CI is now green, thank you.

src/Utils.php
```diff
@@ -381,4 +381,18 @@ private static function idnToAsci(string $domain, int $options, ?array &$info =
 
         throw new \Error('ext-idn or symfony/polyfill-intl-idn not loaded or too old');
     }
+
+    /**
+     * Redacts the password in the userinfo part of a UriInterface.
+     */
+    public static function redactUserinfoInUri(UriInterface $uri): UriInterface
```
GrahamCampbell, line 388: Probably this function should go in the PSR package. We can add it to only 2.x, and drop support for psr 1.x in guzzle 7.9.

GrahamCampbell: Alternatively, we could view this as a breaking change, and just do it in Guzzle 8.

TimWolla: I consider this change a bug fix rather than a new feature (see also the PR description that this technically is a security fix). So it should probably go into Guzzle 7.8. 

Nevertheless I created a PR guzzle/psr7#607

Should I mark this function in guzzle/guzzle as `@internal` until it can be replaced by the one in PSR-7?

GrahamCampbell:
I'll try and get to this, this weekend. 👍

PR #3231, opened 2024-07-17
add validation for `timeout` and `connect_timeout`

esler:
Fixes the problem with rouding values lower than 1ms to zero, resulting in unexpected behaviour.


GrahamCampbell:
This will be fixed in 8.0.0.

PR #3238, opened 2024-07-19
[7.9] Fix HTTP 2.0 - TLS 1.3 check

pcf0:
I think there was a typo here and the TLS 1.3 method should actually be used for checking?


GrahamCampbell:
Thanks, yes, this was a typo.

PR #3252, opened 2024-09-09
If the request is cancelled, the handle may not be returned

alex621:
Need to skip it if the handle does not exist

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -110,6 +110,9 @@
         "sort-packages": true
     },
     "extra": {
+        "branch-alias": {
+            "dev-patchhandle": "7.9.x-dev"
+        },
         "bamarni-bin": {
             "bin-links": true,
             "forward-command": false
--- a/src/Handler/CurlMultiHandler.php
+++ b/src/Handler/CurlMultiHandler.php
@@ -253,7 +253,7 @@ class CurlMultiHandler
                 continue;
             }
             // It may not return the handle if it was cancelled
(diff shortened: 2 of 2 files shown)


composer.json
```diff
@@ -110,6 +110,9 @@
         "sort-packages": true
     },
     "extra": {
+        "branch-alias": {
```
GrahamCampbell, line 113: Please revert this. You can perform this aliasing in your own project when requiring guzzle.

GrahamCampbell (changes requested):

src/Handler/CurlMultiHandler.php
```diff
@@ -252,6 +252,10 @@ private function processMessages(): void
                 // if it's not done, then it would be premature to remove the handle. ref https://github.com/guzzle/guzzle/pull/2892#issuecomment-945150216
                 continue;
             }
+            // It may not return the handle if it was cancelled
```
GrahamCampbell, line 255: Please add test coverage for this change.

GrahamCampbell (changes requested):

GrahamCampbell:
FYI @TimWolla fixed this in the PHP core with https://github.com/php/php-src/pull/16302. Since we support a wider range of PHP versions, I'm going to apply a mitigation inside Guzzle. Your PR was a good start, but included unrelated changes and no tests. I've made a replacement at https://github.com/guzzle/guzzle/pull/3443.

Issue #3250, opened 2024-09-03
CurlFactory no longer sets a default protocolVersion when the request object has protocol version as an empty string


lucasnetau:
**Guzzle version(s) affected**: 7.9.x
**PHP version**: 8.2
**cURL version**: Any

**Description**  
Since 7.9 the CurlHandler does not set a default protocol version is protocol version is empty. The error message given is not clear since no version is set.

I encountered this with a external library that initialised the Request with protocol version as a empty string, A NetworkException was thrown with the following error message: "HTTP/ is not supported by the cURL handler." which made tracking down the issue slower than it should have been.

**How to reproduce**  
$request->withProtocolVersion('')

**Possible Solution(s)**  
Not sure what the best solution is, three possible paths I can think of:

* 7.9 introduced a BC break with 7.8 which previously defaulted to '1.0', so this default value could be used if protocol version is empty.
* MessageTrait->withProtocolVersion could Exception preventing this value from being set
* Handle the case of an empty protocolVersion in CurlFactory::create to say 'Protocol version is not set for cURL handler'

**Additional context**  

7.9 (Doesn't handle the empty string case):
https://github.com/guzzle/guzzle/blob/d281ed313b989f213357e3be1a179f02196ac99b/src/Handler/CurlFactory.php#L50-L58

7.8 (Defaults to 1.0):
https://github.com/guzzle/guzzle/blob/f4152d9eb85c445fe1f992001d1748e8bec070d2/src/Handler/CurlFactory.php#L234-L241

GrahamCampbell:
How are you constructing a request such that it has an invalid request version?

GrahamCampbell:
Ah, `$request->withProtocolVersion('')`. I believe your PR makes the error clearer. Maybe we should instead just treat empty version as if it were 1.0, though.

takutakuaoao:
@GrahamCampbell 

Thank you for confirming the contents.
I have updated the PR to set 1.0 protocol version if the request protocol is empty.

https://github.com/guzzle/guzzle/pull/3257#issuecomment-2398696562

PR #3257, opened 2024-10-03
Fixed to set the default protocol version if the request has an empty-letter protocol version.

takutakuaoao:
Added exception handling so that CurlFactory can return appropriate error messages.
The issue on which this pull request is based is #3250.

When using the create method in CurlFactory, if the first argument, $request, was set to an empty protocol version, a slightly confusing error message was displayed, "HTTP/ is not supported by the cURL handler.".

With this fix, if an empty protocol version is set, it now returns the error message "Protocol version is not set for the cURL handler".

---

_This text was translated from Japanese to English at DeepL._


takutakuaoao:
#3250

takutakuaoao:
Changed to set default protocol version if request protocol version is empty. #3250

PR #3274, opened 2025-03-20
fixes #3273: SetCookie read bool pieces as non-bool bugfix

### LINKED ISSUE #3273: SetCookie::setSecure requires bool, but cookie accepts non-bool
**Guzzle version(s) affected**: 7.9.2
**PHP version**: 8.3.14
**cURL version**: 7.81.0

**Description**  

Some websites apparently send a `Set-Cookie` header with `; Secure=true` instead of just `; Secure`, and Guzzle reads that `true` like a value, and then it passes that value (the word `"true"`) to `setSecure()`, and that's not allowed, because only bools accepted.

They say never trust user input. A cookie header is user input for an http client. Don't rely on websites doing that correctly. I've never seen this before, but there's 1 website in the 15k that I request every night that does this: https://www.praatmeelelystad.nl/

    set-cookie: mett-cor-id=hexdechexdechexdechexdec; Path=/; HttpOnly; Secure=true; SameSite=Lax;

**How to reproduce**  

```php
$guzzle = new Client();
$guzzle->get('https://www.praatmeelelystad.nl/home');
```

**Possible Solution**  

Guzzle should not rely on the cookie params format, but hard-code accepted formats. `HttpOnly` is always a bool. `Secure` is always a bool. Etc. That's what real browsers do too. `Secure` is the same as `Secure=1` and `Secure=true` and `Secure=hellno` and `Secure=0`.

**Additional context**

rudiedirkx:
fixes #3273 by checking the default value for boolness. Bool in defaults -> must be `true` if it exists.

### DIFF
--- a/src/Cookie/SetCookie.php
+++ b/src/Cookie/SetCookie.php
@@ -47,7 +47,7 @@ class SetCookie
         foreach ($pieces as $part) {
             $cookieParts = \explode('=', $part, 2);
             $key = \trim($cookieParts[0]);
-            $value = isset($cookieParts[1])
+            $value = isset($cookieParts[1]) && !is_bool($data[$key] ?? null)
                 ? \trim($cookieParts[1], " \n\r\t\0\x0B")
                 : true;
 


GrahamCampbell:
Working on an alternative fix at #3281.

Issue #3275, opened 2025-03-21
Passing null to parameter #2 ($numeric_prefix) of type string is deprecated


rekmixa:
**Guzzle version(s) affected**: 6.5.8
**PHP version**: 8.3.17
**cURL version**: 7.88.1

**Description**  
Passing null to parameter #2 ($numeric_prefix) of type string is deprecated in vendor/guzzlehttp/guzzle/src/Client.php:445

https://github.com/rekmixa/guzzle/blob/0d137e94480b275aadd1f2536a76d91cf580711c/src/Client.php#L445

error occured because of passing null as second argument in `http_build_query` function now deprecated. I replaced it with an empty string.

rekmixa:
PR - https://github.com/guzzle/guzzle/pull/3276

GrahamCampbell:
Guzzle 6 is EOL and no longer maintained. Note also that Guzzle 6 does not support PHP 8. Guzzle 7 is the first version of Guzzle to support PHP 8. Please upgrade to Guzzle 7 ASAP.

GrahamCampbell:
Indeed, Guzzle 7.8.0 is the first version to support PHP 8.3.0.

PR #3276, opened 2025-03-21
#3275 replace null with empty string in http_build_query in Client

rekmixa:


GrahamCampbell:
Thanks for the PR, but Guzzle 6 is EOL and no longer maintained. Note also that Guzzle 6 does not support PHP 8. Guzzle 7 is the first version of Guzzle to support PHP 8. Please upgrade to Guzzle 7 ASAP.

PR #3278, opened 2025-03-24
Remove explicit content-length header for GET requests

mfn:
## Summary
Fixes #3277 and reverts #1189

Originally this was introduced in #1189 to fix https://github.com/aws/aws-sdk-php/issues/701

In my case however, I receive a 500 when trying to stream content from media.licdn.com because of this header.

I then went back and tried the example from the aws-sdk-php and removed that header only to learn it works as expected.

My conclusion therefore is that in the past 10 years either AWS server or the SDK changed and this workaround, once needed, no longer is necessary and we can remove it.

### Update
- The final PR only removes the content-length header for `GET` requests, keeps it for `POST` & `PUT`

PR #3279, opened 2025-03-24
Run php-cs-fixer on whole codebase

mfn:
## Summary
The CI workflow only runs on the diff, but when you run it locally it showed lots of places which are not up2date (presumably due to changes in php-cs-fixer over time).

I had a violation in #3278 and ran php-cs-fixer locally and got all these file changes reported.


mfn:
I see, so I guess running fixer with PHP 8.3 produces invalid changes for 7.4 et all

PR #3280, opened 2025-03-24
`testHandlesGarbageHttpServerGracefully`: Split test between PHP <= 8.1 and > 8.1

mfn:
## Summary
This fixes the failing tests recently observed, e.g.:
- https://github.com/guzzle/guzzle/actions/runs/13981481062/job/39147439887
- https://github.com/guzzle/guzzle/actions/runs/14036887147/job/39296947529

It seems the behaviour of `fopen()` changed with newer PHP version already returning `false` instead of a resource, presumably due to some HTTP header probing/parsing.

Seeing the behaviour change affects > 8.1, which is the last one to also receive security patches, I guess it's one of the recent security fixes applied.

### DIFF
--- a/tests/Handler/StreamHandlerTest.php
+++ b/tests/Handler/StreamHandlerTest.php
@@ -750,12 +750,33 @@ class StreamHandlerTest extends TestCase
         self::assertFalse(\feof($body));
     }
 
-    public function testHandlesGarbageHttpServerGracefully()
+    /**
+     * @requires PHP <=8.1
+     */
+    public function testHandlesGarbageHttpServerGracefullyLegacy()
     {
         $handler = new StreamHandler();
 
         $this->expectException(RequestException::class);
         $this->expectExceptionMessage('An error was encountered while creating the response');
+
+        $handler(
+            new Request('GET', Server::$url.'guzzle-server/garbage'),
+            [
+                RequestOptions::STREAM => true,
+            ]
+        )->wait();
+    }
+
+    /**
+     * @requires PHP >8.1
+     */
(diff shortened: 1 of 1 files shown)


mfn:
No idea what's up with the Windows 8.4 build and why it doesn't match the Linux behaviour (but lower PHP windows versions do) 🤷🏼 

But, at least the Linux tests are now all ✅ again; better then nothing I guess.

GrahamCampbell:
It's because it's not using the latest PHP 8.4 version.

GrahamCampbell:
These changes were only made this month, to the very latest batch of PHP versions, which is why we are only now encountering these test failures.

GrahamCampbell (changes requested):
Can we check, instead:
* PHP >=8.1.32,<8.2
* PHP >=8.2.28,<8.3
* PHP >=8.3.19,<8.4
* PHP >=8.4.5

Then CI will pass.

mfn:
Sorry, I don't understand what you're asking for?

GrahamCampbell:
Checking PHP >=8.1 is incorrect. We need to only run the newer test if the any of the following conditions are met:

* PHP >=8.1.32,<8.2
* PHP >=8.2.28,<8.3
* PHP >=8.3.19,<8.4
* PHP >=8.4.5

mfn:
Green `\o/`

I thought I can achieve it with `@requires`, but no chance, therefore the solution is "in code" directly 🤞🏼 

ps: I did add temporary debug statements to ensure that indeed the legacy/non-legacy version is executed on the correct version

Issue #3271, opened 2025-03-13
Gadget Chain in FileCookeJar allows arbitrary File Write


mcdruid:
**Guzzle version(s) affected**: 4.0.0-rc.2 <= 7.5.0+
**PHP version**: n/a
**cURL version**: n/a

**Description**  

This is a Security issue, but it's not directly exploitable; a Gadget Chain is a technique that attackers can use to leverage a PHP Object Injection vulnerability if they find one.

I emailed security@tidelift.com (per the project's security policy) before filing this and was advised to file a public issue as the details of the Gadget Chain have been public for a long time.

**How to reproduce**  

https://github.com/ambionics/phpggc/tree/master/gadgetchains/Guzzle/FW/1

```
./phpggc -l guzzle

Gadget Chains
-------------

NAME                     VERSION                         TYPE                  VECTOR        I    
Guzzle/FW1               4.0.0-rc.2 <= 7.5.0+            File write            __destruct 
```

The File Write chain still works in the current version. It (ab)uses `FileCookeJar`'s destructor.

**Possible Solution**  

One approach that's often used to defend against Gadget Chains is to implement a `__wakeup` method that tries to block deserialization if there's no legitimate reason for it.

psr7 does this in https://github.com/guzzle/psr7/blob/2.7/src/FnStream.php#L61

If the FileCookeJar class should allow serialization, one way of mitigating abuse would be to try to specifically prevent the class being used to write out executable PHP scripts, as that's one of the main use cases of the Gadget Chain from the attacker's point of view.

For example, Joomla tries to do something along those line here:

https://github.com/joomla/joomla-cms/blob/6.0-dev/libraries/src/Log/Logger/FormattedtextLogger.php#L250

(Although because that relies on a class property it was possible to disable the protection from a Gadget Chain - see https://github.com/ambionics/phpggc/pull/202/ ).

Perhaps there's other validation that could be done to ensure only legitimate cookiejar content is written.

**Additional context**  

Apologies to Pentesters and Bug Hunters; this Gadget Chain is useful for PHP Object Injection Proof of Concepts :)

PR #3286, opened 2025-05-16
initial idea for adding protection against file write gadget chain

mcdruid:
Context: https://github.com/guzzle/guzzle/issues/3271

Here's a simple (/crude) way to avoid this class being used to write out a file containing malicious PHP.

The idea is similar to what joomla does here (edit - fixed link):

https://github.com/joomla/joomla-cms/blob/6.1.0/libraries/src/Log/Logger/FormattedtextLogger.php#L250

This wouldn't prevent the arbitrary file write, but should mean that the file cannot easily be used to execute a malicious PHP payload.

https://stackoverflow.com/questions/244777/can-comments-be-used-in-json for context (/ bikeshedding) on how to add a comment to a json file.

### DIFF
--- a/src/Cookie/FileCookieJar.php
+++ b/src/Cookie/FileCookieJar.php
@@ -56,7 +56,7 @@ class FileCookieJar extends CookieJar
      */
     public function save(string $filename): void
     {
-        $json = [];
+        $json = ['_comment' => "//<?php die('Security');?>"];
         /** @var SetCookie $cookie */
         foreach ($this as $cookie) {
             if (CookieJar::shouldPersist($cookie, $this->storeSessionCookies)) {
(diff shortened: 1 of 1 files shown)


mcdruid:
It seems the original issue I filed about this has been deleted.

The Gadget Chain is still there in the PHPGGC project (has been since 2017):

https://github.com/ambionics/phpggc/tree/master/gadgetchains/Guzzle/FW/1

... which is why I considered it appropriate to discuss in public.

https://www.drupal.org/project/drupal/issues/3524971 is looking at ways Drupal might mitigate this if Guzzle does not.

Do the maintainers have any feedback / response - even if it's a case of "do not unserialize unsafe data" / "won't fix"?

ohader:
> https://github.com/joomla/joomla-cms/blob/6.0-dev/libraries/src/Log/Logger/FormattedtextLogger.php#L250

https://github.com/joomla/joomla-cms/blob/28fb5935fe080eaee1eec8d431362577cddeb7a5/libraries/src/Log/Logger/FormattedtextLogger.php#L254

GrahamCampbell:
I'm considering a different fix, and targetting only the next major version only: https://github.com/guzzle/guzzle/pull/3334.

mcdruid:
Thanks @GrahamCampbell !

PR #3289, opened 2025-06-09
Add support for HTTP/3

Synchro:
PHP 8.4 has support for HTTP/3, but it's dependent on the underlying libcurl having support for it, which is not enabled by default; see [curl docs](https://curl.se/docs/http3.html).

Nevertheless, we can add HTTP/3 support so that it will be usable if a PHP build has it enabled.

This PR adds a check for whether HTTP/3 is available (it also requires TLSv1.3), adds support for the `version` option to be `3` or `3.0`, provides integration with PHP's constants, and includes check for the presence of curl HTTP/3 support in both runtime and test configs, so it won't break in old PHP versions.

While I was there, I noticed that HTTP/2 wasn't checked in `\GuzzleHttp\Test\Handler\CurlFactoryTest::testProtocolVersion`, so I added that as well.

### DIFF
--- a/src/Handler/CurlFactory.php
+++ b/src/Handler/CurlFactory.php
@@ -49,7 +49,11 @@ class CurlFactory implements CurlFactoryInterface
     {
         $protocolVersion = $request->getProtocolVersion();
 
-        if ('2' === $protocolVersion || '2.0' === $protocolVersion) {
+        if ('3' === $protocolVersion || '3.0' === $protocolVersion) {
+            if (!self::supportsHttp3()) {
+                throw new ConnectException('HTTP/3 is supported by the cURL handler, however libcurl is built without HTTP/3 support.', $request);
+            }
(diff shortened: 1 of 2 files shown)


Synchro:
Ick. Running into all kinds of problems with psalm and phpstan baselines.

Synchro:
I finally got past the phpstan and psalm checks without adding to the baselines. Should I squash and force-push to clean up these commits?
The test errors are inconsistent (a different set of them fails each run) and seem to have to do with the test config, not anything that I've touched.

Synchro:
I squashed everything

Synchro:
Any comments on this? Do you know why the test runner is failing to connect?

Synchro:
I'd still like to see this included.

GrahamCampbell:
Thanks for getting started here. This PR appears incomplete. I'll be working on a replacement PR for Guzzle 8 soon.

PR #3290, opened 2025-06-09
Fix everything that php-cs-fixer complains about

Synchro:
The php-cs-fixer policy doesn't seem to be enforced, though it's checked in CI. This PR simply fixes everything that it complains about, so it will no longer report any issues, providing a clean baseline for future changes.


GrahamCampbell:
You need to run PHP-CS-Fixer on PHP 7.4 please.

Synchro:
Sorry! I just realised that...

PR #3293, opened 2025-07-04
CurlFactory: Move logic for creating and storing curl handles to interface

MrMeshok:
This PR introduces `CurlHandlePool` and `CurlHandlePoolInterface` which gives developer an option to customize logic for creating and storing curl handles.

# Reason
I have concurrency heavy app that powered by Swoole and I want to replace default implementation with basic array with my Swoole specific implementation for better performance and concurrency control.
But with current implementation of `CurlFactory` my only option is to fully reimplement `CurlFactoryInterface` with small edits to storing and creating curl handles

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -487,12 +487,6 @@ parameters:
 			path: src/Handler/CurlFactory.php
 
 		-
-			message: '#^Parameter \#1 \$ch of function curl_close expects resource, CurlHandle\|resource given\.$#'
-			identifier: argument.type
-			count: 2
-			path: src/Handler/CurlFactory.php
-
-		-
 			message: '#^Parameter \#1 \$ch of function curl_error expects resource, CurlHandle\|resource given\.$#'
 			identifier: argument.type
 			count: 1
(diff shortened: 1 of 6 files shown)


GrahamCampbell:
Thanks for the PR and for explaining the Swoole use case. I know this was closed by the stale bot, but I have come back too look over this and give it some real consideration. After looking at this against the current Guzzle 8 handler design, I do not think we should add a public cURL handle pool abstraction to Guzzle.

In Guzzle 8, the cURL handlers have stricter ownership and lifecycle semantics: factories are closed only when owned by the handler, multi-handler cancellation/close paths can discard active handles, request-level cURL options that conflict with Guzzle-managed state are rejected, and cURL share handling is managed by `CurlFactory`. Making handle storage user-replaceable would make that low-level lifecycle part of Guzzle’s public API.

To support this safely, we would need to define release vs discard behavior, explicit close semantics, share-handle compatibility, PHP 7 resource vs PHP 8 object handling, callback cleanup, and multi-handler cancellation behaviour. That is a lot of API surface and maintenance burden for a very specialised runtime optimisation.

For Swoole/coroutine-specific behaviour, I think the better fit is either a Swoole-native HTTP client or a fully custom Guzzle handler/factory owned by the application. Guzzle’s built-in handlers should stay focused on the standard cURL and stream transports.

PR #3299, opened 2025-08-20
feat(retry): enhance RetryMiddleware with Retry-After support and con…

rishi111-code:
Enhances RetryMiddleware in Guzzle HTTP client to support rate-limit awareness and customizable retry logic:

Recognizes and honors the Retry-After header (numeric and HTTP-date formats) for accurate wait times.

Adds configurable maxRetries and baseDelay (in ms) for exponential backoff control.

Retains existing support for connection errors and server (5xx/429) retries.

Why It Matters
This upgrade delivers smarter and more resilient HTTP clients:

Adapts to server-imposed delays (Retry-After), reducing unnecessary retries.

Offers flexible defaults — maxRetries = 3, baseDelay = 100 ms — while allowing overrides.

Reduces boilerplate: developers no longer need to build custom retry logic from scratch.

Key Changes

Modified src/RetryMiddleware.php:

Introduced maxRetries and baseDelay parameters.

Enhanced __invoke() flow to parse Retry-After header and implement delay logic.

Ensured delay continues via exponential backoff when Retry-After isn’t present.

Updated tests/RetryMiddlewareTest.php:

Added new test testRespectsRetryAfterHeader to validate Retry-After handling.

Ensured retries occur as expected when the header exists.

Test Coverage
Includes existing and new tests verifying:

Default retry behavior on connection failures and server errors.

Correct delay and retry count in exponential backoff scenarios.

Respect for Retry-After, ensuring the middleware waits properly before retrying.

Compatibility
Fully backward-compatible. No breaking changes—just enhancements layered onto existing behavior.

### DIFF
--- a/src/RetryMiddleware.php
+++ b/src/RetryMiddleware.php
@@ -1,60 +1,64 @@
 <?php
 
+declare(strict_types=1);
+
 namespace GuzzleHttp;
 
 use GuzzleHttp\Promise as P;
 use GuzzleHttp\Promise\PromiseInterface;
 use Psr\Http\Message\RequestInterface;
 use Psr\Http\Message\ResponseInterface;
+use GuzzleHttp\Exception\ConnectException;
+use InvalidArgumentException;
 
 /**
- * Middleware that retries requests based on the boolean result of
- * invoking the provided "decider" function.
- *
- * @final
+ * Enhanced RetryMiddleware with:
+ * - configurable max retries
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
It seems that there is low demand for this, and/or it can be fulfilled by a third party package. Thanks anyway, but we're going to pass on this for now.

PR #3308, opened 2025-12-02
Add explicit return type declarations

lacatoire:
This PR adds explicit return types to several methods where they were previously implicit.
Happy to adjust anything if needed.


GrahamCampbell:
Thanks for the PR, but many of these changes are unnecessary and/or breaking.

PR #3310, opened 2025-12-04
Allow Bearer Authentication with cURL handler

### LINKED ISSUE #3311: Add support for Bearer tokens
**Description**
Add support for Bearer tokens for authentication.

**Example**
```
    $client->request('GET', '/get', [
        'auth' => ['token', '', 'bearer']
    ]);
```

**Additional context**
We are trying to make nextcloud use Bearer tokens for authentication in webdav requests, in an on-going effort to implement the latest OCM (Open Cloud Mesh [1]) spec in nextcloud.

1. https://datatracker.ietf.org/doc/draft-ietf-ocm-open-cloud-mesh/

enriquepablo:
Here we add a new authentication type for Bearer auth.

Fixes #3311

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 Please refer to [UPGRADING](UPGRADING.md) guide for upgrading to a major version.
 
+### Added
+
+- Add support for Bearer tokens
+
 ## 7.10.0 - 2025-08-23
 
 ### Added
--- a/docs/request-options.rst
+++ b/docs/request-options.rst
@@ -194,7 +194,7 @@ bearer
 
 .. note::
 
-    This is currently only supported when using the cURL handler.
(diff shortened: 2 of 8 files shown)


Nyholm (approved):
Thank you. 

I think it looks good. @GrahamCampbell do you want to give a second pair of eyes on this?

GrahamCampbell:
I don't understand why we'd add this, vs using headers which would work properly even when the curl backend is not used.

enriquepablo:
@GrahamCampbell We want to use this for example here [1]; I was simply following the API I was seeing there. I have tested the way you suggest and it works fine. It's a bit weird using a different api for bearer and basic, but if this is the sanctioned way, I'm fine, we can close this.


1. https://github.com/nextcloud/server/blob/master/lib/private/Files/Storage/DAV.php#L482

PR #3313, opened 2025-12-25
Fix CURLOPT_HTTPAUTH defined() checks on cross-origin redirects

zigzagdev:
This PR fixes an incorrect use of defined() when checking for curl authentication constants during cross-origin redirects.

Previously, the code was calling `defined('\CURLOPT_HTTPAUTH')`, which treats the backslash-prefixed string as a different constant name. As a result, the condition could evaluate to false even when the curl extension was available, preventing the intended removal of curl authentication options on cross-origin redirects.

This change ensures that curl auth options are correctly removed when a redirect crosses origin boundaries, restoring the intended and documented behaviour.

- What was changed
    - Replaced defined('\CURLOPT_HTTPAUTH') with defined('CURLOPT_HTTPAUTH')
    - Ensured related logic is evaluated correctly when curl is available
    - Updated existing tests to reflect the corrected constant check and avoid false negatives

- Tests

    - Updated redirect middleware tests to use the correct defined('CURLOPT_HTTPAUTH') guard
    - All existing redirect behaviour tests remain unchanged
    - No new test cases were required, as existing coverage already exercised the affected code paths

- test results
スクリーンショット 2025-12-25 22 33 56

### DIFF
--- a/src/RedirectMiddleware.php
+++ b/src/RedirectMiddleware.php
@@ -89,7 +89,10 @@ class RedirectMiddleware
         $nextRequest = $this->modifyRequest($request, $options, $response);
 
         // If authorization is handled by curl, unset it if URI is cross-origin.
-        if (Psr7\UriComparator::isCrossOrigin($request->getUri(), $nextRequest->getUri()) && defined('\CURLOPT_HTTPAUTH')) {
+        if (
+            Psr7\UriComparator::isCrossOrigin($request->getUri(), $nextRequest->getUri())
+            && defined('CURLOPT_HTTPAUTH')
+        ) {
             unset(
                 $options['curl'][\CURLOPT_HTTPAUTH],
                 $options['curl'][\CURLOPT_USERPWD]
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
Thanks for the PR, but I don't think there's a bug here.

The premise — that `defined('\CURLOPT_HTTPAUTH')` "treats the backslash-prefixed string as a different constant name" — isn't how `defined()` works. A leading `\` denotes a fully-qualified name in the global namespace, which is exactly where `CURLOPT_HTTPAUTH` lives. Both forms are equivalent, even from inside the `GuzzleHttp` namespace:

```php
namespace GuzzleHttp;

var_dump(defined('\CURLOPT_HTTPAUTH'));           // bool(true)
var_dump(defined('CURLOPT_HTTPAUTH'));            // bool(true)
var_dump(defined('GuzzleHttp\CURLOPT_HTTPAUTH')); // bool(false)
```

The existing tests also confirm this: they `markTestSkipped` only when `defined('\CURLOPT_HTTPAUTH')` returns false, and they've been passing on CI with ext-curl loaded since #2845 was merged in 2022. If the guard were broken, the CVE-2022-31090 fix wouldn't have actually worked for ~4 years — but it has.

Both the source and test changes are no-ops, and there's no security impact. Closing this one.

PR #3316, opened 2026-02-08
Allow omitting Domain attribute in SetCookie per RFC 6265

### LINKED ISSUE #3315: SetCookie does not support omitting Domain attribute
**Guzzle version(s) affected**: 7.10.0 (but also earlier)
**PHP version**: any
**cURL version**: any

**Description**  
When storing a Set-Cookie value in the CookieJar, I would like to store a cookie without a Domain attribute, so that when this information is later reused to render a set-cookie header on the response to the browser, it is interpreted as a "host cookie".

This is correct behavior per RFC 6265 which the SetCookie class claims to follow according to the comment above the validate method.

This is not the same as setting the Domain to that specific host, since that will allow reading on subdomains as well.
We are using Guzzle to query a GraphQL server, which can return set-cookie headers, which we then store in the Guzzle CookieJar and later forward with the actual response to the browser. This GraphQL server may also be queried directly from the browser. Since the GraphQL server does not specify a Domain attribute, the cookie becomes a host cookie. But since Guzzle requires us to set a Domain we get conflicting versions of the same cookie (seen in devtools as a domain with a leading period and the host cookie without a leading period).

**How to reproduce**  
```php
<?php

use GuzzleHttp\Cookie\CookieJar;
use GuzzleHttp\Cookie\SetCookie;

$cookieJar = new CookieJar($strictMode = true);
$setCookie = new SetCookie([
    'Name' => 'test',
    'Value' => 'test',
]);

$cookieJar->setCookie($setCookie); // RuntimeException: Invalid cookie: The cookie domain must not be empty
```

**Possible Solution**  
Allow an unset domain in `SetCookie::validate()`, but I don't know if this will lead to issues elsewhere if the domain is assumed to always be present.

**Additional context**  

RFC 6265 section 5.2.3 specifies how to parse the value of the Domain attribute, I think Guzzle implements this correctly in the validate method, as far as I can tell:
https://www.rfc-editor.org/rfc/rfc6265.html#section-5.2.3

But section 4.1.2.3 specifies how a missing Domain attribute should be handled:
https://www.rfc-editor.org/rfc/rfc6265.html#section-4.1.2.3

The relevant part is:
> If the server omits the Domain attribute, the user agent will return the cookie only to the origin server.

This does not seem to be supported by Guzzle right now, making it impossible to store this in the CookieJar.

QDenka:
## Summary

Per [RFC 6265 Section 4.1.2.3](https://www.rfc-editor.org/rfc/rfc6265.html#section-4.1.2.3), when the Domain attribute is omitted from a Set-Cookie header, the cookie becomes a **host-only** cookie returned only to the origin server. This is distinct from explicitly setting the Domain to the origin, which also allows subdomains.

## Problem

Currently, `SetCookie::validate()` rejects cookies with a `null` Domain, throwing:

> RuntimeException: Invalid cookie: The cookie domain must not be empty

This makes it impossible to store host-only cookies in the `CookieJar` when creating cookies manually (without going through `extractCookies()`).

## Solution

Allow `null` Domain in `validate()` while still rejecting empty string (`''`) domains. The `matchesDomain()` method already handles `null` domains correctly (returns `true`).

## Changes

- **`src/Cookie/SetCookie.php`**: Updated `validate()` to allow `null` domain (host-only cookie)
- **`tests/Cookie/SetCookieTest.php`**: Added test case for null domain validation
- **`tests/Cookie/CookieJarTest.php`**: Added test verifying CookieJar accepts cookies without Domain

Fixes #3315

### DIFF
--- a/src/Cookie/SetCookie.php
+++ b/src/Cookie/SetCookie.php
@@ -482,8 +482,11 @@ class SetCookie
 
         // Domains must not be empty, but can be 0. "0" is not a valid internet
         // domain, but may be used as server name in a private network.
+        // A null domain is allowed per RFC 6265 Section 4.1.2.3 — when the
+        // Domain attribute is omitted, the cookie becomes a "host-only" cookie
+        // that is returned only to the origin server.
         $domain = $this->getDomain();
-        if ($domain === null || $domain === '') {
+        if ($domain !== null && $domain === '') {
             return 'The cookie domain must not be empty';
         }
 
--- a/tests/Cookie/CookieJarTest.php
+++ b/tests/Cookie/CookieJarTest.php
@@ -500,6 +500,21 @@ class CookieJarTest extends TestCase
         self::assertCount($matches ? 1 : 0, $this->jar->toArray());
     }
 
+        /**
+     * @see https://github.com/guzzle/guzzle/issues/3315
+     */
+    public function testAcceptsCookieWithoutDomain(): void
+    {
+        $jar = new CookieJar(true);
+        $cookie = new SetCookie([
+            'Name' => 'test',
+            'Value' => 'value',
+        ]);
+        self::assertTrue($jar->setCookie($cookie));
+        self::assertCount(1, $jar);
+        self::assertNull($jar->toArray()[0]['Domain']);
+    }
+
     private function futureExpirationDate()
     {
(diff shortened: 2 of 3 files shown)


GrahamCampbell:
Thanks for the PR. The issue is valid, but this change would make domainless cookies behave like wildcard request cookies in the current CookieJar. I opened #3364 as the safe 7.11 replacement, so closing this one.

PR #3317, opened 2026-02-15
Add Psalm/IDE type docs for Client::__construct() config (fixes #3194)

### LINKED ISSUE #3194: Document `GuzzleHttp\Client::__construct()` parameters at type level?
**Description**

I just happened to stumble upon https://github.com/guzzle/guzzle/blob/41042bc7ab002487b876a0683fc8dce04ddce104/src/Client.php#L26-L57 while configuring the `Client`.

One thing I noticed here is that the docblock mentions:

```
@see \GuzzleHttp\RequestOptions for a list of available request options.
```

I'm wondering if we could add type-level documentation for this constructor for PHPStan, Psalm and IntelliJ, such as:

```php
/**
 * @psalm-import OptionsArray from \GuzzleHttp\RequestOptions
 */
class Client // snip
{
// <snip>
    /**
     * <snip>
     *
     * @psalm-param OptionsArray $config
     */
    public function __construct(array $config = [])
```

Then, in `RequestOptions`:

```php
/**
 * <snip>
 *
 * @psalm-type OptionsArray = array{
 *    timeout?: int<0, max>,
 *    more?: stuff-here,
 *    ...
 * }
 */
final class RequestOptions
{
// <snip>
```

Note how the 3 dots at the end are necessary to allow options to expand over time.

This kind of addition would be a massive ergonomic improvement, IMO.

Similar to https://github.com/getsentry/sentry-php/issues/1656

randyyates:
- Add @psalm-type OptionsArray in RequestOptions with full option shape
- Add @psalm-import-type and @psalm-param OptionsArray in Client::__construct()
- Align option types with official docs (decode_content, delay, force_ip_resolve)
- Update PHPStan baseline for new typing (uriFor count, booleanNot.alwaysFalse)

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -105,7 +105,13 @@ parameters:
 		-
 			message: '#^Parameter \#1 \$uri of static method GuzzleHttp\\Psr7\\Utils\:\:uriFor\(\) expects Psr\\Http\\Message\\UriInterface\|string, mixed given\.$#'
 			identifier: argument.type
-			count: 2
+			count: 1
+			path: src/Client.php
+
+		-
+			message: '#^Negated boolean expression is always false\.$#'
+			identifier: booleanNot.alwaysFalse
+			count: 1
 			path: src/Client.php
 
 		-
--- a/src/Client.php
+++ b/src/Client.php
@@ -13,6 +13,8 @@ use Psr\Http\Message\UriInterface;
 
 /**
  * @final
+ *
+ * @psalm-import-type OptionsArray from \GuzzleHttp\RequestOptions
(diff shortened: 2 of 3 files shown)


theodorejb:
Couldn't the detailed type be directly in `@param array $config` (replacing `array` there) rather than a separate `@psalm-param` annotation?

GrahamCampbell:
Thanks for the PR. I'm going to close this, as I am working on a more comprehensive round of PRs to improve typing across the entire first-party ecosystem in the next major version series including adding generic typing to promises and array-shape typing where practical instead of the shitty broken param-var syntax.

PR #3318, opened 2026-02-24
[Bugfix] Use Host header if it is present in the default config

1ma:
When a Client object is created, if the `Host` header is passed as a default header it is ignored.

This is because default headers are kept apart, and only set at a later stage in `applyOptions`. However when the Psr7\Request is created without any headers a default Host based on the `base_uri` might be inserted as default, therefore the desired default value won't be set later in `applyOptions` because a Host header already exists in the request object.

This behavior makes `Host` a special case. Therefore if it is present as a default header but not present in the request options, we force the desired header value.

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -181,6 +181,12 @@ parameters:
 			path: src/Client.php
 
 		-
+			message: '#^Cannot access offset ''Host'' on mixed\.$#'
+			identifier: offsetAccess.nonOffsetAccessible
+			count: 3
+			path: src/Client.php
+
+		-
 			message: '#^Parameter \#1 \$data of class GuzzleHttp\\Cookie\\SetCookie constructor expects array, mixed given\.$#'
 			identifier: argument.type
 			count: 1
--- a/src/Client.php
+++ b/src/Client.php
@@ -162,6 +162,10 @@ class Client implements ClientInterface, \Psr\Http\Client\ClientInterface
         if (\is_array($body)) {
             throw $this->invalidBody();
         }
+        // Special case for the Host header
+        if (!isset($headers['Host']) && isset($this->config['headers']['Host'])) {
+            $headers['Host'] = $this->config['headers']['Host'];
+        }
         $request = new Psr7\Request($method, $uri, $headers, $body, $version);
         // Remove the option so that they are not doubly-applied.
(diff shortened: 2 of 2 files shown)


1ma:
The following is a description of the scenario where the current behavior is problematic.

A Docker stack contains an API gateway that sits in front of several upstream servers and manages authentication and authorization for all them:

```yaml
  kong:
    image: kong:3.2.2
    ports:
      - "8080:80"
    networks:
      default:
        aliases:
          - api1.example.com
          - api2.example.com
          - api3.example.com
```

A functional test suite based on PHPUnit that runs outside the stack has a Guzzle client that is reused among all tests to increase efficiency (by reusing the TCP connection), but it must pass the Host header so that the API Gateway knows where to route the request. 90% of tests exercise api1, so the client is configured by default to work with it:

```php
    self::$http = new Client([
        'base_uri' => '127.0.0.1:8080',
        'headers' => [
            'Authorization' => 'Basic xxxxxxx=',
            'Content-Type' => 'application/xml',
            'Host' => 'api1.example.com'
        ]
    ]);
```

The minority of tests that don't hit api1 redefine headers at request time like so:

```php
    self::$http->post('/some/endpoint', [
        'body' => '...',
        'headers' => [
            'Authorization' => 'Basic yyyyyyyy=',
            'Host' => 'api2.example.com'
        ]
    ]);
```

The latter redefinition works, however by default Guzzle 7.10.0 sends requests with header `Host: 127.0.0.1:8080` instead of `api1.example.com`, which the API gateway does not know how to route. This can be easily observed with Guzzle's `debug => true` option, which turns on cURL's verbose mode.

GrahamCampbell (changes requested):
Thanks for the PR, but the host header is not intended to be passed from the outside. The intended way is the base URI config. This PR also has a technical issue that header names need to be handled in the case-insensitive way.

1ma:
@GrahamCampbell In 99% of contexts it's not necessary, that's why I outlined the situation where this becomes a problem. In some testing contexts it makes sense to have this kind of control over the Host header.

Without this fix you now need to edit `/etc/hosts` or equivalent in the host machine to redirect requests to the appropriate IP, whereas with this fix it's not necessary.

```
127.0.0.1     api1.example.com api2.example.com api3.example.com
```

I'd rather be given the opportunity to fix the case-insensitive issue in the code.

PR #3326, opened 2026-04-16
Harden FileCookieJar against PHP object injection gadget

ohader:
FileCookieJar has a `__destruct()` method writing to `$filename`, making it a potential gadget for PHP object injection attacks (see `phpggc Guzzle/FW1`). This change adds a `__wakeup()` method that disables auto-saving to `$filename` on destruction.

Also encode `<`, `'`, and `"` in JSON cookie file output via `JSON_HEX_TAG | JSON_HEX_APOS | JSON_HEX_QUOT` to prevent injected cookie values from being executed as PHP when written to a `.php` file path.

The gadget chain is publicly documented in phpggc as `Guzzle/FW1` and has been known since at least 2023. This change is therefore handled publicly to maximize visibility and mitigate security impact for consuming applications.

---

> [!NOTE]
> The potential security implications of this pull request have been disclosed to the Tidelift/SonarSource security team at security@tidelift.com. After reviewing the report, they advised that the issue does not require a private disclosure process and that opening a regular pull request is the appropriate course of action.

### DIFF
--- a/src/Cookie/FileCookieJar.php
+++ b/src/Cookie/FileCookieJar.php
@@ -20,6 +20,11 @@ class FileCookieJar extends CookieJar
     private $storeSessionCookies;
 
     /**
+     * @var bool Whether to invoke save() automatically on shutdown.
+     */
+    private $autoSave = true;
+
+    /**
      * Create a new FileCookieJar object
      *
      * @param string $cookieFile          File to store the cookie data
@@ -44,7 +49,22 @@ class FileCookieJar extends CookieJar
      */
     public function __destruct()
     {
-        $this->save($this->filename);
+        if ($this->autoSave) {
+            $this->save($this->filename);
+        }
+    }
+
+    /**
+     * Disables auto-save on destruction.
+     *
+     * This class has a __destruct() method that writes to $filename,
+     * making it a potential gadget for PHP object injection attacks.
+     *
+     * @throws \RuntimeException
+     */
+    public function __wakeup(): void
+    {
+        $this->autoSave = false;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


mcdruid:
I filed https://github.com/guzzle/guzzle/pull/3286 to look at this - there was also an accompanying issue, but that seems to have been deleted.

Throwing an exception in `__wakeup()` is great so long as that doesn't interfere with the intended functionality; I'd assumed that it would, but perhaps that's incorrect.

ohader:
> I filed #3286 to look at this - there was also an accompanying issue, but that seems to have been deleted.
> 
> Throwing an exception in `__wakeup()` is great so long as that doesn't interfere with the intended functionality; I'd assumed that it would, but perhaps that's incorrect.

Thanks for your feedback. I was aware of the breaking behavior by throwing an exception during deserialization.
I've adjusted the change a bit, to at least deny auto-saving the file when a previously deserialized instance is destructed.

GrahamCampbell:
Your issue may have been auto-deleted by a bot that I have that deletes things that look like security reports. I agree this is not a security issue, but could be considered security hardening to reduce the chance of security issues introduced by misuse.

GrahamCampbell:
I'm considering a different fix, and targetting only the next major version only: https://github.com/guzzle/guzzle/pull/3334.

ohader:
Closed in favor of → PR #3334

mcdruid:
> Your issue may have been auto-deleted by a bot that I have that deletes things that look like security reports. I agree this is not a security issue, but could be considered security hardening to reduce the chance of security issues introduced by misuse.

Thanks for the explanation, and apologies if I misinterpreted the closure as intentional.

PR #3327, opened 2026-04-22
Add support for curl_share_init and curl_share_init_persistent

### LINKED ISSUE #3307: Improve CURL performance on PHP8.5 with curl_share_init_persistent
**Description:**
PHP 8.5 introduces the `curl_share_init_persistent` function, which allows for the creation of named, persistent shared cURL handles.

Integrating this into Guzzle’s `CurlHandler` would allow clients to share **DNS cache, SSL session data, and cookies** across requests more efficiently. This results in significant speed improvements by avoiding repeated DNS lookups and SSL handshakes.

**Reference & Examples:**
A detailed breakdown of the function and its performance benefits can be found here:
https://zetcode.com/php-curl/curl_share_init_persistent/

**Impact:**
This would provide a massive benefit when using Guzzle within large ecosystem libraries such as **Google Cloud**. These SDKs often instantiate multiple clients or make frequent requests where the overhead of repeated SSL handshakes and DNS lookups creates significant latency. Native support for persistent sharing would drastically optimize these workflows.

**Proposed Solution:**
Update the Guzzle cURL handler to accept a persistent share name (or configuration option) that utilizes `curl_share_init_persistent` when running on PHP 8.5+, allowing the underlying handle to reuse connection data automatically.

samanthadrago:
**Motivation:**

Closes #3307 

PHP 8.5 introduces the [curl_share_init_persistent function](https://www.php.net/manual/en/function.curl-share-init-persistent.php), which allows for the creation of named, persistent shared cURL handles. Supporting this in Guzzle would allow callers to take advantage of the performance benefits that come with using a persistent shared curl handle by reducing the overhead of repeated SSL handshakes and DNS lookups. 

**Changes introduced:**
This PR adds the ability to instantiate a CurlHandler or CurlMultiHandler with some added `$options`:
- `share` - an array of CURL_LOCK_DATA_* constants to share between requests via [`curl_share_init()`](https://www.php.net/manual/en/function.curl-share-init.php)
- `share_persistent` - an array of CURL_LOCK_DATA_* constants to share between requests via [`curl_share_init_persistent()`](https://www.php.net/manual/en/function.curl-share-init-persistent.php)

If the `share_persistent` option is specified and the function `curl_share_init_persistent` exists (it was added in PHP 8.5.0), we will create a persistent share handle. If only `share` is present and `curl_share_init` exists, we will create a regular share handle.

Within the `__destruct` method for both CurlHandler and CurlMultiHandler, we will call [`curl_share_close()`](https://www.php.net/manual/en/function.curl-share-close.php) if the share handle is not persistent and the PHP version is < 8.0.0 (at which point the function is a noop, according to the [docs](https://www.php.net/manual/en/function.curl-share-close.php)).

**Example**

I ran a PHP server with the following script (`php -S localhost:8000 examples/server_persistent.php`):

```php
<?php

use GuzzleHttp\Client;
use GuzzleHttp\HandlerStack;
use GuzzleHttp\Handler\CurlMultiHandler;
use GuzzleHttp\TransferStats;

$handler = new CurlMultiHandler([
    'share_persistent' => [
        CURL_LOCK_DATA_DNS,
        CURL_LOCK_DATA_CONNECT,
        CURL_LOCK_DATA_SSL_SESSION,
    ]
]);

$client = new Client(['handler' => HandlerStack::create($handler)]);
$stats = null;

$response = $client->get('https://example.com/', [
    'on_stats' => function (TransferStats $transferStats) use (&$stats) {
        $curlStats = $transferStats->getHandlerStats();
        $stats = $curlStats;
    }
]);

header('Content-Type: text/plain');
echo sprintf("namelookup_time_us:     %6d\n", ($stats['namelookup_time'] ?? 0) * 1_000_000);
echo sprintf("connect_time_us:        %6d\n", ($stats['connect_time'] ?? 0) * 1_000_000);
echo sprintf("appconnect_time_us:     %6d\n", ($stats['appconnect_time'] ?? 0) * 1_000_000);
echo sprintf("starttransfer_time_us:  %6d\n", ($stats['starttransfer_time'] ?? 0) * 1_000_000);
echo sprintf("total_time_us:          %6d\n", ($stats['total_time'] ?? 0) * 1_000_000);
echo "\n";
```

I then ran `curl localhost:8000` twice to confirm that we are sharing the handle - the second request had negligible namelookup_time, appconnect_time, and connect_time values:
```
➜  guzzle git:(support_curl_share_init_persistent) ✗ curl localhost:8000
namelookup_time_us:      19336
connect_time_us:         31954
appconnect_time_us:     108133
starttransfer_time_us:  126151
total_time_us:          126911

➜  guzzle git:(support_curl_share_init_persistent) ✗ curl localhost:8000
namelookup_time_us:         39
connect_time_us:            39
appconnect_time_us:         39
starttransfer_time_us:   17842
total_time_us:           18326

```

### DIFF
--- a/docs/handlers-and-middleware.rst
+++ b/docs/handlers-and-middleware.rst
@@ -302,3 +302,48 @@ These request options are a subset of request options called
 - :ref:`ssl_key-option`
 - :ref:`stream-option`
 - :ref:`verify-option`
+
+
+Handler Options
+===============
+
+CurlHandler and CurlMultiHandler
+---------------------------------
+
+The ``CurlHandler`` and ``CurlMultiHandler`` accept constructor options that
+configure how cURL handles are created and managed.
+
+share
+~~~~~
+
+:Summary: Array of ``CURL_LOCK_DATA_*`` constants to share between requests via ``curl_share_init()``
+:Types: ``int[]``
+:Default: ``null``
+
+Specify which data types should be shared between cURL handles. Creates a
+cURL share handle that allows multiple cURL handles to share the specified
+data. This can significantly improve performance for applications making
+multiple requests to the same hosts by eliminating redundant DNS lookups
+and SSL handshakes.
+
+share_persistent
+~~~~~~~~~~~~~~~~
+
+:Summary: Array of ``CURL_LOCK_DATA_*`` constants for persistent sharing via ``curl_share_init_persistent()`` (PHP 8.5+)
+:Types: ``int[]``
+:Default: ``null``
+
+In PHP 8.5+, you can specify which data types to share across requests using
+persistent cURL share handles. Persistent share handles survive across multiple
+PHP requests in long-running processes or FPM workers, providing even better
+performance than regular share handles.
+
+If both ``share`` and ``share_persistent`` are specified, ``share_persistent``
+will be used if ``curl_share_init_persistent()`` is available (PHP 8.5+),
+otherwise ``share`` will be used as a fallback.
+
+.. important::
+
+    Do not include ``CURL_LOCK_DATA_COOKIE`` in share handles, as PHP will
+    raise a ``ValueError`` when using persistent share handles. Guzzle manages
+    cookies through middleware, not at the cURL level.
--- a/src/Handler/CurlFactory.php
+++ b/src/Handler/CurlFactory.php
@@ -38,11 +38,18 @@ class CurlFactory implements CurlFactoryInterface
     private $maxHandles;
 
     /**
-     * @param int $maxHandles Maximum number of idle handles.
+     * @var \CurlShareHandle|\CurlSharePersistentHandle|resource|null
      */
-    public function __construct(int $maxHandles)
+    private $shareHandle;
+
+    /**
+     * @param int                                                       $maxHandles  Maximum number of idle handles.
+     * @param \CurlShareHandle|\CurlSharePersistentHandle|resource|null $shareHandle Optional cURL share handle for sharing data between handles.
+     */
+    public function __construct(int $maxHandles, $shareHandle = null)
     {
         $this->maxHandles = $maxHandles;
+        $this->shareHandle = $shareHandle;
     }
 
     public function create(RequestInterface $request, array $options): EasyHandle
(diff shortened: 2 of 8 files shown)


GrahamCampbell:
Thanks for getting started on this. I’m planning to look at this some more this weekend (time permitting!).

samanthadrago:
@GrahamCampbell let me know if there's any more info I can provide/testing I can help with on this. We're eager to get this included to help unblock a large migration we're doing in our codebase (away from curl_exec toward more widespread use of Guzzle). 

Thank you! 🙏

GrahamCampbell:
Thanks for working on this. The core idea looks useful, and the `CURLOPT_SHARE` approach is the right direction. After digging into the cURL handler lifecycle work, though, I do not think we can merge this PR as-is, especially not into `7.10`. The main issue is that this adds new handler constructor options and new native resource ownership semantics. That makes it an 8.x feature, not a 7.x patch-level change.

There are also a few design issues we need to resolve before merging:
- Destructor cleanup is currently being changed in an area we need to make strictly non-throwing. Share-handle cleanup should be integrated with the explicit 8.x `close()` lifecycle, with destructors remaining best-effort and quiet.
- `handle_factory` combined with `share` or `share_persistent` currently creates a share handle that a custom factory will not use. That should either throw or require a share-aware factory API.
- `share_persistent` should not silently do nothing when `curl_share_init_persistent()` is unavailable unless `share` is explicitly provided as a fallback.
- `CURL_LOCK_DATA_COOKIE` should be rejected, not just documented as unsafe, since Guzzle manages cookies through middleware.
- `curl_share_setopt()` and `CURLOPT_SHARE` application failures should be checked and surfaced clearly.
- `curl_share_close()` is deprecated in PHP 8.5 and a no-op on PHP 8+, so cleanup needs to avoid calling it there.
- The tests should verify behavior and lifecycle ownership, not only that requests complete without throwing.

I’m going to take over this work and rework it for the 8.x branch after the cURL cleanup/lifecycle changes land. The feature should still be viable there, with handlers owning created share handles, `CurlFactory` only borrowing the handle to apply `CURLOPT_SHARE`, persistent handles never being closed by Guzzle, and explicit `close()` handling non-persistent cleanup. Thanks again for getting this started.

GrahamCampbell:
Support for handler-level sharing will be available in 7.11 (hard-coded to DNS and SSL session). Support for persistent sharing will be available in 8.0 (hard-coded to DNS, connection, and SSL session), with two modes: prefer (that falls back to 7.11-style handler-level sharing) and require (that ensures the full persistent sharing).

GrahamCampbell:
For your reference, the two relevant PRs are:

* https://github.com/guzzle/guzzle/pull/3449
* https://github.com/guzzle/guzzle/pull/3444

PR #3329, opened 2026-05-10
Add tests for ClientException, ServerException, and TooManyRedirectsException

mdalikadar:


GrahamCampbell:
Thanks for the PR, but we're not interested in reviewing AI generated tests that don't add value. To be very candid - if I wanted to have AI add more tests, I could run that prompt myself. ;)

PR #3332, opened 2026-05-15
Fix relative URI parsing when a path segment contains a colon

### LINKED ISSUE #3306: Relative URI parsed incorrectly by PHP parse_url()
**Guzzle version(s) affected**: 7.10.0
**PHP version**: 8.3.20
**cURL version**: 7.81.0

**Description**  
Some relative in combination with base_uri are parsed incorrectly by PHP parse_url() resulting in invalid URLs.

**How to reproduce**  
```php
$client = new Client(['base_uri' => 'https://bedrock-runtime.us-east-1.amazonaws.com']);
$client->post('model/amazon.titan-image-generator-v2:0/invoke', []);
```

**Possible Solution**  
Guzzle parses the relative path here: https://github.com/guzzle/guzzle/blob/7.10/src/Client.php#L161
parse_url() (https://github.com/guzzle/psr7/blob/2.8/src/Uri.php#L125) interprets `model/amazon.titan-image-generator-v2:0/invoke` as
```
host: model
port: 0
path: /amazon.titan-image-generator-v2:0/invoke
```

A possible solution would be to concatenate the base_uri and uri values instead of parsing them separately.

masskrdjn:
## Summary

- A relative URI with a colon in a non-initial path segment (e.g. `model/amazon.titan-image-generator-v2:0/invoke`) was mis-parsed by `parse_url()` as `host=model`, `port=0`, yielding an invalid request URI once merged with `base_uri`.
- For relative-path references (RFC 3986 §4.2) combined with a `base_uri`, we now bypass `parse_url()` and feed the path verbatim into `Uri::withPath()` before letting `UriResolver::resolve()` do the merging. Query and fragment are split off manually first.
- Absolute URIs, absolute-path references (`/foo`), network-path references (`//host/foo`) and references whose first segment contains a colon (treated as scheme per RFC 3986) keep their previous behavior.

## Reproduction (before the fix)

```php
$client = new Client(['base_uri' => 'https://bedrock-runtime.us-east-1.amazonaws.com']);
$client->post('model/amazon.titan-image-generator-v2:0/invoke');
// → request URI was malformed (host=model, port=0)
```

Fixes #3306 

Test plan
 New unit test ClientTest::testCanMergeRelativeUriWithColonInPathSegmentOnBaseUri covering the reported case.
 Full ClientTest suite passes (63 tests).
 Manual edge-case sweep: absolute URI override, absolute path, query-only, fragment-only, empty relative, scheme-prefixed (foo:bar/baz), query + fragment on the relative part.

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -158,7 +158,7 @@ class Client implements ClientInterface, \Psr\Http\Client\ClientInterface
         $body = $options['body'] ?? null;
         $version = $options['version'] ?? '1.1';
         // Merge the URI into the base URI.
-        $uri = $this->buildUri(Psr7\Utils::uriFor($uri), $options);
+        $uri = $this->buildUri(self::uriForRequest($uri, $options), $options);
         if (\is_array($body)) {
             throw $this->invalidBody();
         }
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
Thanks for the PR, although I don't agree that adding in this code here is the correct approach. I've prepared an alternative PR at https://github.com/guzzle/psr7/pull/645.

masskrdjn:
@GrahamCampbell you're right, my patch should have been at another level. Thanks for the feedback

PR #3333, opened 2026-05-16
chore: Update github action 'actions/setup-node' to version 6

Gared:
This PR updates actions/setup-node to the latest version (6)


GrahamCampbell:
Thanks for the PR, however these changes are incomplete (we also need to move to a non-EOL version of node.js). I've prepared a more complete version at https://github.com/guzzle/guzzle/pull/3336.

PR #3334, opened 2026-05-17
Harden FileCookieJar against unserialize file writes

GrahamCampbell:
Hardened `FileCookieJar` against use as a PHP object injection file-write gadget. Instances restored via `unserialize()` no longer automatically save on destruction, and saved cookie files now JSON-escape tag characters so cookie values cannot write literal PHP open tags to disk.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,16 @@
 
 Please refer to [UPGRADING](UPGRADING.md) guide for upgrading to a major version.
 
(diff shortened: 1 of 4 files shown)


ohader (approved):
This is pretty similar to what I suggested in PR #3326. I'm glad to see this moving forward. Including the change only in the next major version is totally fine to avoid potential side-effects.

mcdruid:
This looks like a neat protection / security hardening. Thanks!

PR #3384, opened 2026-05-20
Cast Content-Length to string in PrepareBodyMiddleware

### LINKED ISSUE #3383: PrepareBodyMiddleware sets Content-Length as int, violating PSR-7 (string required)
## Summary
`GuzzleHttp\PrepareBodyMiddleware` writes `Content-Length` as an `int` value.
PSR-7 `MessageInterface::withHeader()` requires `string|string[]`. Strict
PSR-7 implementations (e.g. `TYPO3\CMS\Core\Http\Message`) throw
`InvalidArgumentException`.

## PSR-7 spec
From [`php-fig/http-message` MessageInterface.php#L131-L145](https://github.com/php-fig/http-message/blob/641e9f812d34d66373fa45ce7a1edf7e123d969a/src/MessageInterface.php#L131-L145):

```php
/**
 * Return an instance with the provided value replacing the specified header.
 *
 * While header names are case-insensitive, the casing of the header will
 * be preserved by this function, and returned from getHeaders().
 *
 * This method MUST be implemented in such a way as to retain the
 * immutability of the message, and MUST return an instance that has the
 * new and/or updated header and value.
 *
 * @param string $name Case-insensitive header field name.
 * @param string|string[] $value Header value(s).
 * @return static
 * @throws \InvalidArgumentException for invalid header names or values.
 */
public function withHeader(string $name, $value): MessageInterface;
```

## Reproduction
`vendor/guzzlehttp/guzzle/src/PrepareBodyMiddleware.php`:

```php
$size = $request->getBody()->getSize(); // int
$modify['set_headers']['Content-Length'] = $size; // int passed downstream
```

Until `guzzlehttp/psr7` 2.9.0, `Utils::modifyRequest()` rebuilt the request as
a Guzzle `Request` whose lenient `Psr7\Message` accepted the int — masking the
bug. Since psr7 2.10.0 ([#655](https://github.com/guzzle/psr7/pull/655)),
`modifyRequest()` mutates the original `RequestInterface` via `withHeader()`,
so any spec-compliant PSR-7 implementation now rejects the int.

Stack trace from a TYPO3 14 / Solarium 6.4.1 setup:

```
InvalidArgumentException: Invalid header value for header "Content-Length".
The value must be a string or an array of strings.

vendor/typo3/cms-core/Classes/Http/Message.php:209
vendor/guzzlehttp/psr7/src/Utils.php:242   (withHeader)
vendor/guzzlehttp/guzzle/src/PrepareBodyMiddleware.php:64
vendor/guzzlehttp/guzzle/src/Middleware.php:31
vendor/guzzlehttp/guzzle/src/RedirectMiddleware.php:55
vendor/guzzlehttp/guzzle/src/Middleware.php:63
vendor/guzzlehttp/guzzle/src/HandlerStack.php:75
vendor/guzzlehttp/guzzle/src/Client.php:333
vendor/solarium/solarium/src/Core/Client/Adapter/Psr18Adapter.php:66
```

## Suggested fix
Cast in `PrepareBodyMiddleware`:
```php
$modify['set_headers']['Content-Length'] = (string) $size;
```

## Affected downstream
- TYPO3-Solr/ext-solr: TYPO3-Solr/ext-solr#4660

## Versions
- guzzlehttp/guzzle 7.10.1
- guzzlehttp/psr7 2.10.0
- PHP 8.2.30

dkd-kaehm:
Fixes #3383.

`PrepareBodyMiddleware` was setting `Content-Length` as an `int` value taken
directly from `Stream::getSize()`. PSR-7 [`MessageInterface::withHeader()`](https://github.com/php-fig/http-message/blob/641e9f812d34d66373fa45ce7a1edf7e123d969a/src/MessageInterface.php#L131-L145)
requires `string|string[]`, so strict PSR-7 implementations throw
`InvalidArgumentException` as soon as the request gets there.

This was previously masked by `guzzlehttp/psr7` < 2.10.0, where
`Utils::modifyRequest()` rebuilt a fresh Guzzle `Request` whose lenient
`Psr7\Message` accepted the int. Since psr7 2.10.0 ([guzzle/psr7#655](https://github.com/guzzle/psr7/pull/655))
the original `RequestInterface` is mutated via `withHeader()` directly, so the
type violation reaches strict implementations untouched.

## Change
- `src/PrepareBodyMiddleware.php`: cast `$size` to string before assigning to
  `Content-Length`.
- `tests/PrepareBodyMiddlewareTest.php`: regression test using an anonymous
  class extending `Request` with a spec-compliant `withHeader()` (mirrors
  `TYPO3\CMS\Core\Http\Message`'s strictness).

## Verified
- Without the cast, the new test errors with the exact `InvalidArgumentException`
  reported downstream (see #3383).
- With the cast, all `PrepareBodyMiddleware` tests pass (14 / 14).

## Downstream affected
- TYPO3-Solr/ext-solr#4660 (TYPO3 14 + Solarium 6.4.1)

### DIFF
--- a/src/PrepareBodyMiddleware.php
+++ b/src/PrepareBodyMiddleware.php
@@ -52,7 +52,7 @@ class PrepareBodyMiddleware
         ) {
             $size = $request->getBody()->getSize();
             if ($size !== null) {
-                $modify['set_headers']['Content-Length'] = $size;
+                $modify['set_headers']['Content-Length'] = (string) $size;
             } else {
                 $modify['set_headers']['Transfer-Encoding'] = 'chunked';
             }
--- a/tests/PrepareBodyMiddlewareTest.php
+++ b/tests/PrepareBodyMiddlewareTest.php
@@ -11,6 +11,7 @@ use GuzzleHttp\Psr7\FnStream;
 use GuzzleHttp\Psr7\Request;
 use GuzzleHttp\Psr7\Response;
 use PHPUnit\Framework\TestCase;
+use Psr\Http\Message\MessageInterface;
 use Psr\Http\Message\RequestInterface;
 
 class PrepareBodyMiddlewareTest extends TestCase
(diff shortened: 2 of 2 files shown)


GrahamCampbell:
Thanks. I've replaced this with a wider reaching PR to fix a few other similar issues and invalid tests.

PR #3426, opened 2026-05-22
Add PSR-17 request and URI factories

GrahamCampbell:
This change adds PSR-17 request and URI factory support for requests and URIs that Guzzle creates internally. Guzzle 8 already targets PSR-7 3.x, where `GuzzleHttp\Psr7\HttpFactory` provides the standard PSR-17 factory interfaces, so exposing these factories as normal client and request options gives applications with custom PSR-7 implementations a supported way to keep client-created request and URI objects in their own implementations.

The new `request_factory` option is used when `request()`, `requestAsync()`, and the shortcut methods create a request. The new `uri_factory` option is used when string request URI values and string `base_uri` values are converted into URI objects. Both options can be configured on the client or overridden per request, and requests passed directly to `send()`, `sendAsync()`, or `sendRequest()` continue to be used as provided.

This is intentionally limited to the first request-creation path. Redirect `Location` parsing, request body stream creation, multipart internals, response creation, and handler-level streams are left unchanged so they can be reviewed separately in follow-up work.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ Please refer to [UPGRADING](UPGRADING.md) guide for upgrading to a major version
 - Add generic PHPDoc annotations to async HTTP and handler APIs
 - Add CIDR notation support for IP no-proxy rules
 - Add network and timeout exception types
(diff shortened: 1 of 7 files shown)


GrahamCampbell:
Superseded by #3433.

PR #3427, opened 2026-05-22
Use PSR-17 URI factory for redirects

GrahamCampbell:
This change builds on the request and URI factory support by applying the configured PSR-17 URI factory to redirect `Location` headers. Redirect handling is another request-side URI creation path, so using the same `uri_factory` option there lets applications with custom URI implementations keep those implementations when Guzzle follows redirects.

The redirect middleware now resolves the effective URI factory from the request options, falls back to `GuzzleHttp\Psr7\HttpFactory` when the middleware is used on its own, and still wraps malformed redirect locations in `BadResponseException` with the original parsing exception attached. The option validation remains separate from redirect parsing so invalid request options continue to fail as request option errors.

Redirect method rewriting continues to use exact case-sensitive method comparisons, matching Guzzle 8's method-casing behavior. This PR only changes redirect URI creation.


GrahamCampbell:
Superseded by #3433.

PR #3431, opened 2026-05-22
Add PSR-17 stream factory

GrahamCampbell:
This change completes the request-side PSR-17 factory work by adding a `stream_factory` request option for request body stream creation. Guzzle already supports configurable request and URI factories, and request bodies are the remaining request-side objects that Guzzle creates from user options before handing a request to middleware and handlers.

The new option can be configured on the client or per request and defaults to `GuzzleHttp\Psr7\HttpFactory`. Guzzle uses it when converting supported `body`, `form_params`, and `json` values into PSR-7 streams, and when redirect handling resets a request body during a non-strict redirect rewrite. Existing `StreamInterface` bodies are used as provided, while callable, iterator, and multipart bodies keep Guzzle's existing stream handling because PSR-17 does not provide abstractions for those cases.

This remains intentionally request-side only. Response bodies, response sinks, multipart internals, and handler-created streams are unchanged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ Please refer to [UPGRADING](UPGRADING.md) guide for upgrading to a major version
 - Add generic PHPDoc annotations to async HTTP and handler APIs
 - Add CIDR notation support for IP no-proxy rules
 - Add network and timeout exception types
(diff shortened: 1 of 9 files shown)


GrahamCampbell:
Superseded by #3433.

PR #3462, opened 2026-05-24
Add Request index to on_headers callback

SergeC:
When using a Guzzle Pool, there is no way to understand which response arrived for `on_headers` callback. This pull request addresses this issue by passing a request index as a 3rd argument to the `on_headers` callback function, and thus we can understand to which request this response belongs.

Here is an example of code:
```php

use GuzzleHttp\Client;
use GuzzleHttp\Exception\RequestException;
use GuzzleHttp\Pool;
use GuzzleHttp\Psr7\Request;
use GuzzleHttp\Psr7\Response;
use Psr\Http\Message\ResponseInterface;

$client = new Client();

$requests = static function ($total) {
    for ($i = 1; $i < $total; $i++) {
        yield $i => new Request('GET', 'http://httpbin.org/redirect/' . $i);
    }
};

$pool = new Pool($client, $requests(5), [
    'concurrency' => 5,
    'options' => [
        'fulfilled' => function (Response $response, $index) {
            // this is delivered each successful response
            var_dump($index);
        },
        'rejected' => function (RequestException $reason, $index) {
            // this is delivered each failed request
            var_dump($index);
        },
        'on_headers' => static function (ResponseInterface $response, Request $request, $index) {
            var_dump($index);
        },
    ],
]);

// Initiate the transfers and create a promise
$promise = $pool->promise();

// Force the pool of requests to complete.
$promise->wait();
```

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -31,8 +31,7 @@ files, builds multipart requests, works with URIs, or inspects streams.
 Also read the [Guzzle Promises 3.x upgrade guide][promises-upgrade-guide] if you
 use async requests, custom handlers, middleware, pools, or promise helpers.
 
-Guzzle 8 now requires `psr/http-factory:^1.0` directly and no longer depends on
-`symfony/deprecation-contracts`.
+Guzzle 8 now requires `psr/http-factory:^1.0` directly.
 
 [psr7-upgrade-guide]: https://github.com/guzzle/psr7/blob/3.0/UPGRADING.md
 [promises-upgrade-guide]: https://github.com/guzzle/promises/blob/3.0/UPGRADING.md


SergeC:
@GrahamCampbell the tests are broken in 8.0 branch. Broken tests are a result of this PR.

GrahamCampbell:
Thanks for working on this. The use case makes sense: when `Pool` starts multiple transfers, `fulfilled` and `rejected` already receive the original iterable key, but `on_headers` currently cannot correlate the early response headers back to the yielded request. I do not think we should merge this implementation as-is, though.

The main issue is that this pushes a `Pool` concern down into the handlers by adding an internal `_index` request option and teaching `CurlFactory`, `StreamHandler`, and `MockHandler` to special-case it. That means the feature only works for Guzzle’s built-in handlers. A custom handler used with `Pool` can still correctly support `on_headers($response, $request)`, but it would not know to pass the pool key as a third argument. Since `on_headers` is a request option and handlers are responsible for invoking it, adding Pool-specific semantics inside only the built-in handlers makes the behaviour inconsistent.

The naming should also be adjusted. This is not really an index. It is the original iterable key from the requests iterable. That key may be a string, may not be sequential, and may not even be unique for custom iterators. `EachPromise` already preserves that key for `fulfilled` and `rejected`, but internally it separately creates its own numeric pending index because iterable keys are not guaranteed to be unique. The docs and tests should call this a key, not an index.

I also think `_index` is the wrong mechanism. It is an undocumented internal request option, it can collide with user or middleware options, and it leaks through the normal request options array. This is fragile API shape for something that is fundamentally about `Pool` bookkeeping.

A better implementation would keep this at the `Pool` layer. For each yielded request, `Pool` can wrap `options['on_headers']` and close over the current iterable key. The handlers would continue to call the normal two-argument `on_headers($response, $request)` callback, and the wrapper could call the user’s Pool-aware callback with the key. That would work consistently with built-in and custom handlers.

An even cleaner API might be a top-level `Pool` config callback, for example `on_headers`, analogous to `fulfilled` and `rejected`, with a signature like `(ResponseInterface $response, RequestInterface $request, $key): void`. That avoids changing the general request option callback semantics and makes it explicit that the key is provided by `Pool`.
If we keep the current “third argument to request option” approach, I think we at least need tests showing the behavior works with a custom terminal handler, not only `CurlFactory`, `StreamHandler`, and `MockHandler`. We should also update the PR example because `fulfilled` and `rejected` are currently shown under `options`, but they are top-level `Pool` config callbacks.

So I am in favour of solving the problem, but I would prefer this PR be revised to implement the behaviour in `Pool`, use “key” terminology, avoid `_index`, and cover custom-handler behaviour.

PR #3510, opened 2026-05-26
Add response exception base class

GrahamCampbell:
This adds ResponseException as the response-aware request exception base. BadResponseException now extends it, generic response-present request failures use it, and legacy response probing on RequestException is deprecated ahead of Guzzle 8. The existing TooManyRedirectsException response behavior is preserved on 7.11 without moving it under ResponseException yet, so userland construction remains compatible for this release line.


GrahamCampbell:
Closing this in favor of doing the response exception hierarchy work on the 8.0 branch only. Keeping it out of 7.11 avoids changing exact exception classes in a minor release while still letting us keep deprecated response-probing APIs during the 8.x line for a smoother ecosystem migration.

PR #3554, opened 2026-05-30
Add StructArmed to QA

samsonasik:
Hi, this PR adds StructArmed to github workflow static analysis for QA.

https://github.com/boundwize/structarmed

StructArmed is a configurable PHP architecture guard, that we can define layers and rules, then keep them enforced. 

This PR uses the PSR4 preset for starter, and already found violations that fixed and included in this PR.

Since it support php 7.x as well, this only add on github workflow on php 8.2 👍

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -11,3 +11,4 @@
 /phpunit.xml.dist      export-ignore
 /tests/                export-ignore
 /vendor-bin/           export-ignore
+/structarmed.php       export-ignore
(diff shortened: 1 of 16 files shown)


samsonasik:
Ready to review/merge 👍

GrahamCampbell:
Thanks for the PR, but I'm not looking to expand our OSS supply chain risk and do things in a style inconsistent to our other vendor-bin style.

PR #3645, opened 2026-06-16
Reject multipart non-finite floats

GrahamCampbell:
This rejects non-finite float multipart contents before creating multipart request bodies. It extends the existing non-finite float request option enforcement to the multipart option.


GrahamCampbell:
Closing because multipart contents validation is delegated to PSR-7 rather than owned by Guzzle.

PR #3700, opened 2026-06-24
Preserve QUERY method and body across redirects

### LINKED ISSUE #3699: Feature Request: Support HTTP QUERY method (RFC 10008)
## Summary

RFC 10008 standardizes **QUERY**: a *safe, idempotent* method that carries a body.

Sending one already works in Guzzle any method is just a string, and
`guzzlehttp/psr7` carries a body on any method. The gap is narrower: **middleware
with method-specific logic doesn't know QUERY is safe**, and silently rewrites it.

## Concrete problem: `RedirectMiddleware`

On the current 7.x branch, `RedirectMiddleware::modifyRequest()` does:

```php
if ($statusCode == 303 ||
    ($statusCode <= 302 && !$options['allow_redirects']['strict'])
) {
    $safeMethods = ['GET', 'HEAD', 'OPTIONS'];
    $requestMethod = $request->getMethod();

    $modify['method'] = in_array($requestMethod, $safeMethods) ? $requestMethod : 'GET';
    $modify['body'] = '';
}
```

QUERY isn't in `$safeMethods`, so a QUERY following a `303` or a `301`/`302` in
the default non-strict mode is rewritten to `GET` and has its body emptied. For
QUERY that's a semantic break: the body *is* the request.

## Reproduction

Minimal, self-contained (Docker, PHP 8.5, `guzzlehttp/guzzle` 7.12.3 pinned via `composer.lock`):

https://github.com/codexofc/guzzle-query-issue

```bash
make repro
```

Output:

```
PHP:    8.5.x
Method: GET     # expected: QUERY
Body:           # expected: {"q":"foo"}
```

## Proposed direction (input wanted before a PR)

1. Add `QUERY` to the preserved-method set so it isn't downgraded to GET.
2. Make the unconditional `$modify['body'] = ''` conditional on an *actual*
   method change, so a preserved QUERY keeps its payload.
3. **303 specifically**: RFC 10008 §2.4 lets a QUERY response advertise a
   `Location` for an equivalent resource the client can then GET, so `303 → GET`
   may be intended. Keep the downgrade for 303 only, while 301/302 (non-strict)
   preserve QUERY + body? Happy to follow your preference.

## Out of scope / already fine

- **psr7**: `Request` already carries a body on any method — nothing to change.
- **RetryMiddleware**: decider-based, no built-in idempotency list to update.

## Contribution

Glad to submit the PR (fix + `RedirectMiddlewareTest` cases) once the 303
question is settled. Which branch should I target?

**References:**
- RFC 10008: https://www.rfc-editor.org/rfc/rfc10008.html
- HTTP WG draft: https://httpwg.org/http-extensions/draft-ietf-httpbis-safe-method-w-body.html

codexofc:
RFC 10008 standardizes the HTTP `QUERY` method: safe, idempotent, and carrying a request body. `RedirectMiddleware::modifyRequest()` only preserved the method for `GET`/`HEAD`/`OPTIONS` and emptied the body unconditionally inside the redirect branch, so a `QUERY` that followed a 303 or a 301/302 in the default non-strict mode was silently downgraded to a body-less `GET`.

This treats `QUERY` like the other safe methods on a non-strict 301/302 redirect, keeping both the method and the body. The behavior of every other method is unchanged (verified for `GET`/`HEAD`/`OPTIONS`/`POST` on 301/302/303, and for strict mode and 307/308).

Fixes #3699.

### Open design question — 303 handling

This PR makes a `QUERY` follow a **303 See Other** as a body-less `GET`, matching HTTP 303 semantics and RFC 10008 section 2.4 (fetch the advertised `Location` with `GET`). This is the one point still under discussion in #3699 — a maintainer should confirm or override it. Preserving `QUERY` on 303 instead would be a one-line change (drop the `|| $statusCode == 303` guard).

### Target branch

I asked in #3699 which branch this should target; opened here against `7.12` (the current default) and happy to rebase onto whatever you prefer.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,13 @@
 Please refer to [UPGRADING](UPGRADING.md) guide for upgrading to a major version.
 
 
+## 7.12.4 - Upcoming
+
+### Fixed
+
+- Preserve the `QUERY` method and request body when following a non-strict 301/302 redirect (RFC 10008)
+
+
 ## 7.12.3 - 2026-06-23
 
 ### Changed
--- a/src/RedirectMiddleware.php
+++ b/src/RedirectMiddleware.php
@@ -169,11 +169,19 @@ class RedirectMiddleware
         if ($statusCode == 303
             || ($statusCode <= 302 && !$options['allow_redirects']['strict'])
         ) {
-            $safeMethods = ['GET', 'HEAD', 'OPTIONS'];
             $requestMethod = $request->getMethod();
 
-            $modify['method'] = in_array($requestMethod, $safeMethods) ? $requestMethod : 'GET';
-            $modify['body'] = '';
+            // RFC 10008 defines QUERY as a safe, idempotent method that carries
+            // a request body, so a non-strict 301/302 redirect keeps both the
+            // method and the body instead of downgrading to GET like the other
+            // body-enclosing methods. A 303 still resolves to a body-less GET,
+            // per RFC 10008 section 2.4.
+            if ($requestMethod !== 'QUERY' || $statusCode == 303) {
+                $safeMethods = ['GET', 'HEAD', 'OPTIONS'];
+
+                $modify['method'] = in_array($requestMethod, $safeMethods) ? $requestMethod : 'GET';
+                $modify['body'] = '';
+            }
         }
 
         $uri = self::redirectUri($request, $response, $protocols);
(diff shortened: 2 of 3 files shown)


GrahamCampbell:
Thanks for getting started here. I've gone with a slightly different approach, and am treating this as a new feature and not a bug fix. This will land in 7.13 and 8.0 shortly. :)

